PR #6, opened 2014-10-03
Initial shinano driver updates

LalitMaganti:


chirayudesai:
:-1: because of comments on the individual commits.

I think we should be fine with just the qpnp-charger commit, but that does need testing.

sonyxperiadev:
please push only charger commit and specify kernel version from where it is picked

PR #8, opened 2014-10-13
MSM Thermal: Add a generic temperature controller

aloknnikhil:
Implemented an algorithm for better userspace control and simpler automation

NOTE: Will require disabling of the thermal-engine binaries. Should be done via the Ramdisk.
Kernel-based thermal control should offer much better response and should be more efficient


andersson:
Hi nikhil93uf,

Your commit message (and pull request) state that you add an algorithm, but the actual code removes thousands of lines without adding anything. Could you please describe what you're trying to do and why it's better?

Regards,
Bjorn

aloknnikhil:
Hi Anderrson,

My commit completely replaces the exisitng algorithm in the MSM thermal driver.

My algorithm is written from scratch and is thus only a few hundred lines - 

1) This driver introduces fine grained user control by allowing them to specify the CPU frequency steps via sysfs
2) This driver is "smart" in that it implements a moving average filter to consider the last 5 temperature readings 
3) This driver is "smart" in that it measures temperatures in three different thermal zones on the phone and is customizable to include more zones
4) This driver is "smart" in that it monitors as quickly as the temperature goes higher i.e. as the temperature goes higher, it takes more readings, but if the temperature stays below a safe difference, the monitoring is less aggressive
5) This driver also aggressively enforces the CPU frequency by constantly setting it at every monitoring phase

This all could be implemented only by removing the overhead and delays the previous driver was causing.

Also, this has been successfully tested on Xperia Honami (Z1) and the stabiity test on Antutu shows very minor variations in the graph and every benchmark repeated successively yielded almost the same score, thus showing this driver handles thermal throttling effectively.

Best Regards,
Alok

jerpelea:
no updates?

aloknnikhil:
At this point of time, there isn't anything necessary. I have already provided a fully functional code.

PR #40, opened 2014-11-22
adjust regulators and battery values

erikcas:
Regulators 8226_l5 and 8226_l27 differ in Eagle. It causes some drivers to fail to load.
Also took batteryvalues from stock referencekernels


jerpelea:
instead of copying those files you can overlay needed values

erikcas:
the dtsi thing is quite new to me as a die hard 7x27 and 7x27a user. I get the point, will redo it :-)

jerpelea:
lets talk on xda or gtalk

erikcas:
gtalk = erikcas1972@gmail.com

Redid it, will test in half an hour or so. If tests oke, I will open a new pullrequest

PR #64, opened 2014-12-08
changes for eagle

erikcas:
the voldown key might break the stk3x1x driver. I did not test it yet with the current driver in the kernel, which did not work anyway

I replaced this one with the stock stk3x1x for testing, which initially did not work, but did after this commit.

So you might want to hold as I will test later


jerpelea:
please push only regulators in this push request!
input has to be reimplemented

erikcas:
Will do
Op 9 dec. 2014 15:24 schreef "jerpelea" notifications@github.com:

jerpelea:
gpio keys fixed

PR #69, opened 2014-12-11
Seagull hacks

kholk:


jerpelea:
not the best way to do it 
has to checked later

PR #76, opened 2014-12-13
Add CCI HW ID driver

galaxyfreak:
Needed to Identify model variant. Tested on eagle.

### DIFF
--- a/arch/arm/mach-msm/Makefile
+++ b/arch/arm/mach-msm/Makefile
@@ -197,6 +197,8 @@ ifdef CONFIG_MSM_SUBSYSTEM_RESTART
 endif
 obj-$(CONFIG_MSM_SYSMON_COMM) += sysmon.o
 
+obj-$(CONFIG_CCI_HARDWARE_ID) += cci_hw_id.o
+
 #obj-$(CONFIG_CPU_IDLE) += cpuidle.o
 
 ifdef CONFIG_MSM_CAMERA_V4L2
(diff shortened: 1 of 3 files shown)


erikcas:
afaik identification does not work on eagle

galaxyfreak:
root@eagle:/proc # cat cci_hwid_info
cat cci_hwid_info
hwid=5 projectid=1 multisim=0 bandid=0 customer=1

With this commit, it works, now we can use that baseband script again.

erikcas:
I knew that for some time.

galaxyfreak:
Yeah, you actually pointed out cci_hw_id is missing in unified kernel, so I added it back. no problem?

jerpelea:
gsm works nice with aosp ril and there is no need to add this driver

galaxyfreak:
Ehm...Eagle have many variants, how do we set a proper name and LTE/DS props then.

Or we should do another build for each of six variants?

jerpelea:
aosp project suports only LTE devices

galaxyfreak:
Okay then how about D2306, it's also SS LTE, just US version, Or you support International variants only?

erikcas:
@galaxyfreak just clone the repository and add it, if you want to use it.

Personally, I intend not to add it. I opt for SS, DS and LTE versions in separate device subtrees
It is much cleaner

For DS you have to recompile the kernel anyway (offcourse, you can decide to always compile the kernel with DS support, but you always have rubbish in kernels on SS, it might break deepsleep, not tested) so building all in one is not an option anyway

galaxyfreak:
@erikcas, Was that my question? No.
I'm asking about D2306 as thats also LTE.
I was asking jerpelea not you. Am not a fan of butchering everything to "common" repos.

PR #81, opened 2014-12-19
Voice Routing, V4L2 fixes and minor UID-based routing addition

md5555:


md5555:
Sorry about the first, I applied a patch locally instead of picking.

jerpelea:
please pick next time

PR #82, opened 2014-12-19
Updates

Krabappel2548:


chirayudesai:
See my comments on #81, and on http://review.cyanogenmod.org/82332 too

prima is in a separate repo on Sony, and built as a module, so it can be outside the kernel.

The vidc patch shouldn't be needed on AOSP.

sonyxperiadev:
please push only 
rhine mic fix 
msm: vidc: Send resolution along with output buffer
Handle 'sk' being NULL in UID-based routing

PR #85, opened 2014-12-22
Yukon clocks cleanup and mdss logic fix

kholk:
The clocks cleanup eventually permits to compile one kernel for all, obviously including only the right DT in the boot image. That may be "funny" for Android build purposes, where we will be able to delete all the prebuild kernels from the device-specific trees.
This way we'll be able to distribute the zImage into the common yukon folder and only the DTB into the device-specific folder... oh, and we won't get overhead on the device itself, too. People won't download the same kernel (apart some bits) 4 times this way.
An eventual (not going to happen, I know) bootloader update could use different board-ids for the four Yukon devices to make us able to use the very same boot.img for all.

Apart this, I was examining the Flamingo badness a lil bit again, also because effeectively Seagull isn't suffering from this and, if it was eMMC, then... you know. Still not sure, but, well.
Was saying, logically, if the panel isn't on it can't receive the needed commands to display the actual "image", so, it's mentally insane to send the display ON command AFTER unblanking the display. Corrected this, but I still don't really know if this fixes what we want. Just logic.

### DIFF
--- a/arch/arm/boot/dts/msm8926-yukon_eagle-720p-mtp.dts
+++ b/arch/arm/boot/dts/msm8926-yukon_eagle-720p-mtp.dts
@@ -18,7 +18,7 @@
 
 / {
 	model = "Qualcomm MSM 8926 MTP";
-	compatible = "qcom,msm8926-mtp", "qcom,msm8926", "qcom,mtp";
+	compatible = "qcom,msm8926-mtp", "qcom,msm8926", "qcom,mtp", "somc,eagle";
 	qcom,board-id = <8 0>;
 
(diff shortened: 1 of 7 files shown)


andersson:
This is cool!

We're trying to put the necessary pieces in place in mainline Linux to be able to bring this further - i.e clocks and gpiomux (and everything else...) are coming from DT instead of hardcoded values in c-files. But for now, I really like what you're achieving with this old kernel!

kholk:
Thanks!

kholk:
Ah, about your project, perhaps we may talk about it in private.

PR #90, opened 2015-01-11
MSM8226 NFC

stellirin:
Phew! this took a while to track down, the PN547 driver in this kernel silently fails without this. Only when I tried tianchi's stock driver did I see an error about this.

This fixes NFC on tianchi, but flamingo remains broken and crashloops. I had issues with flamingo using its stock kernel on CM (otherwise it was fully functional), so there's probably some other issue on this device. More vague errors:

01-14 04:39:35.611: E/NxpHal(1244): read status error status = 16f
01-11 03:09:04.619: E/NfcService(1244): Watchdog triggered, aborting.
01-11 03:09:04.620: E/BrcmNfcJni(1244): nfcManager_doAbort: abort()
01-11 03:09:04.620: A/libc(1244): Fatal signal 6 (SIGABRT), code -6 in tid 1464 (enableInternal)


kholk:
Better examine this. NFC worked without this on one of my private aosp builds.
Unless it's because of later changes on the kernel, this shouldn't be needed for Flamingo and seagull.

Moreover, your commit makes disorder in the clocks file: you've commented out another entry that was for another pin.
My previous commit introduces device-specific clocks being initialized AFTER the standard ones, hence "overwriting" 'em. You should really insert this clock in device-specific struct and without ifdefs.

stellirin:
@kholk the original tianchi driver uses this: see [19.1.1.A.0.xxx/drivers/misc/pn547.c](https://github.com/sonyxperiadev/kernel-copyleft/blob/19.1.1.A.0.xxx/drivers/misc/pn547.c#L517)

which is selected by `CONFIG_NFC_PN547_PMC8974_CLK_REQ` in the generic driver.

I am using the relevant pn54x patches to libnfc-nci from google's gerrit, which is hard coded to `/dev/pn544`.

Also apologies for forgetting the device-specific clocks, I was dicking around in userspace all day because the generic kernel driver was reporting 'success', even though this is needed. I was just glad to finally get it working.

I'm curious to know what your solution is, I've always had a problem with NFC on flamingo :weary:

kholk:
About NFC working, I wasn't talking about tianchi! I was talking about just only seagull and flamingo! 

Sure tianchi needs this clock to bring nfc up, as we can see from its factory default kernel and I'm not saying it doesn't.  This is obviously needed and only for it because, as I said, I remember it was working on the others. About how, well, see the kernel config for the two devices for kernel things. 
About userspace (AOSP) I used two patches for pn547 support that were out of approved ones in AOSP. 
At the end of all, my userspace library was called nfc_nci.pn547.so (if I remember well). Can't tell anything more right now because I'm not at home, but I can remember I could read an NFC tag and transfer a Web link between my z1 and the other two devices. 

Further informations will obviously come asap! 

Ah, last thing. Don't worry about the mess. Just fix it ;)

PR #91, opened 2015-01-15
New WiFi driver and board-wifi bringup

kholk:
Absolutely needs to be tested.


LalitMaganti:
Not sure if I'm missing some piece but my kernel is not compiling with these commits for sirius. I get bunches of errors complaining about missing symbols.

chirayudesai:
@kholk, what firmware do you use this with? We haven't had luck with anything but sony stock, AOSP firmware works but breaks one thing or another.

PR #118, opened 2015-01-29
eagle: fix the dancing menu's

erikcas:
- M2 users should not have to chase menu's
- rotation is much smoother now
- @kholk @jerpelea can adapt this to the unified kernel

Change-Id: I9ba1a09af764580d292cbd761fda4b5d13df2d96
Signed-off-by: erikcas erikcas1972@gmail.com

### DIFF
--- a/drivers/video/msm/mdss/mdss_mdp_pipe.c
+++ b/drivers/video/msm/mdss/mdss_mdp_pipe.c
@@ -930,9 +930,14 @@ static int mdss_mdp_image_setup(struct mdss_mdp_pipe *pipe,
 	if (mdss_dsi_panel_flip_ud()) {
 		if (pipe->mfd && pipe->mfd->panel_info &&
 			pipe->mfd->panel_info->pdest == DISPLAY_1)
+#ifndef CONFIG_MACH_SONY_EAGLE
 			dst_xy = ((pipe->mixer->height -
 				   (pipe->dst.y + pipe->dst.h)) << 16) |
 				pipe->dst.x;
+#else
+			dst_xy = (((960- dst.y - dst.h) << 16) |
+			(540- dst.x - dst.w));
+#endif
 		else
 			dst_xy = (pipe->dst.y << 16) | pipe->dst.x;
 	} else


erikcas:
I know there is probably a more elegant way to fix it instead of hard values. To me it is important that it is functional on Eagle. 
Also I donot know it would break other devices if I donot ifdef it.

Feel free to reject and fix it like you want to. Rootcause is exposed now

galaxyfreak:
more elegant way http://review.cyanogenmod.org/#/c/87284/4/drivers/video/msm/mdss/mdss_mdp_pipe.c

kholk:
@galaxyfreak This code is used only for Eagle. Remove the ifdef and send a pull request here, please :)

erikcas:
amazing. I reported earlier, you have the device you said, but you did not fix

kholk:
@erikcas Sorry for that, Eagle is not the only device I look at. Sincerely, I totally forgot this little detail in Eagle after fixing LOTS of bugs on other devices.

erikcas:
Oh sure, respect for all your and @jerpelea work here. But declaring a kernel stable in an xda thread while one of the most important parts of the phone is not visibly stable, makes me think one has to do his homework better. Maybe recheck stability of 3.4 kernel before porting instable things to 3.10 kernel?

Anyway, spend 15 minutes on this kernel to make it useable to endusers and a remark to someone to redo the commit and open a pull request is a "whole lot of disrespectfull" to the original committer here don't you think?

@galaxyfreak open a pullrequest in your own name. But donot forget you were not able to solve it

erikcas:
and yup this frustrates me.

kholk:
I told him to re-commit because I've just seen his solution and seemed more clean than yours (even if it's based on it).
For me it's not important who makes the pull request, btw, didn't mean and didn't want to disrespect you in any way, so I'm sorry for this too.
Everyone, as people and as developers HAS to be respected and I'm doing my best to do that.

Btw, it's obvious you did a good job by exposing the possible solution.

erikcas:
In the end, users are to be served with fixes. I donot care about who commits it either.

In my kernel I never had this problem, it took me 15 minutes to find it (haven't checked this kernel in weeks). What does concern me is that other persons use this kernel as a base for weeks and are not able to detect the cause. 

When it is common that in such a case someone is allowed and motivated to commit as his fix, it is not right to me.
Donot get me wrong, I have nothing against @galaxyfreak and again, how it goes in, I dont care.

But it vanishes my motivation to research and share any improvement. So actively I am out again.

Good luck with your project and keep up the good work.

Erikcas out :)

PR #119, opened 2015-02-01
Update msm8226-yukon_eagle-720p-mtp.dtsi

Wadim111:


jerpelea:
Why don't you use the driver that is already in kernel?

PR #120, opened 2015-02-01
Update input.c

Wadim111:


jerpelea:
Please add commit description

jerpelea:
with proper driver in place this fix should not be needed

PR #121, opened 2015-02-01
Update Makefile

Wadim111:


jerpelea:
This makefile without the actual driver will only cause kernel compile error
Add driver and commit description

erikcas:
Alin, I see my commits coming in from other users now. So I did not go mad. Light and proximity sensors donot work on eagle with current driver. To make this really work current stk3x1x needs to be reworked or Eagles driver from stock kernel has to be brought in.

This does not work without mentioned Eagles driver. Which you will not merge.

All last 3 pull requests I have already have had in my kernelsources and together with the missing eagledriver it makes light and proximity sensor work on eagle

jerpelea:
The reason i have not accepted those commits is that they are yours and i would like to see you push them, you are investing your time in this project and is your right to push.
regarding light sensor i think that we should try and use the one that is in kernel instead of adding new kernel driver. 
thanks

erikcas:
Oke clear thanks. And sure I agree on your point of view about using the
existing one. Just dropping in a separate driver is usually a workaround
instead of fixing a rootcause :-)
Op 3 feb. 2015 16:54 schreef "Alin Jerpelea" notifications@github.com:

jerpelea:
Hi Erik,
Please investigate existing driver. I suspect that the root problem is incorrect dts config.
Thanks

erikcas:
oh and I was tired from developing.. Oke, back to source, adb and kmsg :-)

PR #144, opened 2015-02-08
[3.4] PN547 for Rhine

kholk:


alviteri:
@kholk Is this will update for 3.10? thanks

kholk:
This is only an example and it's still not working, unfortunately.

There are some problems to solve with it (NFC init fails in userspace for some reason) and... well, yes, there's a possibility that this will be taken back into consideration for 3.10.
However, we're actually solving other problems on other devices.

Our plan is to make all devices to be perfectly working, so, we're solving bugs with the highest priority, first.... and this commit, well, that's a cleanup and driver unification, more than a bug solver, so this will come as last, unless we get positive help from someone else on this matter.

I mean, if someone tells me that this is working with X modification, I'd be really happy to test immediately and merge ;)

alviteri:
@kholk  other question and maybe last .... when you test this .. did you made the proper changes in device tree ? and I don't know about kernel 3.4 because I just test 1 time aosp when was working with kernel 3.4 ... but I can see you are changing the clock from 6-0028 to 2-0028 and actually I can see the PN544 is using 2-0028 while PN547 is using 6-0028.....

kholk:
To reply to your question, look on the commits, there is one that is updating Rhine clocks..........

alviteri:
@kholk yeah in this
https://github.com/kholk/kernel/commit/86f56b8baa82cc50ab1be5d782582ec85a6e1c5b
But you are updating from 6-0028 to 2-0028 and PN547 is using 6-0028 on shinano targets 

Anyway ... Thanks for the info and patience

PR #145, opened 2015-02-09
Migrate to in-tree kernel compilation, by defining the toolchain in devi...

locusf:
...ce trees.

This is due to making it easier to build the kernel in-tree with special manifest
that pulls in kernel sources for each device, like some sony devices.

### DIFF
--- a/AndroidKernel.mk
+++ b/AndroidKernel.mk
@@ -66,30 +66,30 @@ $(KERNEL_OUT):
 	mkdir -p $(KERNEL_OUT)
 
 $(KERNEL_CONFIG): $(KERNEL_OUT)
-	$(MAKE) -C kernel O=../$(KERNEL_OUT) ARCH=arm CROSS_COMPILE=arm-eabi- $(KERNEL_DEFCONFIG)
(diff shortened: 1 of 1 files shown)


chirayudesai:
This makes it hardcoded to a specific version in a specific path, what you could do is just export PATH=$(ANDROID_BUILD_TOP)/prebuilts/gcc/linux-x86/arm/arm-eabi-4.8/bin:$PATH

jerpelea:
Since this is intended to be an alternate to prebuilt kernel i don't see the point of path extension 
since we always use android compiler

chirayudesai:
For in-tree compilation, if you run `source build/envsetup.sh`, then `lunch`, then a 'which arm-eabi-gcc', you'll see it picks up the version from prebuilts, and so does a `which arm-linux-androideabi-gcc`.

Now that I looked at the build system and checked locally, I don't see how this patch is needed at all.

And even if you were do it,

CROSS_COMPILE ?= thepath
Then a CROSS_COMPILE=$(CROSS_COMPILE) like you did earlier would be the best approach IMO

PR #148, opened 2015-02-09
Replace AndroidKernel.mk with a much nicer version

chirayudesai:
- Less hardcoded paths, more configuration supported.
- From CyanogenMod/android_build@3d29d65eab279133e7be1c679c6c6b7de6ea37b1
- Set TARGET_KERNEL_SOURCE to the kernel dir, and TARGET_KERNEL_CONFIG
  to the defconfig.

Change-Id: I2f8af563b21cdafd0d57ce5d8d5f6c6f4162a1fe


chirayudesai:
TODO: 
- ~~Add a flag to include this only when we want to build inline.~~
- ~~Get rid of stuff we don't use~~
- ~~Hardcode more things as this is going to stay here.~~
- ~~Copy dtbs to somewhere else in $OUT~~

Update: All done! :)

PR #155, opened 2015-02-10
changes for cam (only for reference)

erikcas:
compare https://gist.github.com/erikcas/e6afb1ce1182b343af76 (without this patch) to https://gist.github.com/erikcas/09f9a500650a77835736 (after applying this)


erikcas:
I will close this as another pull request is open for this

PR #159, opened 2015-02-14
yukon: Add tianchi_dsds support

abhinav-jhanwar:


jerpelea:
Please implement clocks using same way as the other msm8226 devices

Abhinav1997:
As in separately for tianchi and tianchi_dsds(without using OR)?

PR #186, opened 2015-02-22
eagle: fix the dancing menu's

erikcas:
- M2 users should not have to chase menu's
- rotation is much smoother now
- @kholk @jerpelea can adapt this to the unified kernel
- thanks to @galaxyfreak for simplifying original code
  Change-Id: I9ba1a09af764580d292cbd761fda4b5d13df2d96


erikcas:
now tested :)

PR #189, opened 2015-02-23
arm: DT: 8226-regulator: Enable 8226_s1_corner, set min uV to 5

kholk:
The minimum uV is 5 for the proxy consumer and this is what actually
makes the regulator stable.
Also, enable the regulator when probed, so that wcnss hardware will
have the required time to finish its powerup sequence and when we
are actually probing and uploading the firmware, we'll find it ready.

### DIFF
--- a/arch/arm/boot/dts/qcom/msm8226-regulator.dtsi
+++ b/arch/arm/boot/dts/qcom/msm8226-regulator.dtsi
@@ -129,10 +129,13 @@
 			compatible = "qcom,rpm-smd-regulator";
 			regulator-name = "8226_s1_corner";
 			qcom,set = <3>;
-			regulator-min-microvolt = <1>;
+			regulator-min-microvolt = <5>;
 			regulator-max-microvolt = <7>;
 			qcom,use-voltage-corner;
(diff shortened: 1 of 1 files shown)


erikcas:
If this is to unbreak reboot on splash boot logo, I have to disappoint you.
8226_s1_corner still breaks boot.

It boots if I rename it to 8226_s1 (like it is on stock) but then wcnss reports a regulator constraints issue. See https://gist.github.com/erikcas/348146c670dadb0f06fd

erikcas:
it boots fine on 18.0.C.1.13 firmware like you mentioned.....

kholk:
Yeah, new firmware will crash... don't ask me why! :-1: 

We'll see what the hell is happening with that as soon as possible, also because Flamingo and Seagull are crashing, too, because of wcnss.
In the meanwhile, this is the only way to make it to work normally, so... :)

erikcas:
haha... Btw does deepsleep work on AOSP? Device stays on as long as on usb. But when de-attached it reboots in about a minute (which I set on the device as suspend time)

jerpelea:
It works on 3.4 kernel
3.10 kernel is WIP and is not working yet

PR #200, opened 2015-02-26
tianchi: Enable kernel logo from common tianchi dtsi

abhinav-jhanwar:
Its better to enable kernel logo from a dtsi both ss and ds variants use.


kholk:
This is not the structure we want to give to Yukon. Actually this will break the "human-readable" logic in DT and it's not good to do that.

Also, we're about to restructure that and put it in Yukon common, so this shouldn't be merged.

abhinav-jhanwar:
I asked jerpelea about putting it in Yukon common(some days ago) but he told me for now it would be better to keep it separately for each device. If its OK now, then yeah it'll be better to put it in Yukon common. Will close this. Thanks :)

PR #235, opened 2015-03-09
[3.10] Solve vmalloc problems, ASoC cleanup, more MDSS calibration

kholk:


kholk:
Yeah, I absolutely know about that.... and don't worry, I'm 99% sure this is happening only on x86 architecture.
ARM should be totally fine :)

jerpelea:
in case we run into problems we will follow mainline revert 
for now seems ok on ARM

Tommy-Geenexus:
I think it will be fine, the issue could not be 100% related to this commit either way, just wanted to bring this up in case anything bad happens ;)

kholk:
Ah yeah, that's obvious, we appreciate that ;)

PR #237, opened 2015-03-09
Fix eagle deepsleep

erikcas:
dev_get_drvdata is apperently needed to prevent a reboot on deepsleep.
If I use dev_to_iio_dev it reboots

edit: corrected in CONFIG_PM
dev_to_iio_dev can be used after all


jerpelea:
https://github.com/erikcas/kernel/commit/6adb7114ffe4a02950bea95795590a62712cc847 should  be part of https://github.com/erikcas/kernel/commit/d8bbdcd048ea64948b9306f8e59e8d1bbca89b08
for code readability

PR #244, opened 2015-03-11
Aosp/au linux android la.bf.2.1 rb1.05.00.00.173.012

savoca:


kholk:
This is not sane, the whole commit set is actually duplicating what Qualcomm's implementation WITH our unified display driver for Sony devices already does and actually, we are controlling that in DTs (not sysfs).

While the concept may be good for some extents, this is adding complexity to the code and, you know, increased complexity means decreased human readability, making the whole MDSS implementation way more difficult to understand.

What you want to implement with all those lines of code (compromising original QC code too) can be done in really A LOT less lines, just in one driver that, as already said, already implements all those functionalities.

But I don't want you to misunderstand me: as I said, the concept is good. I'm only complaining about the way you implemented it.

savoca:
Just had a second run through of your MDSS, and I had no idea you guys where already using picture adjustment in your own Sony drivers. Bear in mind this was originally implemented for literally any other device using MDSS (I have a different one for msm8x10) and since I saw no other OEM using it, I assumed you guys weren't either.

Can you elaborate on the "compromising original QC code" part?

kholk:
Sure!
What I mean is that you're actually modifying original code by Qualcomm (that works just fine) to forcefully enable color calibration, to be used subsequently with your driver implementation.

Actually, you should take in mind that not all MDP HW revisions can really do that, and some will just be really slow (even if they work). I can see you're checking for rev < MDP_HW_REV_103, but you're underestimating the situation because you ignored most of the situation.

To be really working for everything, your code should be checking (hence again duplicating functionalities) for all the revisions and decide whether to support or not the picture adjustment! You may ask why we aren't doing that in our Sony driver, and that's simple: we know that we won't have any bad situation because we won't have to deal with old MDP, simply because there's no device we support using that.

Why we can do that in our driver and you can't?
Also simple: you're writing a "global" thing, we're writing a "vendor-specific" (so, dealing only with devices we know) driver :D

Again, by forcing enablement of the picture adjustment, you're also underestimating the badnesses that may happen in other parts of the MDSS driver, like delays in command send and/or potentially making impossible to communicate with a display prior it to get online for the first time.

This is actually one of the (major?) reasons why I've implemented this in DT, making it to be enabled only after full display initialization and poweron.

savoca:
Can you share some device-specific limits here? Qualcomm seems very generous about which devices can handle picture adjustment (their only limit is actually the same HW_REV_103 check that I have) and the only issue I've ever seen is on msm8x26 devices as they cannot handle high hue adjustment very well.

Also the KCAL driver itself has a late_initcall, I had issues during initial development with an earlier call but absolutely none now.

Do you have any suggestions for improvement on this driver (for my other devices/users) ? Thanks!

kholk:
Some devices actually need to recognize which display they have and this will need exchanging some informations between the IC and the kernel display driver, to make the right decision.
At this stage, we want only basic "low-level" communication (TX commands, RX the reply) and nothing shall interfere with it.

About other devices, I really don't know but, from my experience, you have to be really cautious on what you do with Qualcomm things. Those drivers are really fragile, doing only what's strictly necessary to work at the first stages (then, when everything is working that's another story because there will be lots of checks).

My suggestion, actually, apart being careful while writing anything on this, is to remove every single line you added (and/or removed) in original Qualcomm code and work only on the OEM-specific layer.
The display driver is not meant to be modified by users/OEMs/developers and there's a specific part of it that is meant to be: the "final" panel driver.

So, your sysfs implementation, detached from the actual -whole driver- (new *.c file, like you did), is good... but it can be done in the panel driver.
Moreover, it should be more synthetic... I mean, all those lines aren't necessary.

If you want, you can actually shrink the size of your code by at least 50%, increasing human readability to a certain extent.
Apart that, I didn't examine your whole code, but I recommend to apply color calibrations only after powering on the display.

savoca:
I see your points, and I'll see if I can cut it down. Currently I haven't actually altered any Qcom code, just added functions to the PP driver so far, called by the KCAL driver.

Thanks for your input!

PR #258, opened 2015-03-17
dts:msm8226-720p-mtp: Remove nodes already present in  yukon-common

galaxyfreak:
Tired of seeing same stuff throughout all files.

Note: Only leds are have some diff between yukon-common, rest is completely same.


jerpelea:
you are not supposed to clean qcom files!

PR #273, opened 2015-04-10
msm: mdss: mdss_dsi_panel_driver: replace msleep() with usleep_range()

Tommy-Geenexus:
- msleep < 20ms can sleep for up to 20ms (see Documentation/timers/timers-howto.txt),
  thus replace it with usleep_range().


kholk:
Togari isn't hitting that code at all!
As I already said, yes it should be fine. Don't worry, we'll test it for you asap! :D

Tommy-Geenexus:
I know, no rush, just wanted to point that out :D

kholk:
Oh yes, I knew, and I totally forgot about it so thanks for bringing the topic back to my head :D

PR #279, opened 2015-04-14
shinano: wifi: add missed wifi implementation for sdhci

humberos:
this change fix kernel compilation process.

Change-Id: I40bf6f22631e2af8a662bf10381e628a58dd271a


jerpelea:
for the moment we will use legacy sdcc implementation

PR #285, opened 2015-04-22
shinano: leo: switch jdi panel to video mode

humberos:
some leo variants (like D6643 - samba) are using JDI panel (cmd mode).
new 3.10 kernel uses video mode

Change-Id: I06236b53e713711716dc9f3c38b9d31fb6b86d66


humberos:
moved to 299.

PR #293, opened 2015-05-04
prima: undouble a constant

erikcas:
declaring double const seems useless to me, hence remove one


erikcas:
speaks for itself I presume

jerpelea:
commits have to have a description message

erikcas:
oops wrong button

PR #297, opened 2015-05-10
driver: video: mdss: fix boot process for cmd mode

humberos:
this change fix boot process for panel with cmd mode
the weird display corrupting still present at least for AUO and JDI panels

Change-Id: Ie80c4a266f81d43f70681994a94ba2a92a4a2378

### DIFF
--- a/drivers/video/msm/mdss/mdss_dsi_host.c
+++ b/drivers/video/msm/mdss/mdss_dsi_host.c
@@ -2077,6 +2077,7 @@ int mdss_dsi_cmdlist_commit(struct mdss_dsi_ctrl_pdata *ctrl, int from_mdp)
 	int ret = -EINVAL;
 	int rc = 0;
 	u32 ctrl_rev;
+	bool hs_req = false;
 
 	if (mdss_get_sd_client_cnt())
 		return -EPERM;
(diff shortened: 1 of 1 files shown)


kholk:
Interesting! Looks hackish, though. 
Can we maintain compatibility with qcom's req->flags? 

If not, we're merging it as it is.

humberos:
Actually it is present at CAF sources.
I think it is ok to go.
One question: why did we remove switch panel mode feature?

kholk:
If it's on CAF, then you should merge this commit by cherry-picking it, not changing the ownership of it (it'll anyway count as your contribution).... that's the right way of doing it ;)

Send another PR with the correction I said and it'll be merged for sure.

About comm mode switch, this feature wasn't removed: it has NEVER been implemented on our panels!
I have plans to implement it soon because.. yes, it's a very interesting feature, indeed, and I'd like to use that for all devices. Meaning not only Shinano, but also Rhine and Yukon, if possible.

Like I've replied to another guy, we're solving the highest-priority bugs first, prior adding features, so, the auto switch will come soon, but not really soon.

Btw, good job about finding the commit ;)

humberos:
Ok!
I am sorry!
I am uploading new PR with CAF changes.
Thanks for replying me.
Thank you!

jerpelea:
As general rule when you get patches from somewhere chereypick instead of recomit
Br
Alin

PR #299, opened 2015-05-10
AU_LINUX_ANDROID_LA.BF64.1.2.1

humberos:
tested on shinano | leo | JDI panel with video mode.


kholk:
Merged after testing on other Shinano, Rhine and Yukon.

kholk:
I mean, testing tomorrow!

humberos:
No problem. I am still testing another upstream commits. I will upload all commits today into this PR then you'll can test all synced commits. Thanks!

alviteri:
I will test on rhine - honami

kholk:
This is such a long list of commits!

So you've completely updated the MDSS to the latest available version from QC. Good job about the last 4 commits, I completely missed those (I usually don't care about updating if not really needed, as you see)....

Anyway, I don't know if this is a REALLY good idea. I see some fixes for things that apparently don't need to be fixed (because ofc they're working).

Testing this PR will require some time, as the commits you've added after my first reply will require extensive tests on all devices.
Btw there are some particular commits in this PR that are particularly interesting for me.

Starting the tests now, but I can't promise an ETA, obviously.

humberos:
I see!
If you want we can split this work by using the important change[s] first.
Just let me know ;)

kholk:
Yeah, don't worry. I'm testing the full set, first.
If anything goes wrong, I'll let you know!

humberos:
We have two models of panels for Shinano Leo: AUO and JDI (at least).
Although we are moving all panels to video mode but I am stuck on AUO and JDI panels with a weird display corrupting (noise, glitches, ....). JDI panel is broken only when I am using CMD mode but it is ok with video mode (https://github.com/sonyxperiadev/kernel/pull/285). AUO panel is broken for both :(
I saw your Scorpion commit which you made some h and v porch changes. So, did you use a specific datasheet? Is there a way to get this datasheet for AUO?

kholk:
I didn't use the datasheet as it's not available. I've reverse engineered the whole thing and, after understanding it, I wrote that solution.

I don't think that there's any publicly available datasheet for that AUO panel.....

galaxyfreak:
Booted fine on Yukon (eagle). Doesn't break anything obvious.

erikcas:
fix, revert or delete https://github.com/humberos/kernel/commit/c3e91d7cd76fe0779609ca0d8ff642953ab98c1e

it will leave us with a black screen when booting to recovery. Tested this already a few days ago.

We can as well merge the entire caf LA.64.1.2.1 (at least for yukon) when it comes to booting fine.

See https://github.com/erikcas/android_kernel_caf_msm8x26/commits/b9a0a4e1c6237279c0b06c3d50e350c7795bcd68

(yes, while doing caf merging, also updated to 3.10.76 upstream. It boots fine, but this is hobby :P
But also tested LA.BF.64.1.2.1 on plain sonyxperiadev tree )

humberos:
Ok, I can confirm that recovery issue on Shinano Leo. Thank you @erikcas!
I will revert it just preserving the commit history reasons.

erikcas:
your welcome :)

kholk:
Revert what? :P :P :P 

Reset your tree and do not include that commit at all, don't make me see bad reverts please.
About the Leo panel switch, PLEASE not in this PR. The commit will not be merged until I test my plans about it.

@erikcas Good job! You should also run AnTuTu to make sure the benchmark can get to completion ;)

erikcas:
LOL, will do. First solve latest small issue on LP-5.1 omnirom in combination with updated 3.5.1 branch and have camera working on it

kholk:
@erikcas And please, also measure performance between the old tree and the one with this PR as it may impact on it in a good or bad way.
Also, your comment about updating the whole thing and re-porting on a new branch.... this will not be done, and I mean... NEVER!

We have a reason to use this specific version of the tree, so, don't lose time with new complete trees... but fix this, if necessary, instead.
Thanks for the help on testing this. I've got a lot of things goin' on here!

erikcas:
@kholk thats why I commented "its a hobby" to it. My point there is, that booting fine is not enough, it needs further testing like you said in for example antutu.

For merging caftree, updating trees to 3.10.x or reI will NEVER open pull requests, and all comments here about kernel are always based on testing with current most up-to-date sonyxperia kernel tree

kholk:
Yep, I was saying that to not make you to lose time ;)

erikcas:
will not lose time ;-) having multiple buildenvironments has advantages like doing things simultaniously

and if my hobbythings lead to interesting facts for the kernel I know how to contact you ;-)

kholk:
Absolutely!

humberos:
PR updated.
Ok, I removed that commit as you said @kholk .

kholk:
@humberos also remove 756a9c2 and c3e91d7 please

humberos:
ok :)

humberos:
Done!

necioerrante:
I still get "black screen when booting to recovery" with this update on Shinano Aries with JDI panel.
Is there still something else needing to be done?

humberos:
Your problem is panel DTS setup.
We need to do a refactoring for panel timings.

humberos:
Errrr.... I am sorry. I think your problem is kernel driver. I will check it today later.

humberos:
Do you getting this error even without this change, right?

alviteri:
Wait @necioerrante you are talking about boot into recovery as new bootloader for Yukon targets??? If yes nope that isn't implemented yet...  and if you hardboot another boot.img for example from cm you will still have recovery working

necioerrante:
@humberos 
Yes, I have the same issue before and after the change.
It is an issue with 3.10 kernel.

@alviteri 
No. Shinano aries: With 3.10 kernel recovery does not work, only blank screen.
Even with custom ROM.
With custom ROM on 3.4 kernel it worked.

jerpelea:
In this case there is no recovery enabled for shinano

---- necioerrante wrote ----

@humberoshttps://github.com/humberos
Yes, I have the same issue before and after the change.
It is an issue with 3.10 kernel.

@alviterihttps://github.com/alviteri
No. Shinano aries: With 3.10 kernel recovery does not work, only blank screen.
Even with custom ROM.
With custom ROM on 3.4 kernel it worked.

alviteri:
Thanks Alin that was my point :)

humberos:
The point is: recovery is not working for Shinano Aries (at least).
It's working fine on Shinano Leo and Sirius (CMD or Video mode).

alviteri:
Ok for now I worked with 

Fastboot boot cmboot.img (after vol up for enter to cmrecovery)

I thought you are talking about recovery as new way for Yukon....

alviteri:
Anyway thanks for the info

necioerrante:
Thank you all for the responses
Also, recovery worked with cm kernel.
And, I believe it worked also with cleaned 3.5.2.2 aosp branch, 
Will appreciate if you have any ideas or suggestions for me to test?

PR #317, opened 2015-05-21
drivers: tty: serial: Clean up SONY serial driver merge

humberos:
a call to msm_hs_disable_flow_control() was replaced with a sequence of register
reads/writes that performed an identical function EXCEPT for changing the RFR
register to RFR_HIGH

However, the subsequent call to change the status register to RFR_LOW was removed.

This mismatch is the cause of Bluetooth reconnect issues

the original change:
https://github.com/sonyxperiadev/kernel/commit/9cdddb2555dcdd45f49107b33e6f0993a640348a

Signed-off-by: Andrew Dodd atd7@cornell.edu
Change-Id: I05840280110bc4b51e72e30bd1c108368a971bdc

### DIFF
--- a/drivers/tty/serial/msm_serial_hs.c
+++ b/drivers/tty/serial/msm_serial_hs.c
@@ -1063,7 +1063,20 @@ static void msm_hs_set_termios(struct uart_port *uport,
 	msm_hs_write(uport, UART_DM_IMR, 0);
 
 	MSM_HS_DBG("Entering %s\n", __func__);
+
+#ifdef CONFIG_MACH_SONY_SHINANO
+	/*
+	 * Clear the Rx Ready Ctl bit - This ensures that
+	 * flow control lines stop the other side from sending
+	 * data while we change the parameters
+	 */
+	data = msm_hs_read(uport, UART_DM_MR1);
+	/* disable auto ready-for-receiving */
+	data &= ~UARTDM_MR1_RX_RDY_CTL_BMSK;
+	msm_hs_write(uport, UART_DM_MR1, data);
+#else
 	msm_hs_disable_flow_control(uport, true);
+#endif
 
 	/*
 	 * Disable Rx channel of UARTDM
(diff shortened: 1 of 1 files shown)


humberos:
Not sure if only Shinano devices have this issue!
Please, could you confirm it and then I will change this patch.

kholk:
should be only shinano... 99% sure

erikcas:
still is heavily reconnecting on Eagle.
:( have to reconfirm pairingcode multiple times a day on my Smartwatch 3. After second reconnect pairing is impossible and to have connection I have to reset my watch.

Previous kernelbranches donot have this issue.

Imo reconnecting is related to heavy lagging issues this kernel has on occasions. When it lags, connection dissapears, wifi resets, mobile network resets. 

It seems not to appear on gapps-less phone. But for android-wear I need gapps. (Antutu scores are showing instability as well. scores are between 15000 and 20000)

will take some longer periods of logging. I see no abnormalities in the ad-hoc logs I take when problems occur.

(testing on AOSP, not on omni this time, with unmodified kernelbranch)

humberos:
Oh! This change was merged/tested only on shinano devices.

erikcas:
It does not break anything :-)
But no improvements either :-(

:-)
Op 27 mei 2015 12:49 schreef "Humberto Borba" notifications@github.com:

humberos:
@erikcas 
well, no improvements for yukon are you meant.
because BT was fixed on shinano devices.
Could you please provide more information like dmesg or logcat?
I don't know yukon devices but we can try to understand the problem by checking logs.

humberos:
I think if yukon has BT issues then rhine should have the same too due to both are using the same way.

erikcas:
@Humberos, like I mentioned, imo it is not BT driver related.

It seems to me it is related to lagging (on yukon) in combination with
google apps. It seems not to occur on plain aosp rom. BT headset is paired
okay and persistent (no reconnection issues)
Op 27 mei 2015 12:58 schreef "Humberto Borba" notifications@github.com:

humberos:
@erikcas humm... ok! thanks!

erikcas:
well, will start logging first and then try to reproduce. Like mentioned, take logs when problem is already occured show no abnormalities.

Your welcome keep up the good work :-)

Kali-:
shinano use broadcom and yukon qcom bt

humberos:
yeah! it's true! Good point btw!

PR #331, opened 2015-06-15
misc: qcom: qdsp6v2: fix uninitialized variable

humberos:
Wrong variable is used for copy_from_user().
It causes user setting is not set properly to audio drivers.

Signed-off-by: Fred Oh fred@codeaurora.org
Change-Id: I1d54b9b20d3664045c24ffa2066ecc1b7abe5e87


jerpelea:
code is not, used but is still a fix!

PR #337, opened 2015-06-16
PM / devfreq: governor_cpufreq: Rewrite locking to avoid deadlocks

humberos:
A devfreq governor store in parallel with a cpu freq update can cause
deadlock as shown below.

Assume current devfreq governor is cpufreq, and user tries to change
to some other governor.

| Write to sysfs store_governor | cpufreq driver updating cpu freq |
| --- | --- |
| echo bw_hwmon > governor |  |
|  | takes rcu_read_lock and calls all |
|  | cpufreq transition callbacks for |
|  | PRECHANGE or POSTCHANGE |
|  |  |
| GOV_STOP on governor_cpufreq. |  |
| unregister_cpufreq() accquires |  |
| state_lock mutex. |  |
|  | try to accquire same state_lock in |
|  | cpufreq_trans_notifier(). Blocked. |
| unregister from cpufreq |  |
| transition notifier and wait for |  |
| all rcu_readers to finish. |  |

```
                        Deadlock
```

A similar deadlock can happen with governor change and policy notifier
callbacks.

The state_lock currently protects multiple unrelated critical
sections: registering/unregistering of cpufreq notifiers, read/writing
the device list, and tracking the cpu states and updating device
frequencies. There is no need for register/unregister of the cpufreq
notifiers to be mutually excluded against the other critical sections
using the same lock.

Split state_lock into two locks to protect the register/unregister of
cpufreq notifiers from the rest of the critical sections.

Change-Id: Id06d326748a5cb0c84c4787da5d0910f44eb5c3c
Signed-off-by: Pan Fang fangpan@codeaurora.org
Signed-off-by: Arun KS arunks@codeaurora.org
Signed-off-by: Junjie Wu junjiew@codeaurora.org
Suggested-by: Saravana Kannan skannan@codeaurora.org

### DIFF
--- a/drivers/devfreq/governor_cpufreq.c
+++ b/drivers/devfreq/governor_cpufreq.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2014, The Linux Foundation. All rights reserved.
+ * Copyright (c) 2014-2015, The Linux Foundation. All rights reserved.
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License version 2 and
@@ -52,6 +52,7 @@ struct devfreq_node {
 };
 static LIST_HEAD(devfreq_list);
 static DEFINE_MUTEX(state_lock);
+static DEFINE_MUTEX(cpufreq_reg_lock);
 
 #define show_attr(name) \
 static ssize_t show_##name(struct device *dev,				\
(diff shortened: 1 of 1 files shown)


kholk:
I remember seeing this quite some time ago. I may be confusing it, or it could be a different version of it.

Btw, from what I remember I didn't commit this because it was doing something nasty, like panics, bad device deep sleep, or something else.
Please, check this in any and every aspect of it.

humberos:
@kholk
Yeah! I will test it carefully.
Btw, it is present at LA.BF64.1.2.1_rb2.18
https://github.com/humberos/kernel/commits/LA.BF64.1.2.1_rb2.18

humberos:
It is fine.
Tested on shinano leo. No bad effects so far so good!

PR #338, opened 2015-06-16
[TESTME ON ALL] video: msm: mdss: Support for ULP modes. Code style cleanup.

kholk:
Modify display power on function to support multiple modes and
to use the enumeration for checking the new power mode.
While at it, add the function for LP and ULP modes (LP1/LP2)
for the displays that will support those in the future.

Also, cleanup code style where possible without having impacts
on human readability.


galaxyfreak:
Looks fine on eagle.

jerpelea:
tested on all 
OK

PR #339, opened 2015-06-16
yukon: set the correct screen dimensions

stellirin:
Not sure how Android uses these, but tianchi has wrong values and flamingo
lacks them completely. Fortunately Android functionality seems unaffected.

However Mozilla's B2G uses these to determine screen density. Incorrect (or
missing) values gives incorrect scaling.


jerpelea:
should be checked on all panels

kholk:
It's just for userspace things. Won't change anything kernel-side, safe to merge

stellirin:
Now all yukon devices have them

jerpelea:
please look into all platforms

stellirin:
All rhine / shinano have `somc,mdss-phy-size-mm` but not the qcom ones. However I don't have any issues on sirius, and the mozilla guys don't have any issues on aries / leo so it looks like the qcom ones are not needed for the other devices and nothing else needs to be done.

Kali-:
not sure if that is related to ZU issue on cm, where the button are huge

Kali-:
nvm
was http://review.cyanogenmod.org/#/c/101228/

PR #344, opened 2015-06-23
ARM: dts: amami: fix panels dtsi

Kali-:
partially revert commit e8cdf3e6672 where video mode was
accidentally merged into cmd one
- Addresses black lines on JDI panels


Kali-:
need to be tested

kholk:
First, I didn't accidentally merge video and cmd, that comment in that commit is hilarious.
Second thing, that's a switch to CMD mode, basically.... why can't we just fix video parameters?

Third and last, we actually need to fix video mode, because dynamic switch is coming :)

PR #346, opened 2015-06-27
msm: mdss: fix transitions from low power mode to off

humberos:
Current implementation incorrectly handles the transitions from display
low power modes to display off. This could lead to DSI panel and
regulators not getting properly turned off. Fix this issue by ensuring
that DSI interface is correctly shut down during this transition.

Conflicts:
    drivers/video/msm/mdss/mdss_mdp_intf_cmd.c

Change-Id: I7c20c04d20fd2f443c61b83f229ce4bd2b46f814
Signed-off-by: Aravind Venkateswaran aravindh@codeaurora.org


kholk:
This has to be tested on devices that support low power modes in both kernel and userspace.
Actually, the only ones that support those modes are Kitakami family (Karin, Ivy) and we don't have a valid userspace for those, still.

...Hold on a lil' :)

PR #348, opened 2015-07-08
Update aosp_rhine_honami_defconfig

optimumpr:
Enabled all ciphers, hushes, blocks and random number generation


optimumpr:
This commit should work at least for all rhine devices, possibly more...

jerpelea:
What is the reason ?

Kali-:
also learn "make savedefconfig"

optimumpr:
Better entropy, random number generation, less lag, perhaps? Support for apps providing encryption?

kholk:
We've already got QC drivers for hw backed enc/decryption, did you see QCrypto, QCE etc?

You're enabling software enc/dec, it's simply not right and it will actually slow down the whole thing :)

optimumpr:
Proprietory qc? NSA, anyone? And this is not only for disk encr, but for ssl etc.

optimumpr:
Even if we trust proprietary hardware based random number generation (which we really can't), the best option is to mix it with software generation.  By the way, when I disabled qc crypto modules/drivers, then encrypted data partition in lollipop and then disabled hardware overlays in Developer option, I got rid of any noticeable lag associated with encryption. And it is not a placebo, because the lag came back after enabling hardware overlays.

Kali-:
lag with mdp composition?

optimumpr:
It happens regardless of whether mdp is set to 1 or 0. Only when hw layers are disabled (via init.d, because manual value won't stick after reboot), and qc modules are disabled in kernel, the lag is gone.

Kali-:
kholk: i guest this bug was not affecting 3.10 kernel

Kali-:
i just noticed "CONFIG_SECURITY_*" are also disabled

optimumpr:
That's my mistake.  How should I better fix it? By closing this and sending a new commit? Is there a way to edit this commit? Thanks

erikcas:
Just out of curiousity. Was this actually tested on the 3.10 branch or on the 3.4 kernel that is present in the AOSPME GitHub account?

I tested parts of this commit earlier on amami on 3.10 and I was not satisfied with it

optimumpr:
It wasn't tested on 3.10, but there should be no difference in this between 3.4 and 3.10.  Could you describe your dissatisfaction in more detail?

erikcas:
Cannot really remember if I enabled all of this.

Enabled parts of this in smaller to greater parts. Lesser part, it was slow. No boot after enabling a load of this.
So decided to leave it there.

You really should test it first on 3.10 branch
I see a load of commits on the AOSPME 3.4 branch that are not present in this tree. Without further checking those, it seems incomparable tot me

kholk:
@Kali- No, that bug wasn't affecting 3.10.... but the last time we tested was on the old LA.BF! Anyway I don't see any reason for it to come back...................

@optimumpr If you aren't protecting top secret government informations you don't need to mix that.... and if you are, mixing the RNG and using software crypto engines would still not be secure enough.

While I agree with you that proprietary hardware crypto engines MAY (and I still have my doubts) have something nasty, I have to completely disagree about the enhanced security of software CPU-driven engines as, in this case, those may contain nasty things as well.

Anyway, we're driving this to PURE SCIENCE FICTION, and I think it's just not worth it.

The answer is simple: HW and SW engines are practically the same for security. The only difference is that HW engines are faster.

optimumpr:
@kholk 

While we can argue about the application of different ciphers, the mix between software and hardware in random number generation is the only way to assure true randomness.  Read here what Linus Torvalds is saying.  There was a petition to remove Intel's random number generator from linux kernel and his reason for not removing it was this:  "We use rdrand as _one_ of many inputs into the random pool, and we use it as a way to _improve_ that random pool. So even if rdrand were to be back-doored by the NSA, our use of rdrand actually improves the quality of the random numbers you get from /dev/random"

http://www.theregister.co.uk/2013/09/10/torvalds_on_rrrand_nsa_gchq/

Software random number generation is enabled in Linux kernel by default. Sony, however, disabled those in favor of hardware only...

@erikcas 

As, Kali noticed, I mistakenly disabled a few settings including Selinux and that could be a reason for no boot.  

@jerpelea 

Anyway, are you guys interested in this? Then I will close this and resubmit the commit with SElinux and Security Keys enabled. (I have already made changes in my fork).

kholk:
With --> NSA <-- SELinux, you mean? :)

Linus is right, indeed.... the problem lives elsewhere in an Android phone, though :)
I'd rather continue this conversation privately if you're still convinced of this being good, instead of spamming GitHub........

erikcas:
For me. I did not mistake mentioned settings in defconfigs when testing afaik. I properly enabled selinux and other security settings

optimumpr:
@kholk 

Actually, in my CM based roms, I do disable Selinux and use Tomoyo instead. That's why (since I was doing cut/paste from one of my CM kernels)  those settings accidentally got in here.  

Yes, there are settings in /system/vold where you can increase iterations from 2000 (which is pretty weak) to 5/7/10 and that improves random number generation, but if you leave software generation disabled in kernel, then you are disabling those other sources of inputs Linus was talking about.

optimumpr:
In the same article:

"Random-number generation for the kernel space were implemented in 1994 by Theodore Ts'o using secure hashes instead of ciphers. As Tso wrote here following the latest selectively released information by journalists allied to Snowden:

```
I am so glad I resisted pressure from Intel engineers to let /dev/random rely only on the RDRAND instruction... Relying solely on the hardware random number generator which is using an implementation sealed inside a chip which is impossible to audit is a BAD idea"
```

optimumpr:
There is a new pull request with corrected entries:  https://github.com/sonyxperiadev/kernel/pull/351

PR #351, opened 2015-07-16
Update aosp_rhine_honami_defconfig

optimumpr:
Enabled disabled by default ciphers, hushes and random number generation. In fact, Sony relies only on hardware based random number generation, which is unsafe. Enabling software number generation mixed with hardware provides optimum security


optimumpr:
This is a corrected pull request related to https://github.com/sonyxperiadev/kernel/pull/348

I am going to be closing pull request 348 accordingly.

optimumpr:
I have compiled the latest 3.10 for Honami and there is no visible slowdown.  In fact, the interface may have become more fluid because of better random number generation...

optimumpr:
Here is the boot.img.  And by the way, camera still does not work...
http://d-h.st/2zBX

erikcas:
or it might be a result because of the additional adj rbtree commits.

Did you expect camera to work based on only the kerneldriver commits?

optimumpr:
"Did you expect camera to work based on only the kerneldriver commits?"

My expectation was based on the following statements: "camera enabled" and "welcome Sony camera."

jerpelea:
camera sensors are now detected by kernel

PR #356, opened 2015-07-18
eagle: fix performance

erikcas:
In current situation M2 suffers serious performance drops on various additional added apps from market. For expample, opening Facebook and Chromium webbrowser do practically freeze the device. Updating Facebook from market takes about 2 hours before it gets cancelled by OS.

I traced this behaviour back to lmk. Reverting the adj rbtree commit did make the behaviour dissapear.
Discussed this with @kholk outcome was that reverting is no option.

So I did a little crusade to the origin of this commit, and found some additional commits coming with the original implement of this driver. Applying these commits to our tree solves the behaviour as well.

Tested for unwanted side-effects on msm8974 on z1c. I could not notice any.

### DIFF
--- a/drivers/staging/android/lowmemorykiller.c
+++ b/drivers/staging/android/lowmemorykiller.c
@@ -570,14 +570,16 @@ DEFINE_SPINLOCK(lmk_lock);
 struct rb_root tasks_scoreadj = RB_ROOT;
 void add_2_adj_tree(struct task_struct *task)
 {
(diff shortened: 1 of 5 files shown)


kholk:
Good job! 

Btw, I feel like it's necessary to say the reason why revert was no option. 
Simply, red/black trees are really efficient when dealing with data like the one being analyzed by LMK.

This implementation is way faster than the old one. 

Now we may also go on with testing the new rbtrees implementation, that should result in way less overhead while scanning for those processes to be maybe killed.

PR #372, opened 2015-08-06
mdss: dsi: panel_driver: Ignore EBUSY on gpio_request

andersson:
The gpio lifecycle in this driver is a mess, this hack makes the code
ignore any -EBUSY errors so that it still will reset and bring the
display up.

Signed-off-by: Bjorn Andersson bjorn.andersson@sonymobile.com

### DIFF
--- a/drivers/video/msm/mdss/mdss_dsi_panel_driver.c
+++ b/drivers/video/msm/mdss/mdss_dsi_panel_driver.c
@@ -322,19 +322,19 @@ static void mdss_dsi_panel_bklt_dcs(struct mdss_dsi_ctrl_pdata *ctrl,
 
 static int mdss_dsi_request_gpios(struct mdss_dsi_ctrl_pdata *ctrl_pdata)
 {
-	int rc = 0;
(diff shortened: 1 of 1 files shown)


kholk:
I completely agree and I'm sorry for this, but eh... It was practically the only way to make every damn display to work on one driver. 

About the EBUSY, it can be fixed in a proper way but won't come soon (as I'm on vacation and I was anyway planning other things). So yes, merge, merge, merge! 

.... And ty Bjorn!

andersson:
I set out to fix it in a "proper" way, but after writing panel drivers for mainline I got bored with the caf/mdss mess really fast.

As this is being completely redesigned for future releases (read mainline) let's just keep the patches as small as possible.

kholk:
I completely agree. This kernel needs to work as clean as possible but this stuff is anyway messy....and it has always been since the very beginning. 

I understand your feelings about this, I've been working on that for a little while too (as you may know).

jerpelea:
Still needs some testing before mergr

PR #373, opened 2015-08-14
dt: amami: fix blacklines on novatek jdi panels.

erikcas:
copy values from stock for jdi variant. But keep dsi_video_mode.
Might be ugly but takes care of the black lines issue w/o affecting the other panels.

Signed-off-by: Erik Castricum erikcas1972@gmail.com

### DIFF
--- a/arch/arm/boot/dts/qcom/dsi-panel-amami.dtsi
+++ b/arch/arm/boot/dts/qcom/dsi-panel-amami.dtsi
@@ -520,8 +520,7 @@
 				15 01 00 00 00 00 02 26 00
 				15 01 00 00 00 00 02 FF 00];
 		somc,mdss-dsi-init-command = [15 01 00 00 00 00 02 BA 03
-				15 01 00 00 00 00 02 C2 03
-				39 01 00 00 00 00 06 3B 03 02 03 08 04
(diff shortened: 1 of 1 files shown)


erikcas:
Spend two days on this change with warm phone and in full sunlight (circumstances the issue appears). With this in, the issue did not occur anymore

jerpelea:
I will try to reproduce it

erikcas:
To be clear: With ugly, I mean the fix itself. Imo there is no negative result in panels look and feel 
:)

kholk:
Does it work without copying the new porches but only the other diffs?

erikcas:
@kholk I will test that this afternoon and come back on it

erikcas:
Just tested with original porches and a 500W light on it. It seems I can leave the porches untouched. Tot be sure... I expect full sunshine in half an hour that will be the ultimate test.

For now I expect a force push in about an hour

kholk:
The reason why I asked you to try is that, mathematically, the porches ARE correct ;) 

Btw, it's fine. Notify when it's ready. 

Thanks ;)

erikcas:
force pushed. Tested all fine

PR #385, opened 2015-08-31
cpufreq: updates from CAF - LA.BF64.1.2.1_rb2.29

humberos:
1. Save state for entire cluster when the last CPU goes offline.
2. Check current frequency in device driver.

Tested on Shinano Leo.


humberos:

PR #386, opened 2015-08-31
cpufreq: interactive governor updates

humberos:
cpufreq: interactive governor updates from CAF (LA.BF64.1.2.1_rb2.29).

Tested on Shinano Leo.


humberos:

PR #387, opened 2015-08-31
spi: spi_qsd: Reduce latency for non-shared cores

humberos:
Resource aqquistion on a per transfer level incures
a noticable latency penalty.  This change removes
this latency by doing resource aqquisition as part
of runtime-pm on non-shared cores.

Change-Id: I112c10f52342ca7143b2286c72bba49ba7c8e49f
Signed-off-by: Dan Sneddon dsneddon@codeaurora.org
Signed-off-by: Kiran Gunda kgunda@codeaurora.org


humberos:

PR #391, opened 2015-09-02
drivers: clk: update to LA.BF64.1.2.1_rb2.29

humberos:
Tested on Shinano Leo.


humberos:

PR #394, opened 2015-09-02
drivers: devfreq: up to CAF - LA.BF64.1.2.1_rb2.29

humberos:
tested on Shinano Leo.


humberos:

PR #395, opened 2015-09-04
drivers: qcom-sps-dma: fixes

humberos:
1. qcom-sps-dma: Avoid re-allocation of pre-allocated resources
2. qcom-sps-dma: fix a memory leak and improve performance

Also, up to LA.BF64.1.2.1_rb2.29


humberos:

humberos:
Actually dma driver is not used for Shinano Leo. Not sure if it exists on Sony open devices. So reducing diffs between CAF is the main reason of this patch.

PR #397, opened 2015-09-06
i2c-msm-v2: avoid unnecessary DMA init and channel free

humberos:
Ensure that DMA initialization occurs once and
DMA channel API is called in appropriate mode.

Change-Id: I347aad9848737be2d0fd5c5bd4c5f909c7c69e19
Signed-off-by: Naveen Kaje nkaje@codeaurora.org


humberos:

PR #399, opened 2015-09-07
Clean up inline kernel build

stellirin:
We don't need these crazy variables as we build our kernel sanely


stellirin:
This doesn't append the dtb to the kernel yet, but it saves a lot of wasted time because we were rebuilding the .config multiple times, and trying to build modules that we don't have.

PR #400, opened 2015-09-08
Bluesleep pinctrl

BuzzBumbleBee:
Needs testing on msm8974 devices with broadcom


kholk:
Why revert the cleanup?

PR #411, opened 2015-09-19
adapt LMK + fixes to dts

mustermaxmueller:
Im no kernel developer. So please triple check :)


mustermaxmueller:
make new and clean

mustermaxmueller:
make new and clean

PR #415, opened 2015-09-23
Ramoops

Kali-:
i cant test, but buffer start was not in reserved mem 
i guess this error can produce random reboot on device where it was enabled


jerpelea:
Please rebase

PR #416, opened 2015-09-23
arm: DT: fix ramoops buffers

Kali-:
i cant test, but buffer start was not in reserved mem 
i guess this error can produce random reboot on device where it was enabled

### DIFF
--- a/arch/arm/boot/dts/qcom/msm8226-yukon_flamingo-mtp.dtsi
+++ b/arch/arm/boot/dts/qcom/msm8226-yukon_flamingo-mtp.dtsi
@@ -18,7 +18,7 @@
                compatible = "ramoops";
                status = "ok";
 
-               android,ramoops-buffer-start = <0x3eae0000>;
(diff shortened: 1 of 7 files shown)


kholk:
3e9e(blah) is not valid for 8974!!!

Also, carefully test this on Yukon devices prior merging. Ramoops should, indeed, be in reserved (and protected across reboots) memory portions.

kholk:
Addition. If you want to fix that, you should modify the reservation addresses :P

Kali-:
<6>[    0.000000] Node android,ramoops-hole memblock_reserve memory 3e9e0000-3eae0000

are u sure it doesnt exist on 8974? i have not changed reservation, just moved the buffer at start of the reservation as it should be
i'm with u about this buffers can be placed better, but they looks completly wrong on every device, on z2 family is disabled (i guess because it crash on boot), on some z3 and yukon is enabled but reservation is on 3e9e0000-3eae0000 and buffers are outside this range and this cause unpredictable results

Kali-:
merged in newer branch

PR #429, opened 2015-10-02
drivers: tty: serial: Clean up SONY serial driver merge

humberos:
a call to msm_hs_disable_flow_control() was replaced with a sequence of register
reads/writes that performed an identical function EXCEPT for changing the RFR
register to RFR_HIGH

However, the subsequent call to change the status register to RFR_LOW was removed.

This mismatch is the cause of Bluetooth reconnect issues

the original change:
https://github.com/sonyxperiadev/kernel/commit/9cdddb2555dcdd45f49107b33e6f0993a640348a

Signed-off-by: Andrew Dodd atd7@cornell.edu
Change-Id: I05840280110bc4b51e72e30bd1c108368a971bdc


humberos:
Not sure if this should be there since it is required for shinano devices. misc-dev maybe?

PR #436, opened 2015-10-14
video: msm: mdss: Fix unused function warning

abhinav-jhanwar:
drivers/video/msm/mdss/mdss_dsi.c:129:12: warning: 'mdss_dsi_labibb_vreg_ctrl' defined but not used [-Wunused-function]
 static int mdss_dsi_labibb_vreg_ctrl(struct mdss_dsi_ctrl_pdata *ctrl,
            ^


kholk:
The function will never be used out of there, hence it has to be static!

PR #445, opened 2015-10-19
Misc dev

alviteri:
related to usb midi


alviteri:
tested on leo and honami

kholk:
Ok! But you pushed to the wrong branch!

alviteri:
@kholk which is the correct....

kholk:
This branch is for commits entirely made by "external" developers, such as you and as well me, mostly modifying QC implementation somehow and for some reason.

Push that change to drivers ;)

alviteri:
thanks @kholk

PR #447, opened 2015-10-22
fix mounts for (internal) sd in marshmallow

erikcas:
patch allows us to properly mount internal sd
External sd is now useable as adoptable and as portable device.

tested on D2403 and D5503

### DIFF
--- a/fs/fuse/inode.c
+++ b/fs/fuse/inode.c
@@ -891,6 +891,11 @@ static void process_init_reply(struct fuse_conn *fc, struct fuse_req *req)
 				fc->async_dio = 1;
 			if (arg->flags & FUSE_WRITEBACK_CACHE)
 				fc->writeback_cache = 1;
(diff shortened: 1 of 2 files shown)


erikcas:
in order to test adoptable sd feature, please drop "noemulatedsd" from the external SD card configs in fstab (and offcourse add "encryptable=userdata" as in the @alviteri pullrequests)

When testing on D2403 I was not able to make the external sd adoptable when "noemulatedsd" was set

alviteri:
good job @erikcas ...... and with the problem about memory I tested adoptable and yes the device encrypt your exSD for make it work as SD

jerpelea:
Please push also to misc-up branch

jerpelea:
And 1.2.1 branch

erikcas:
Alin, will do later today. Currently handling a huge escalation @ work

PR #455, opened 2015-11-04
media from angler kernel

alviteri:


alviteri:
@jerpelea same for this ..... media on new kernel is updated

PR #456, opened 2015-11-04
arm:dt:msm8994 Correct regulator timming

mmaret-geny:
Documentation describe slew-rate as "time in us it takes for the regulator to change votlage value in one step".
2s seems a lot and qpnp-lab-slew-rate is only 5000us.
So correct it as if a 1000 factor was added.

This reduce the kernel boot time of the karin windy by 80s
Tested on karin windy only

### DIFF
--- a/arch/arm/boot/dts/qcom/msm-pmi8994.dtsi
+++ b/arch/arm/boot/dts/qcom/msm-pmi8994.dtsi
@@ -470,7 +470,7 @@
 
 				qcom,qpnp-ibb-min-voltage = <1400000>;
 				qcom,qpnp-ibb-step-size = <100000>;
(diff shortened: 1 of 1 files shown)


mmaret-geny:
qpnp-ibb-slew-rate is used in drivers/regulator/qpnp-labibb-regulator.c in the qpnp_ibb_regulator_set_voltage() function.
This function calculate the diff between the current voltage and the asked one.
Then the number of step to get to this voltage is calculated (40 for the karin windy).
After it waits for the voltage to regulate so it waits 40*qpnp-ibb-slew-rate. (so 80s)

jerpelea:
Please cherry-pick to misc-dev branch

PR #459, opened 2015-11-07
kitakami: karin: Update panel backlight name and it levels

humberos:
Change-Id: Ifc26446a909c02fc10a9b60c8f3d6a33798e4b89

### DIFF
--- a/arch/arm/boot/dts/qcom/msm8994-kitakami_karin_common.dtsi
+++ b/arch/arm/boot/dts/qcom/msm8994-kitakami_karin_common.dtsi
@@ -127,12 +127,12 @@
 			compatible = "ti,lp8557";
 			reg = <0x2c>;
 			status = "ok";
-			linux,name = "lp8557";
+			linux,name = "wled:backlight";
 			linux,default-trigger = "bkl-trigger";
 			mode = "register based";
 			chip_name = "lp8557";
-			max_br = <0x0FFF>;
-			init_br = <0x666>;
+			max_br = <0x0FF>;
+			init_br = <0x66>;
 			somc-s1,br-power-save = <0x83>;
 			slope_reg = <0x00>;
 			dev-ctrl = <0x01>;


humberos:
@jerpelea 
We need @kholk 's ninja sword on lp8557 driver.
Or I can fix it by changing userspace library.
here:
https://github.com/sonyxperiadev/device-sony-kitakami/blob/master/liblights/lights.c#L225

and by creating new brightness overlays for karin.

humberos:
@kholk :
This patch changes brightness sysfs entries to match with others kitakami variants.
But now the main differences are brightness levels.
We need to change lp8557 driver that should be working with 102 and 255 (min/max) levels instead of default 1638 and 4095.

jerpelea:
https://github.com/sonyxperiadev/kernel/blob/drivers/drivers/leds/leds-lp855x.c

You have a flag to switch to 8 bit brightness
Pr for drivers branch!

humberos:
Yeah! I did it but it is not working as expected.

https://github.com/sonyxperiadev/kernel/blob/drivers/drivers/leds/leds-lp855x.c#L622
I've changed it to "true" .

I think the hardware (lp8557) uses 12bits.

Also I've tried by adding bl_scale to DT entries but I had no look.

kholk:
_pulls out his sword_
I will kill the daemons that are haunting this driver on Monday ;)

Jokes apart, I don't really understand this commit..... LP855x chip is being used on other devices as well, and their names were in line with Karin.
If we want to go in this direction, you should as well examine the Shinano tablets (castor and scorpion) and do whatever it takes, either revert this commit or change the names on those as well.

In the meanwhile, I will in any case examine the 8bit problem (as I said, on monday).

jerpelea:
We need this on all devices!

kholk:
Unfortunately, this commit is wrong.

Proper fix in two commits, one on platform, one on drivers.

PR #460, opened 2015-11-09
arm: DT: Kitakami: Karin: Correct backlight name and levels

kholk:
This commit is a partial revert of 5fc2d22 by @humberos.
The backlight shall be registered as lp8557 because we have to
update its level differently.

We are allowed to use the Linux trigger "bkl-trigger" for this,
otherwise we can access it directly from (discouraged) its
directory in class sysfs.

On Qualcomm platforms, we can update it through MDSS backlight
triggers: this will allow display correction algorithms to adjust
the displayed image's RGB levels considering the current backlight
level while calculating the color deltas.

### DIFF
--- a/arch/arm/boot/dts/qcom/msm8994-kitakami_karin_common.dtsi
+++ b/arch/arm/boot/dts/qcom/msm8994-kitakami_karin_common.dtsi
@@ -127,12 +127,12 @@
 			compatible = "ti,lp8557";
 			reg = <0x2c>;
 			status = "ok";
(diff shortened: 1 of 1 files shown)


humberos:
Oh! Using lp8557 as linux,name (sysfs entry) then I think the userspace requires updates.
https://github.com/sonyxperiadev/device-sony-kitakami/blob/master/liblights/lights.c#L62-L94

Or, /sys/class/leds/lcd-backlight/brightness will be used as generic sysfs entry for lp8557.

PR #463, opened 2015-11-16
Add ZRAM

humberos:
Finally it is present at AOSP.
It requires defconfig and init.rc changes to enable ZRAM.
Tested on msm8994 karin.


kholk:
Test it on devices on which it may be ok to use it (like Yukon, for example).....
...I don't think it's that useful to use it on a device with 3GB RAM!

humberos:
@kholk it is useful even on devices that have 3GB. 
I think Linux based systems expect a swap area and ZRAM is the best solution for I/O latency.
Try it on your karin then you'll see.
Btw, angler (Nexus 6P) is using it.

kholk:
It's not really true.... I mean, Linux systems do not expect a swap area "by default". It's just common sense to use a swapfile (for example, on Desktop distributions of Linux with X11 et al) because usually you don't have enough RAM for all you do.
Embedded systems, on the other side, do normally work without swapfiles and the kernel is ok with that.

In any case, if you say that there's some improvement on Karin, because of my knowledge, I can believe you if you can provide me some numbers (use benchmarks -- commandline style, the serious ones, not antutu etc).

Actually, the way ZRAM works is well known.. and it's also well known that compression algorithms will create a lot of overhead, being those CPU intensive tasks, especially on an embedded-style system that is using a low power processor with slow cores (compared to x86 ofc -- not saying that 8994 isn't fast as hell).

humberos:
Please, You don't need to prove your knowledge. I am just following your ideas for a long time. I am sorry if I did it.
But I think exists a reason why Google enabled it on devices with 3GB ram. ;)

kholk:
Sorry, don't misunderstand me, I didn't mean to be rude and/or offend you in any way!

Yes there may be a reason why Google did that... and that's the reason why I'm asking you some numbers ;)

jerpelea:
can someone post a proof of usage on nexus

humberos:
@jerpelea 
init.rc :
https://android.googlesource.com/device/huawei/angler/+/android-6.0.0_r26/init.angler.rc#43

zram parameters:
https://android.googlesource.com/device/huawei/angler/+/android-6.0.0_r26/init.angler.rc#37

fstab:
https://android.googlesource.com/device/huawei/angler/+/android-6.0.0_r26/fstab.angler#28

Tommy-Geenexus:
@kholk
The complete Z5-series uses it too (+ KSM).

Tommy-Geenexus:
One could pull ZRAM from the Z5 Source drop as there are some Sony-modifications inside.

erikcas:
@Tommy-Geenexus https://github.com/erikcas/android_kernel_sony_msm/commits/sumire/zram
have fun with it

humberos:
Hi!
IMHO I think we could use AOSP/CAF patches (just preserving commit history) then applying Sony diffs over them after. What do you think?

humberos:
Btw, I've just checked Sony sources but I'd say that AOSP is more updated with zram features.

NVM, Sony wins!!!
We will need to work on new Sony sysfs.

Tommy-Geenexus:
@erikcas 
@humberos 
I just talked to Alin, ZRAM will not be added.

jerpelea:
please PR an clean ZRAM. I will not merge it for now but is a good reference for other devs

kholk:
IMHO Sony loses... 
If we need some clean ZRAM, I think that this PR would be the best one, as the drivers do come from  "updated" mainline code which, like the original committer says, contains commits that are largely improving its performance.

In the end, this is the PR that should be taken as reference.

@Tommy-Geenexus It's a pity, because KSM is one of the worst things ever: performance degradation with it is really large on ARM/AArch64 systems!

Tommy-Geenexus:
Yeah, that's why i disabled it on my Stock Kernel Builds.
Memory gains are neglible, but it causes higher CPU-usage and noticeable battery-drain.

I hope it was just a try to compensate for LP mem leaks and will be disabled in M.

tigermeng:
why stock rom turned on ZRAM (and other features)? because you can't let user see the brand new phone is running so many shits on it, and provides <100M free ram :p    

while open devices is clean, free ram is large.  but who knows how many apps an user might install.  i like the CM way, provide ZRAM but let user decide if turn on (e.g. in their settings menu)...   by default is off i remember...

humberos:
@tigermeng 
Yeah it is a good idea which remindes me that I already did it long time ago.
https://github.com/omnirom/android_device_samsung_p1-common/commit/525cc2dc10c965f8001b56f4142020c9d992b54f

Tommy-Geenexus:
I fully agree with you @tigermeng, and i don't question ZRAM at all, i find it useful on a 2GB RAM device like suzuran.

But KSM is another story, i see next to none benefits (very few MBs RAM gained) but noticeable drain. Thus i question the sense of enabling it.
(Thats why stuff like UKSM and PKSM was coded to majorly improve  over the normal KSM and greatly reduce CPU-usage).
KSM is very conservatively adjusted, so it may be better to completely drop it (or make it switchable via Settings).

kholk:
@Tommy-Geenexus Useful on a 2GB RAM device you say? Suzuran is 64-bits hence RAM consumption is higher compared to 8974 and 32-bit friends, but also don't forget that the screen is smaller and less pixels to render also means less RAM consumption in most of the cases.

I have actually tried to open and put in background >15 (or maybe was it 20?) apps on my Z5 (ok, I know, 3GB -- but see the display thing) and nothing was closed. Also this was on Sony stock, which has a REALLY higher memory footprint compared to our clean and light AOSP.

I sincerely have doubts on the usefulness of ZRAM on our devices (if not, MAYBE, Yukon -- but CPU usage..?! to evaluate...) but if anyone proves me that it's actually useful and doesn't hurt performance with real numbers, I'll give my personal OK to merge.

Tommy-Geenexus:
If one is a heavy multitasker like me, ZRAM can make that tiny bit of difference on a 2GB RAM 64Bit-device.

But i don't think it is necessary on AOSP due to smaller memory footprint (and on 2GB+ 32Bit devices either way unneeded).

jerpelea:
devices already have enough RAM

PR #468, opened 2015-11-18
shinano: fix Front camera mount angle

JulienBolard:
New camera binaries makes the Z3 front camera upside down.

Signed-off-by: Julien Bolard jbolard@genymobile.com


kholk:
All Shinano devices should have front camera upside down!!

Please fix all

JulienBolard:
I only have a Z3 6603 to test that. I can modify the other but can't do the tests.

kholk:
No worries, it's a known bug on all, we simply forgot to update the DTs. Do it :)

PR #487, opened 2015-12-01
flamingo: DT: merge typo fix

tigermeng:
fix i2c@9925000/9926000 and 9927000 as sub node &soc {}

without the fix, the following err kmesg was seen flamingo:
qup_i2c f9925000.i2c: Missing 'qcom,master-id' DT entry
qup_i2c f9927000.i2c: Missing 'qcom,master-id' DT entry


tigermeng:
close this PR...  
because we might be able to to safely delete those lines,  instead of moving them...  they might be useless lines...

PR #488, opened 2015-12-01
flamingo: DT: cleanup useless entires

tigermeng:
first of all, the removed lines are in wrong position,
because i2c@9925000/9926000 and 9927000 should be sub node &soc {}
seems to be a merge typo here.   so, removing wrong lines brings no impact.

secondly, i can't find similar lines in CAF reference, lines should be useless. 
so i deleted them instead of moving them into &soc{}


tigermeng:
"vdddsi" is stranger to me,  i deleted and verified on flamingo brown, see no impact
i double check by grep the code, found no src code ...  

but please triple confirm

PR #500, opened 2015-12-07
USB Extensions and MHL

humberos:


humberos:
Depends on:
- https://github.com/sonyxperiadev/kernel/pull/499
- https://github.com/sonyxperiadev/kernel/pull/501

PR #501, opened 2015-12-07
MHL

humberos:


humberos:
Please, don't merge it.
It requires fuel gauge and smbcharger extensions updates.

PR #509, opened 2015-12-10
[1.1] Fix Shinano MHL

kholk:


humberos:
Is it required on kitakami devices too?

kholk:
No, it's not

PR #495, opened 2015-12-05
ARM64: config: Add CONFIG_SONY_USB_EXTENSIONS to Kitakami devices

humberos:
Signed-off-by: Humberto Borba humberos@gmail.com
Change-Id: Ib992f154710dcb6ccae4658910797546d4ae7d36

PR #511, opened 2015-12-10
[1.1] arm64: configs: Kitakami: Update configs for new USB and MHL

kholk:
Now we have Sony USB extensions for smbcharger and fuelgauge...
and for MHL as well, which uses a new driver, config-called
CONFIG_MHL_SIMG_SONY, which is also responsible for creating
a supply called 'ext-vbus', strictly needed for the new USB
to work.


humberos:
@kholk how about kanuti devices?
Shall I get MHL options enabled too?
It was missed on 1.1 and 1.2.2 trees.

humberos:
nevermind:
https://github.com/sonyxperiadev/kernel/pull/495

PR #521, opened 2015-12-13
net: wireless: bcmdhd: Fix wifi for scorpion

chirayudesai:
commit ec3bcce2afb7771a94f849a4ed1d06b4ed41310c
"net: wireless: bcmdhd: Fix older firmwares support for BCM4354"
removed this check for scorpion or scorpion_windy, but now
that those two are unified as scorpion, the check failed.
Remove the windy check to fix wifi.

Change-Id: I01823a2eceec8f301ee8958873594ce9c1b7a463


jerpelea:
Not really.you still

PR #522, opened 2015-12-13
net: wireless: bcmdhd: Set random MAC generation for Kitakami devices…

humberos:
… as default

This change fixes MAC address generation for all Kitakami devices that aren't
using /sys/devices/platform/bcmdhd_wlan/macaddr (wlan sysfs path).

Also avoid this error during initalization:
<3>[    9.260469] somc_wifi_get_mac_addr: Failed to get information from file /sys/devices/platform/bcmdhd_wlan/macaddr (-2)

Change-Id: I50412a74cf3565cbc02772416c0603c025c8d48b

### DIFF
--- a/arch/arm/vfp/vfpmodule.c
+++ b/arch/arm/vfp/vfpmodule.c
@@ -731,9 +731,6 @@ static int __init vfp_init(void)
 {
 	unsigned int vfpsid;
 	unsigned int cpu_arch = cpu_architecture();
-#ifdef CONFIG_PROC_FS
-	static struct proc_dir_entry *procfs_entry;
-#endif
 	if (cpu_arch >= CPU_ARCH_ARMv6)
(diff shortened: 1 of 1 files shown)


humberos:
Please don't merge it until Z5 series confirmation.

humberos:
ping @erikcas
ping @chirayudesai
ping @alviteri

chirayudesai:
Not being able to read from the file that early is fine, -2 = ENOENT, the file isn't created that early.

The code will fallback to the random mac anyways, even if the file doesn't exist or can't be read (or is empty after my last commit) - see all the 'goto random_mac's

humberos:
Thank you @chirayudesai !
Yeah! I saw your patch and I did this PR based on it, btw.
IMHO I just think totally unnecessary to handling rules for a file that we know it doesn't exists.
But anyway you're right.

PR #526, opened 2015-12-15
arm: DT: Sony: Fix modem initialization on Shinano

humberos:
Signed-off-by: Alexandre Lissy lissyx@lissyx.dyndns.org


humberos:
It is required on Shinano Leo: Kernel 1.2.2 running android L (5.1.1).

PR #530, opened 2015-12-20
mdss dsi panel updates

humberos:
Tested on shinano leo (msm8974).


humberos:
Please, don't merge it.
We need more tests on others platforms.

Ping:
@kholk
@erikcas
@alviteri

humberos:
it needs more love.

PR #532, opened 2015-12-23
Backport SECCOMP

stellirin:
git merge is still fine


kholk:
One super-duper-smart question: why would we need SECCOMP?

stellirin:
Mostly for my convenience when testing on FxOS :wink: We don't need to enable it in the defconfigs here.

PR #533, opened 2015-12-23
Enable SECCOMP on all devices

stellirin:


stellirin:
Not necessary, Kconfig already selects it

PR #536, opened 2016-01-03
Change some governor attriutes

stellirin:
We don't have QCOM's proprietary PowerHAL so to change any governor
settings we need to loosen the attributes so that we can add some
advanced governor switching to our Simple PowerHAL.

Change-Id: I5702e86c08a6708967c8bbb8689ac84115b4bded


stellirin:
We can't do this from init scripts because when we change governors these files pop in and out of existence, so when the come back they have default 0644 root root perms.

PR #540, opened 2016-01-07
Clearpad fixes for Kitakami

Tommy-Geenexus:
This is the first patch-set. Function retval checking is largely broken too, will look at it when i'm free.

### DIFF
--- a/drivers/input/touchscreen/clearpad_core.c
+++ b/drivers/input/touchscreen/clearpad_core.c
@@ -4699,7 +4699,7 @@ static int clearpad_pm_resume(struct device *dev)
 #endif
 	(void)(rc ? rc : clearpad_resume(&this->pdev->dev));
 
-	return 0;
+	return rc;
 }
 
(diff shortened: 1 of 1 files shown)


Tommy-Geenexus:
@kholk i see the driver is outdated compared to the one present in the stock kernel.
Perhaps one should pull the latest stock clearpad version.

kholk:
We are spinlocking prior reading the dev_busy variable to make sure we get a correct reading: if you lock AFTER that, you're creating rare-yet-possible race conditions.

About the other 3 commits, those are OK for me.
Remove the locking optimizations!

Tommy-Geenexus:
Yep, i know that. I saw the early noirq suspend does not use spinlocks at all to perform the same check, thus i figured it'd be ok to protect relevant bools only, as conditions get checked either way (which my testing confirms). Also , it's unnecessary to hold the lock for printing info messages.

I made these patches based on a newer stock driver which is slightly different though.

Btw, have you seen the broken function retval checks i mentioned?

PR #543, opened 2016-01-11
arm64: crypto: Updates from AOSP

humberos:
These changes was incorporated into AOSP@3.10 tree (from Linaro team).
This patchset enables some optimizations in recent Cortex-A5x cores that improves it performance.

### DIFF
--- a/arch/arm64/crypto/Kconfig
+++ b/arch/arm64/crypto/Kconfig
@@ -27,20 +27,19 @@ config CRYPTO_AES_ARM64_CE
 	tristate "AES core cipher using ARMv8 Crypto Extensions"
 	depends on ARM64 && KERNEL_MODE_NEON
 	select CRYPTO_ALGAPI
-	select CRYPTO_AES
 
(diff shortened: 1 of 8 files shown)


kholk:
Is this merged on android-3.10 for MSM or generic?
May be breaking hardware crypto... did you check?

humberos:
https://github.com/android/kernel_msm/commits/android-msm-angler-3.10-marshmallow-dr/arch/arm64/crypto

humberos:
BTW, this aosp branch is based on CAF BF64 1.2.1 and up to 1.2.9.   I think 1.2.9 will not be public  released from CAF  is it?

Tommy-Geenexus:
@humberos 

https://github.com/torvalds/linux/commit/08c6781cfaa196d4b257ec043c17e8746d9284e3
This is the correct fix for the priority issue (altough i think its fine as long as beeing under 300).

I tested these patches on Stock (which uses HW-Crypto) and nothing broke, so this should be just fine.

PR #544, opened 2016-01-11
eagle: add dsds support

erikcas:
I know, we donot have dualsim support (yet) on aosp but we can better be ready ;)

This is tested on eagle D2302; w/o there is no sim detected.


jerpelea:
It is needed also on platform branch

erikcas:
coming up

PR #549, opened 2016-01-16
leo_dsds: added support to leo_dsds variant

ffosilva:
- add support for leo dual sim variant

(ported from sony open source releases)


ffosilva:
Reopened pull request until get a cleaner solution to unify leo & leo_dsds.

ffosilva:
Updated patch. Noew using unified approach.

PR #555, opened 2016-01-18
[1.2.2] msm: mdss: dsi: Add missed CAF implementations

humberos:
Update specific panel driver with CAF implementations.


humberos:
@kholk
Lights HAL and panel brightness transitions between bootsplash and bootanimation are broken after this change.

https://github.com/humberos/kernel/commit/b3ef1be5052290e72dfd4c185ec371c4e9ec96c3

We can't revert it because it is necessary for panel on/off process.
Not sure where is the problem.
I need your help here.
Thank you in advance!

PR #565, opened 2016-01-25
net: ipc_router: security: Grant sensors access to NET users

kholk:
This is needed because we don't want to run everything as root
in our userspace for security reasons.

### DIFF
--- a/net/ipc_router/ipc_router_security.c
+++ b/net/ipc_router/ipc_router_security.c
@@ -302,6 +302,10 @@ int msm_ipc_check_send_permissions(void *data)
 	if (uid_eq(current_euid(), GLOBAL_ROOT_UID))
 		return 1;
 
+	/* ...or it has required NET permissions */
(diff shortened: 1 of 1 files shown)


jerpelea:
This has to be properly checked

jerpelea:
this is a security risk and will be reverted after sensors are up

kholk:
Agreed about the security risk.
Proper fix to be done in userspace.

Kali-:
is not handled by sec_config ?

kholk:
Well, first of all, the destination has rules defined, so it's detected as server.... second thing, the defined security rule defined "somewhere" appears to be wrong, because:

```
    if (!gid_valid(rule->group_id[i]))
        continue;
    if (in_egroup_p(rule->group_id[i]))
        return 1;
```

That code is self-explainatory.

jerpelea:
this is why we should properly fix it

PR #570, opened 2016-01-27
default cache size

Kali-:
imho, this must be set for device in userspace


Kali-:
wrong branch ... again ...

PR #571, opened 2016-01-27
Revert "mm: Increase MIN and MAX default readahead sizes"

Kali-:
imho, this must be set for device in userspace

### DIFF
--- a/include/linux/mm.h
+++ b/include/linux/mm.h
@@ -1624,8 +1624,8 @@ int write_one_page(struct page *page, int wait);
 void task_dirty_inc(struct task_struct *tsk);
 
 /* readahead.c */
-#define VM_MAX_READAHEAD	1536	/* kbytes */
-#define VM_MIN_READAHEAD	32	/* kbytes (includes current page) */
+#define VM_MAX_READAHEAD	512	/* kbytes */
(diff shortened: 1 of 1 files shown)


kholk:
This is valid for the majority of boards/devices with MSM89226/8974 and higher.
Special exceptions can be made in userspace just normally... devices which MAY need a different value can be the worst 8226 ones with very low RAM, such as Eagle and Seagull.

Also, this commit is effective on "external memory" as well (such as sd cards), which would require other lines to execute in userspace (android-speaking: init.rc), hence raising boot time.

humberos:
So I think we can remove readahead handler from init.[platform].rc

humberos:
I mean,
https://github.com/sonyxperiadev/device-sony-shinano/blob/master/rootdir/init.shinano.rc#L99-L101

Tommy-Geenexus:
@kholk

https://github.com/Tommy-Geenexus/android_kernel_sony_msm8994_suzuran_5.x/commit/15ab1c0a58df12146340493f8153ac672dbfe8a3

The readahead is settled too high. Anything above 128 kbytes causes performance regressions, as described in above commit and confirmed by my testing experience on various devices.

PR #575, opened 2016-01-29
[1.2.2] drivers: misc: nfc: Add pn547 pinctrl support

humberos:
Create a duplicate entry for nfc pn547 driver.
This is for device that uses pinctrl.

This solution will reduce sources changes and also
it will give an easier possibility to be updated in the future.

The driver was extracted from 32.0.A.6.x kernel tree.

Change-Id: I1d7a630a94da4646aa950ec7487f264297c6d8f2

### DIFF
--- a/drivers/misc/Makefile
+++ b/drivers/misc/Makefile
@@ -19,7 +19,11 @@ obj-$(CONFIG_LKDTM)		+= lkdtm.o
 obj-$(CONFIG_TIFM_CORE)       	+= tifm_core.o
 obj-$(CONFIG_TIFM_7XX1)       	+= tifm_7xx1.o
 obj-$(CONFIG_PHANTOM)		+= phantom.o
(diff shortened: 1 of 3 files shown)


humberos:
It is a clean implementation for NFC pinctrl version for kitakami devices.
Kernel driver is ok but user space.

User space logs: http://paste.omnirom.org/view/5ff24e7d
- 01-29 18:33:49.616  3715  6311 E NxpTml  : _i2c_open() Failed: retval ffffffff

chirayudesai:
Uh, the stock driver won't work with the HAL we have available since it doesn't have an ioctl interface at all.
That is why I had to adapt a driver from another device to work on 8974 early on.

humberos:
IOCTL: https://github.com/sonyxperiadev/kernel/pull/576

PR #588, opened 2016-02-05
4 kb

Kali-:


kholk:
Erm, sorry for the eventual ignorance..... I don't see benefits in 4k chunks for IOMMU.

Graphics, which is basically vector drawing on Android 2DHW (which should be just a very small chunk of memory transfer per every draw), we have 3DHW which would benefit from this a lot, OK.

And about everything else? Why?
Are we sure we want to force 4K chunks allocation? Pages should be a lot smaller in like 80% of the daily use cases, this will logically create some overhead: do we really want this?

PR #592, opened 2016-02-07
arm64: config: remove unneeded stuff

oshmoun:
reduce size of the kernel by removing stuff that aren't used (qcom liquid, usb ethernet, vifs debug etc..)


humberos:
@jerpelea ,
btw, I wondering why did we do not use an auto-generated .config with all enabled/disabled stuff. IMHO it is so easier to check and maintain.

chirayudesai:
@humberos I would say that .config != defconfig, and so they should be used like that.
Not having the .config file here directly also means that you need to make savedefconfig to get this file, and that ensures a build with the changes which might cause any error.

These configs are quite similar anyway, with the differences being in only a few components (aka very few lines), so we can just usually do the change for one, test that, and apply it to the rest via the same patch. I know because I've done that many times :)

PR #598, opened 2016-02-11
[1.2.2] kitakami: karin: Fix brightness

humberos:
Light HAL unification requires same levels on all devices.
Those reverts bring back max brightness to 4095 for lp8557x driver


humberos:
Hold on!
Due to /sys/class/leds/lcd-backlight/max_brightness still 255 .
It requires more work.
Checking DT values.

PR #600, opened 2016-02-11
[1.2.2] qseecom updates (4.9)

abhinav-jhanwar:


kholk:
We don't really need watchdog_v2 updates. This will only increase battery consumption for no real reason, as even not "stock kernels" from any manufacturer shouldn't have those commits.

About qseecom, make sure it's not breaking the encryption!

abhinav-jhanwar:
I've tested encryption and removed the watchdog updates, for some reason github isn't updating the commits in PR. I'll just create a new PR :)

PR #604, opened 2016-02-12
cpuquiet: use original config naming

Kali-:


Kali-:
fix a small typo in rqbalance which prevent it to be set as default cpuquiet governor if other governors is enabled

PR #609, opened 2016-02-14
tianchi: leds-lm3630_ng: scale 1-255 brightness range to 128-255

stellirin:
This is a hack, if you can see where we adjust the actual current based on this value then please tell me! (I can't see it.)


jerpelea:
ugly but works

kholk:
#ifdef HALF_RESOLUTION
    value_scaled = (value + 255) / 2;
#else
    value_scaled = value;
#endif

ofc, this is just to make it evident in the source!

PR #617, opened 2016-02-19
karin: DT: panel: Fix backlight brightness

humberos:
this patch change mdss@BL max brightness to 255 (8 bits).

Change-Id: I7c6e1b65e1f4bbdc1ad34ffb97f7e655dba3a2bf


humberos:
I don't know if others devices/platforms requires this but for sure it is required for Karin.

PR #618, opened 2016-02-22
update and fix to cpuquiet framework

Kali-:


stellirin:
tested, working well, no regressions. cpuquiet sysfs are now useful

stellirin:
I've been experimenting with nr_cpus and if you change the number too quickly it can cause CPUs in the BIG cluster to get stuck at max freq.
screenshot_20160301-112123

PR #620, opened 2016-02-22
include/fuse.h unbreak build

erikcas:
currently include/fuse.h in device-sony-common breaks build on AOSP

system/core/sdcard/sdcard.c: In function 'handle_opendir':
system/core/sdcard/sdcard.c:1369:8: error: 'struct fuse_open_out' has no member named 'padding'

adjust kernel to AOSP and update kernel-headers in device


erikcas:
see also https://github.com/sonyxperiadev/device-sony-common/pull/18

PR #622, opened 2016-02-23
Revert "Revert "msm_mdp: Fix header for Android system build""

erikcas:
this original fixed builds for msm8226/msm8974 for the msm_mdp.h header, but was reverted in the latest aosp kernelupdates from fxpdev.

If this is accepted, the msm_mdp.h header in common needs to be updated to.
Currently, that header breaks build for msm8974 (tested) and I presume it will break msm8226 too. (not tested that yet):

https://gist.github.com/erikcas/dbd9830228719f777168

I think the alternative is not likely (build 8994 displayhals for all platforms) but one never knows


Kali-:
need to be merged wiht https://github.com/sonyxperiadev/device-sony-common/pull/20 to fix build

PR #624, opened 2016-02-26
mmc: add support for deferred resume

oshmoun:
suzuran developed a regression after fixing wifi, where it crashed after/during going to sleep.
it seems some ifdef blocks for deferred resume support were missing, this commit adds them back.
maybe it's possible to slim down this commit, as it might be the case that not all ifdefs are needed.


kholk:
Thumbs up!

PR #629, opened 2016-02-28
cpuquiet: Set max CPUs from Power and Thermal profiles

stellirin:
There are two reasons we want to limit the number of active cores
- Power saving profile
- Thermal throttling

These two are independent, but with only a single sysfs to control
the number of active CPUs we can easily find that a power profile
has overridden more restrictive thermal limits.

This change adds two sysfs, one for power and one for thermal.
nr_max_cpus is now read only, taking its value from nr_power_max_cpus
unless nr_thermal_max_cpus is smaller.

### DIFF
--- a/drivers/cpuquiet/sysfs.c
+++ b/drivers/cpuquiet/sysfs.c
@@ -43,6 +43,9 @@ struct cpuquiet_dev {
 	struct kobject kobj;
 };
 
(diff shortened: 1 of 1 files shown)


rcstar6696:
Works on sirius and castor_windy :)

PR #639, opened 2016-03-09
[1.2.2] Satsuki 4K awesomeness

kholk:
video: mdss: Add logic to reset dual-display panels (4K awesomeness)

On dual-display panels (read: 4K displays found in satsuki) we
need to reset ONLY the MASTER one and NEVER the SLAVE.

Dual display is only an hack for panels that needs to be connected
to TWO DSI channels because of bandwidth issues.


kholk:
DON'T MERGE!!!!!!!!!!!!! NEEDS TO BE TESTED ON ALL DEVICES!!!!!!!!!!!

PR #641, opened 2016-03-14
yukon: camera: fix clocks

erikcas:
In previous yukon camera commit I obviously forgot to hire an accountant to count the correct number of clocks in DT.
Also correct clocknames in DT to match cameradriver and add absent clocksetting to clocks-8226.c

@kholk can you please be so kind to double check this?


kholk:
Looks right. We were really missing it.... and yes, that clock naming differences between 8226 and ... all other chips is SICK.

Well done :)

PR #642, opened 2016-03-14
BFQ-v7r8-3.10.8+

mustermaxmueller:
add BFQ I/O scheduler

### DIFF
--- a/block/Kconfig.iosched
+++ b/block/Kconfig.iosched
@@ -61,6 +61,27 @@ config CFQ_GROUP_IOSCHED
 	---help---
 	  Enable group IO scheduling in CFQ.
 
+config IOSCHED_BFQ
+	tristate "BFQ I/O scheduler"
+	default n
+	---help---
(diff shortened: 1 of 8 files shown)


kholk:
BFQ is heavy. Really. What usually people don't understand is that we are using a small embedded device that should run lightweight software on it.
Good examples of the concept are old "PDAs" with PocketPC.

Okay, things have evolved and we have a huge, fat system called Android here, but why should we use a scheduler that guarantees disk throughput by doing slow calculations about blocks vs size on a system that doesn't make heavy usage of MySQL, Apache and other server-based programs?

zachariasmaladroit:
ROW by all means,

I don't understand it either, there are several reports that state that latency and interactivity are worse with BFQ,

it was even reported on the discussion forum of BFQ.

BFQ is **awesome** on the Linux Desktop and multimedia, really cuts down the crap,

but on Android it's the wrong tool for the job and too "bloated",

Why not try out cfq instead and tweak that ?

If that doesn't work, stick to ROW, noop or deadline

PR #643, opened 2016-03-14
update f2fs to current version

mustermaxmueller:
https://kernel.googlesource.com/pub/scm/linux/kernel/git/jaegeuk/f2fs-stable/+log/linux-3.10.y


jerpelea:
without a f2fs userspace this commit is useless

PR #644, opened 2016-03-14
rhine: set initial gpu frequency to 27mhz

mustermaxmueller:


jerpelea:
Please add commit message to each commit

freexperia:
clock is too low and will cause problems in most usecases

PR #645, opened 2016-03-17
[1.2.2] Refactored Sony Unified Display Driver (and other fixes)

kholk:
This is a major refactoring of our unified display driver.

The new driver supports multi-resolution (and multiple timings)
panels, with dynamic resolution switch, which are features
required by SoMC Satsuki's 4K display.

Now we also have a polling mechanism, taken from Sony's copylefts
and modified for our own driver, which will ensure a smooth
transition between continuous splash bootloader logo and OS
framebuffer takeover (usually displaying an OS logo).

Apart new features, now we have nice improvements in code
readability and nice cleanups:
the driver has now been organized in various specialized files,
which are managing only a specific "function". This can be further
improved (TODO in the future) by making the driver modular,
making us able to compile it with the complete omission of some
specific feature that we can find only in a specific platform
and/or a specific device (example: cmd-detect only on Rhine, Yukon).

Nice cleanups come on both code movements and refactoring:
I have been able to remove some hacks from "standard" QC code and
either reimplement them properly in our driver or destroying them
in favor of the resolution of some bugs.

About other cleanups, I've finally got rid of some arrays in our
panel specific struct(s) and simplified the DT parsing / panel
parameters re-set after detection. Moreover, the Sony Unified
Display Driver specific structures, types and definitions have
been removed from common QC headers and moved into our own ones
for enhanced portability across kernel releases.

This whole refactoring will also decrease this driver's footprint
and slightly speed up the boot process.

---

Other fixes include: new Fingerprint Sensor driver (FPC1145), DT updates for MHL and thermal on Kitakami and fixes to HDMI.

PR #647, opened 2016-03-19
flamingo: fix black screen after sleep

Dudulle:
~~This functionality has been reverted in commit d609697.
Flamingo, and maybe some others, still require this.~~

Without this, the panel stays black after going to sleep ~~and the battery does not charge properly.~~

### DIFF
--- a/arch/arm/boot/dts/qcom/dsi-panel-flamingo.dtsi
+++ b/arch/arm/boot/dts/qcom/dsi-panel-flamingo.dtsi
@@ -64,6 +64,8 @@
 		qcom,mdss-dsi-panel-framerate = <60>;
 		qcom,platform-reset-gpio = <&msmgpio 25 0>;
 		qcom,mdss-dsi-reset-sequence = <1 10>, <0 10>, <1 120>;
+		somc,pw-off-rst-seq = <0 10>;
+		somc,pw-on-rst-seq = <1 10>, <0 10>, <1 25>;
 		somc,lcm-bl-gpio = <&msmgpio 15 0>;
 		qcom,panel-phy-regulatorSettings = [03 0a 04 00 20 00 01]; /* Regulator settings */
 		qcom,mdss-dsi-panel-timings = [75 18 10 00 3A 3E 14 1A 13 03 04 00];
@@ -112,7 +114,6 @@
 	        somc,first-boot-aware;
 	        somc,panel-id-read-cmds;
 		somc,dric-only-detect;
-	        somc,panel-pwron-reset;
 	};
 
 	dsi_ofilm_hx8379c_fwvga_vid: qcom,mdss_dsi_ofilm_hx8379c_fwvga_video {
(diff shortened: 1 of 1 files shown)


mikey0000:
Just the commit I was looking for :D

Kali-:
if is just for flamingo make it working only for it as there is no need to waste time to set gpio every time

and this driver is going to be replaced with a newer and cleaner one look at 

https://github.com/sonyxperiadev/kernel/pull/645

Dudulle:
Hi,

The third commit takes care of not doing this on every device by checking the value of pwron_reset.

I have also tested the new driver in #645 and it has the same issue.
This fix works with the new driver too and git will merge the two pull requests correctly.

humberos:
Sorry @Dudulle 
I don't think so!
since mdss_dsi_panel_driver.c will be moved to somc_panel/panel_driver.c on PR #645 
so this change should be applied to new panel_driver.c if it is the case.

Dudulle:
As you said, mdss_dsi_panel_driver.c is moved.
The move is in git's history so it will be handled without any issue.

I have tested already by merging this first and merging #645 first. In both cases the driver was correctly edited.

humberos:
Oh! ok!
thank you!

Dudulle:
I think I found a better way to fix it without touching the gpios.
I will do more tests and I'll update this PR.

mikey0000:
I am having regular lock ups on the unified driver + this PR on flamingo. Trying to track it down.

Dudulle:
You mean the phone gets stuck then reboots?
I have the same issue but it happens already on the main branch.
By the way, I will update this PR in a few minutes for a better solution.

mikey0000:
I have to force it to reboot with the power button. Wasn’t getting it so much on an older kernel I built but yeah.

Dudulle:
Just updated this PR.
This did not actually need to edit the driver.

Dudulle:
@mikey0000 yes I had that problem a few times too.
Not sure if I had this before the previous version of this fix but if that was the cause then it should not happen anymore.

mikey0000:
@Dudulle Thanks for your work.

rcstar6696:
i have a similar issue on sirius if the device goes to sleep for about 2 hours i can't turn the display on again it stays black and only reboot helps to get it back to work its often happen when the device is on charger

Dudulle:
@rcstar6696 the Sirius should not be affected by what was causing this.
On the flamingo, the screen was black but the backlight was on. That was happening almost every time the screen was turned off.
I've experienced the same issue you're describing but I do not know yet it is fixed. Most likely not if the sirus does that too.

The tianchi however could be affected by this. I do not have one so I do not know for sure.

kholk:
I'm glad you found a hack for this.... but it's a hack because pulling HIGH the reset line on the IC prior shutdown doesn't look right. It will constantly drain current when display is off (because pullup drain).

Normally, you want the reset line to be LOW when the display is OFF.
Have you tried messing with the poweron reset sequence? :)

Dudulle:
Haha yes I did not notice.
I've tried multiple values and pushed the one which seemed to work best.
Yes I've tried using both pw-on and pw-off sequences. At least one combination was working. I did not test that much values so it should not be hard to find it back.

kholk:
If you can find a proper thing, then it's fantastic.
If you cannot, I guess this solution will be temporarily accepted until me or you or someone else, whatever, comes up with something better :)

Notify when you're done!

rancidfrog:
Aries, msm8974, also has issues, rarely but it happens once in a while, where unresponsive black screen after sleep. Phone did not register missed calls, and after a long while it responded.

Dudulle:
@kholk this should do it :)
Thanks for your help.

kholk:
Looks sane, can be merged.

PR #651, opened 2016-03-23
wifi update from aosp + bcm pcie improvment

Kali-:
require new kernel config


rcstar6696:
Which new kernel configs are needed?

jerpelea:
tested on all shinano and kitakami 
defconfig needs to be updated

PR #656, opened 2016-03-23
[1.2.2] drivers: misc: pn54x: Fix frame size for 32bits

humberos:
This patch fixes frame size for 32bits targets and avoid
gcc warnings.

kernel/sony/msm/drivers/misc/pn547.c: In function 'pn547_dev_write':
kernel/sony/msm/drivers/misc/pn547.c:308:1: warning: the frame size of 1032 bytes is larger than 1024 bytes [-Wframe-larger-than=]
 }
 ^
kernel/sony/msm/drivers/misc/pn547.c: In function 'pn547_dev_read':
kernel/sony/msm/drivers/misc/pn547.c:254:1: warning: the frame size of 1048 bytes is larger than 1024 bytes [-Wframe-larger-than=]
 }

Change-Id: I5700a6ba970a5882bbbdef205020c256a73495fc

### DIFF
--- a/drivers/misc/pn547.c
+++ b/drivers/misc/pn547.c
@@ -48,9 +48,13 @@
 #include <linux/clk.h>
 #endif
 
-#define MAX_BUFFER_SIZE		512
 #define MAX_NORMAL_FRAME_SIZE	(255 + 3)
-#define MAX_FIRMDL_FRAME_SIZE	(1023 + 5)
+
+#if defined(CONFIG_ARM64) || defined(CONFIG_64BIT)
+#define MAX_FRAME_SIZE	(1023 + 5)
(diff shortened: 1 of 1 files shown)


kholk:
This is wrong, actually. A better option would be to dynamically allocate the variable (and remember to free it).

I already have the patch that I've described, didn't push it and won't push it until Yukon will be fixed.

humberos:
Oh ok!
Sorry.
I've just wanted to get back the original driver and avoid any possible driver error caused by pinctrl support.
Thank you!

jerpelea:
I agree with Kholk, dynamic is the best but this at least corects initial mistake. We should check yukon's problem !

PR #665, opened 2016-03-28
[1.2.2] drivers: mdss: logic revision - part I

humberos:
MDSS driver logic revision - part I
1. drivers: mdss: Fix logic of statement
   - This logic is wrong when comparing it with original driver.
2. drivers: mdss: Bring back bklt_max value
   - bklt_max was missed on unified panel driver.
   - It value is expected for mdss_dsi_panel_bklt_pwm()
3.  drivers: mdss: Check if "gpio" is valid
   - Check if gpio is valid instead of variable verification.

### DIFF
--- a/drivers/video/msm/mdss/somc_panel/panel_driver.c
+++ b/drivers/video/msm/mdss/somc_panel/panel_driver.c
@@ -352,7 +352,7 @@ static int mdss_dsi_panel_reset_seq(struct mdss_panel_data *pdata, int enable)
 		pr_debug("%s:%d, disp_en line not configured\n",
 			   __func__, __LINE__);
 
-	if (!ctrl_pdata->rst_gpio)
+	if (!gpio_is_valid(ctrl_pdata->rst_gpio))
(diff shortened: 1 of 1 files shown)


drivers/video/msm/mdss/somc_panel/panel_driver.c
```diff
@@ -206,6 +206,8 @@ void __mdss_dsi_panel_cmds_send(struct mdss_dsi_ctrl_pdata *ctrl,
 	/*Panel ON/Off commands should be sent in DSI Low Power Mode*/
 	if (pcmds->link_state == DSI_LP_MODE)
 		cmdreq.flags  |= CMD_REQ_LP_MODE;
+	else if (pcmds->link_state == DSI_HS_MODE)
```
kholk: Default mode should be HS, that's why there is no check for that

humberos:
Fixed!
Thank you!

kholk:
Should be ok to merge.

PR #666, opened 2016-03-29
[1.2.2] drivers: mdss: Bring back default tearcheck value

humberos:
From stock kernel we have te->sync_cfg_height = 0xfff0;
This value is used for tearcheck as default.

After mdss unified driver change (49dc516)
the te->sync_cfg_height value was reduced by using other logic as following:

```
vporch = pinfo->lcdc.v_back_porch +
     pinfo->lcdc.v_front_porch +
     pinfo->lcdc.v_pulse_width;

height = (pinfo->yres + vporch) * 2;
```

This patch bring back default value (0xfff0)
since it works better on userspace.

This change will affect only bootanimation loading (low panel
sync) then it needs more investigation.

This change was tested on Kitakami Karin (10.1 Tablet).

Change-Id: Ib1a9621643a218e3caf98238e98a727c10924296


humberos:
it needs to be tested on Z5 family.

humberos:
Seems it breaks panel partial update on recovery mode.
This patch needs more investigation of why panel is broken during its initialization.
On userspace it is fine at least on karin.

PR #668, opened 2016-03-30
[1.2.2] Leds updates for kitakami R2

humberos:
1. Add somc,init-br support
2. Add somc,bl-scale support

To do: check init-br value and convert it to 8 bits if necessary.

### DIFF
--- a/drivers/leds/leds-qpnp-wled.c
+++ b/drivers/leds/leds-qpnp-wled.c
@@ -155,7 +155,8 @@
 #define QPNP_WLED_SC_DLY_MS		20
 
 #define QPNP_WLED_CURR_SCALE_MAX	100
-
+#define QPNP_WLED_BL_SCALE_MAX	1000
+#define QPNP_WLED_BUFF_SIZE	50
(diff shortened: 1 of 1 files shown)


kholk:
IMHO, useless diverging from caf..... All those lines to solve just a 0.5 seconds backlight flash on Android logo after continuous splash? Hmm :)

humberos:
oh! so...  is 0.5 seconds the reason of Sony made it ?

btw, I don't have a K-R2 device but I know that ivy has black screen issue during boot process.
I was thinking this patch could fix the issue.

Also, it makes sense if we check those changes:
- https://github.com/sonyxperiadev/kernel/commit/cfdc9bd5c665eac2a0af6cb81dc37013e30994fd
- https://github.com/sonyxperiadev/kernel/commit/7e4cffb15b7d2a6dd62f4bede621fa41ad4c5b12
- https://github.com/sonyxperiadev/kernel/commit/41545aae984de16dd3e82547b6a8f9dd1bbed173

PR #671, opened 2016-04-01
[1.2.2] qcom: rpm-smd: Add NULL check for input data variable

humberos:
Add NULL check to prevent NULL pointer dereference during memcmp.


Tommy-Geenexus:
Why not simply leave the conditional check a oneliner and verify which pointer is NULL in the error message?

humberos:
yeah!
like:

```
pr_err("%s(): Invalid input (%s %s)\n", __func__,
    (!handle) ? "handle" : "",
    (!data) ? "data" : "");
```

humberos:
I think this change is not required... we will get error message anyway if one of both is not present....
I will remove it.

humberos:
Thank you @Tommy-Geenexus

PR #676, opened 2016-04-12
update primadriver to LA.BR.1.3.3_rb2.14

erikcas:
also a reimplementation of the driver. Instead of solving compile errors in the driver by ifdeffing, enable configs in defconfig, exept:
- Linux regulatory feature is not implemented in kernel, so ifdeffed
- Iris name is not exposed, reverted it in primadriver
- pick some wcnss caf commits prima driver has implemented

prima driver is taken from https://us.codeaurora.org/cgit/external/gigabyte/platform/vendor/qcom-opensource/wlan/prima/log/?h=caf/LA.BR.1.3.3_rb2.14

exact changes done to caf source, see https://github.com/erikcas/prima/commits/LA.BR.1.3.3_rb2.14


erikcas:
tested on amami and eagle

@jerpelea left debugging on currently, for testing purposes. If tested oke we can disable it to save space on kernelblob

erikcas:
debugging is disabled now

PR #677, opened 2016-04-12
dts: set qcom,mdss-dsi-bl-max-level to 4095 for all devices

oshmoun:
This fixes the low brightness issue on these devices.
Tested on suzuran but should be the same for sumire. Testing is welcome of course.

lights HAL still sees 0-255 range because it accesses lcd-backlight not wled. msm_fb scales the value of wled/brightness according to lcd-backlight/brightness. Code for that is here:
https://github.com/sonyxperiadev/kernel/blob/aosp/LA.BF64.1.2.2_rb4.7/drivers/video/msm/msm_fb.c#L184
With this commit setting lcd-backlight/brightness to 255 sets wled/brightness to 4095, which is max brightness according to this:
https://github.com/sonyxperiadev/kernel/blob/aosp/LA.BF64.1.2.2_rb4.7/drivers/leds/leds-qpnp-wled.c#L1406
notice that changing wled max_brightness to 255 actually lowered max brightness, so there is no scaling happening for wled to ensure maximum possible brightness is achieved whatever the chosen range.

### DIFF
--- a/arch/arm/boot/dts/qcom/dsi-panel-amami.dtsi
+++ b/arch/arm/boot/dts/qcom/dsi-panel-amami.dtsi
@@ -460,7 +460,7 @@
 		qcom,mdss-dsi-t-clk-post = <0x0c>;
 		qcom,mdss-dsi-t-clk-pre = <0x1a>;
 		qcom,mdss-dsi-bl-min-level = <1>;
-		qcom,mdss-dsi-bl-max-level = <255>;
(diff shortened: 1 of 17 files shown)


zachariasmaladroit:
will give this a test,

thanks !

jerpelea:
cat /sys/class/leds/wled:backlight/max_brightness && cat /sys/class/leds/lcd-backlight/max_brightness
255
255

PR #683, opened 2016-04-15
video: msm: somc_panel: Fix detection fallback for CMD china-panels

kholk:
Chinese spare parts for various Sony (and other OEMs) devices
will not have an original Sony panel, but something that is
usually compatible with the default panel entry in the DTs.

This is funny, as actually no Sony original panels will use
the default entry! Anyway, CMD panel detection has fallback
problems because of a code logic bug, causing MDSS backlight
to not be registered, but panel working.

Fix the fallback entry by adding an unlikely check on the actual
detection status, instead of stupidly and blindly saying that
we have detected the panel even when we haven't.


alviteri:
+1 fixed my china panel

PR #684, opened 2016-04-20
[1.2.2] drivers: mmc: Fix mmc block deferred resume

humberos:
Fix the logic for mmc block requesting process.
The logic was extracted from 32.1.F.0.x.

The patch fixes this BUG:

<4>[   52.765552] IRQ39 no longer affine to CPU1
<2>[   54.071351] EXT4-fs error (device mmcblk0p25): ext4_lookup:1437: inode #147305: comm SharedPreferenc: deleted inode referenced: 147159
<3>[   54.082730] Aborting journal on device mmcblk0p25-8.
<0>[   54.092589] Kernel panic - not syncing: EXT4-fs (device mmcblk0p25): panic forced after error

### DIFF
--- a/drivers/mmc/card/block.c
+++ b/drivers/mmc/card/block.c
@@ -2805,13 +2805,14 @@ static int mmc_blk_issue_rq(struct mmc_queue *mq, struct request *req)
 		mmc_rpm_hold(host, &card->dev);
 		/* claim host only for the first request */
 		mmc_claim_host(card->host);
+		if (mmc_card_get_bkops_en_manual(card))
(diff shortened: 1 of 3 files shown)


kholk:
Wonderful fix, congrats!
Prior merging, though, @erikcas has this problem on Sumire as well, so I'd like to hear from him!

(or, well, any other people who has KitakamiR2, ofc)

erikcas:
@kholk tested 2 out of 3 patches since yesterday. Problem is gone.

Low voltage patch not tested but since it is inline with stock kernel for sumire think that is good to go

PR #689, opened 2016-04-25
[1.2.2] arm: config: shinanoR2: Add CONFIG_MMC_DEV_DRV_STR_TYPE4

humberos:
Fix MMC bus width.

It is also needed on Aries and Leo.
Scorpion is ok!

Change-Id: I5c7c4463dd0a78a56b2fd5c1050ba54603e94bd4


humberos:
Improvements:

http://postimg.org/gallery/2345cey6w/57df5de9/

Using patch (001...img): 
Read: 216MB/s
Write: 45MB/s

Without patch (002...img): 
Read: 215MB/s
Write: 44MB/s

Also, this flag is present at stock version.

kholk:
Why wasn't it enabled?! This config option was actually made just for Shinano R2 devices! Merge!

PR #690, opened 2016-04-25
[1.2.2] drivers: mmc: sdhci: Implements mutex for next_data

humberos:
Implements mutex control for next_data and avoid unbalanced jobs.

This implementation was extracted from stock kernel.

Change-Id: I3eeff8990329eb7cdd7d25e3d03178bbd5ab3bc8

### DIFF
--- a/drivers/mmc/host/sdhci.c
+++ b/drivers/mmc/host/sdhci.c
@@ -598,6 +598,9 @@ static int sdhci_pre_dma_transfer(struct sdhci_host *host,
 				  struct sdhci_next *next)
 {
 	int sg_count;
+	unsigned long flags;
+
+	spin_lock_irqsave(&host->next_lock, flags);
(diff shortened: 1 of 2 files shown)


kholk:
This impacts performance. What are we trying to fix here?

humberos:
@kholk : This change was made for 3.4 and 3.10 trees by Sony.

I was trying to fix sdhci timeout warnings but now it is happening during reboots cycles.

WARNING: at ../../../../../../kernel/sony/msm/drivers/mmc/host/sdhci.c:1755 sdhci_request+0x40/0x2b4()

Seems this patch "reduces" sdhci warnings but unfortunately it still present (as I mentioned now it happens only during reboots cycles).

Is there something related to external sd ?

Btw, this warning is present on Shinano R2 Leo.

PR #696, opened 2016-04-29
[1.2.2] msm8974: Fix torch

humberos:
Fix torch values by using stock configuration.

This patch set fixes : 
- torch duration
- code style

Tested on shinano Leo.

### DIFF
--- a/arch/arm/boot/dts/qcom/msm8974pro-ac-shinano_leo_camera.dtsi
+++ b/arch/arm/boot/dts/qcom/msm8974pro-ac-shinano_leo_camera.dtsi
@@ -21,7 +21,7 @@
                 cell-index = <0>;
                 compatible = "qcom,camera-flash";
                 qcom,flash-type = <1>;
-		qcom,torch-source = <&pm8941_torch>;
+                qcom,torch-source = <&pm8941_torch>;
                 qcom,flash-source = <&pm8941_flash0>;
         };
 };
--- a/drivers/leds/leds-qpnp.c
+++ b/drivers/leds/leds-qpnp.c
@@ -27,6 +27,10 @@
 #include <linux/regulator/consumer.h>
 #include <linux/delay.h>
 
+#if defined(CONFIG_MACH_SONY_SHINANO) || defined(CONFIG_MACH_SONY_RHINE)
+#define SONY_LEGACY_TORCH
+#endif
+
 #define WLED_MOD_EN_REG(base, n)	(base + 0x60 + n*0x10)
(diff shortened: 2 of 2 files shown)


humberos:
ping @rcstar6696

rcstar6696:
Torch is okay but flashlight while takeing a photo is not okay flashlight turns of to early

humberos:
Yeah, flash requires camerav2 changes. But it brightnnes was increased as I mentioned in PR/commit message. Torch is the main fix here.

rcstar6696:
Okay then we can merge it sirius is okay here castor has no torch :)

rcstar6696:
When i use torch atm it flickers in a high frecuency but it does not turn off like before

humberos:
You are right. It happens due to VFE IRQ problem. Just check logs then you'll see the mess

humberos:
But if you use torch from a fresh reboot it works perfectly.

kholk:
OMG!!!! Are you really sure about those current and max-current value?!??! That looks HUGE!!!
I have written those values based on what Sony's driver for flash was doing.......

About the qpnp-led changes, those look really scary, but well, if they're working... okay.

jerpelea:
this is dangerous and we should avoid burning hw!

jerpelea:
please remove the max-curent changes untill we check the hw

humberos:
ok, working on it.

jerpelea:
The patch has to be tested on all platforms not only shinano
please post the steps to replicate the bug

humberos:
Without this patch, torch (or Android Flashlight option)  works only during 12 seconds and its brightness is not so good when compared with stock version.

Although this patch fixes torch on Shinano devices (at least) but the VFE still have IRQ problems during torch or camera flash enabled (using this patch or not).

VFE logs:
http://paste.omnirom.org/view/34b0b271

humberos:
ping @erikcas
please, could you confirm if this issue is present on rhine nor kitakami ? 
thank you.

erikcas:
@humberos torch off after 12 seconds is confirmed on amami. Brightness however seems fine to me, but I have not seen stock on amami for ages.

Sumire atm on stock, will flash M in half an hour or so and edit here

humberos:
@erikcas 
Thank you!
Great! So, could you apply this patch for rhine too and see if it will get flashlight duration fixed.
If it is ok then I will update this patch by adding rhine.

erikcas:
@Humberos will do. On Sumire NO problems with flashlight duration

erikcas:
@humberos adjusted for rhine and tested on amami: +1

alviteri:
Tested on honami +1

PR #699, opened 2016-05-01
kitakami: scale backlight to 12bit on devices that need it

stellirin:
Android uses 8bit (256 values) but on some devices the backlight
needs 12bit (4096 values). The display driver automatically scales
android's 8bit value to a 12bit value with transition.


stellirin:
devices tested here that need this fix:
- karin (no)
- ivy (yes)
- suzuran (yes)

Devices not tested here:
- sumire (yes)
- satsuki (unknown, no?)

If the two devices can be tested I can add satsuki or remove sumire if necessary.

kholk:
Add satsuki

PR #702, opened 2016-05-05
rqbalance : ensure MAX cpu is set when storing nr_run_thresholds from…

BuzzBumbleBee:
… userspace

Tested on shinanio (Z3C) and kitakami (Z3+)

Stops rqbalance max cpu from being unset when user sets from sysfs


kholk:
Looks ok to me!

PR #710, opened 2016-05-05
[1.2.2] shinano: drivers: mmc: Fix card resume/initialization voltage

humberos:
The card resume/initialization should be done at 3.3 V I/O voltage
for shinano devices.

These values was extracted from kernel-copyleft-23.4.A.0.x.

Change-Id: Ib22585dba975c3d771b9cbc391bdeae1d9874747


humberos:
damn... it still present....
needs more investigation

zachariasmaladroit:
With what microSD are you experiencing this ?

humberos:
@zachariasmaladroit not sure if the problem is really the microSD. Btw I am using 32GB Samsung SDHC (I think it is class 10).

PR #714, opened 2016-05-10
dts: add new msm-id for castor to match hw again

rcstar6696:


jerpelea:
10000 are you shure?

rcstar6696:
Yes it  boots with 10000

PR #718, opened 2016-05-15
ARM: dt: msm8226: Add AV timer specific data

alviteri:
LPASS exposes registers from which timestamp can be
read by clients for audio & video synchronization.
Change adds the register address of AV timer in LPASS.

CRs-fixed: 677082
Change-Id: I4677a745a48913d53d91e7bdeb62ad3ee5358cb3
Signed-off-by: Laxminath Kasam lkasam@codeaurora.org
Signed-off-by: Shreyas Nagasandra Chandrasekhar snagas@codeaurora.org

### DIFF
--- a/arch/arm/boot/dts/qcom/msm8226.dtsi
+++ b/arch/arm/boot/dts/qcom/msm8226.dtsi
@@ -703,6 +703,13 @@
 		qcom,msm-auxpcm-interface = "primary";
 	};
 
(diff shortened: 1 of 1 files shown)


alviteri:
@jerpelea triple check please before merge.... not 8226 available here

kholk:
8226 shouldn't have avtimer :\

PR #722, opened 2016-05-27
[1.3.3] Add support for Loire Suzu (Xperia X/AX RoW/Dual)

kholk:
Experimental.

Also includes new charger driver.


oshmoun:
1.3.3 build for suzuran breaks with these commits
here's a diff to make it build
http://pastebin.com/NLN0gz43
the diff is by no means a fix, should just help to point out where the build breaks ;)

oshmoun:
situation of usb/charging on suzuran is unchanged with these commits + diff above to allow build. tested by booting into twrp recovery.

kholk:
This is a real pity but.. expected. In any case, if you check the commits, those are implementing the new methods for the smbcharger on the PHY driver: that's because Xperia X uses a ChipIdea controller.

Top-tier chips, like 8974, 8994, 8996 are using the DWC3 controller, which obviously uses another driver.
The new SMBCharger implementation is still TBD for dwc3 (and don't forget: we also have a variation in the impl because of the MHL bridge)!

oshmoun:
usb/charging on suzuran is working now
:+1:

PR #723, opened 2016-05-27
[1.2.2] asoc: shinano: Remove Add UltraLowLatency

humberos:
It is breaking audio for notifications on shinano devices.
So move it to rhine platform which was only the tested target.

Change-Id: I71416d943e9ca016843f9c94f6f13ba35f03af3d


humberos:
Ping @rcstar6696 
Ping @alviteri

alviteri:
@humberos revert it .. all are 8974 .... like offload there is anything in kernel which make us break the things ....

PR #728, opened 2016-06-02
yukon: battery/charger: fixes and cleanup

Dudulle:


Dudulle:
The last commit requires https://github.com/sonyxperiadev/device-sony-flamingo/pull/43 to work as intended on flamingo.

arch/arm/boot/dts/qcom/batterydata-yukon-flamingo.dtsi
```diff
@@ -72,7 +72,7 @@ qcom,flamingo-batterydata {
 	};
 
 	qcom,rbatt-sf-lut {
-		qcom,lut-col-legend = <(-20) 0 25 40 60 70>;
```
kholk, line 75: This looks REALLY wrong. You are lowering the algorithm precision by removing that extra value.

Dudulle, line 75: I'm pretty sure this value was added by mistake. I checked in kernel 18.5.C.0.25 and it is not there.
Besides that, if the value is correct, then there would be some data missing in qcom,lut-data.

kholk, line 75: My bad, then :)

PR #731, opened 2016-06-07
DO NOT MERGE: try and calibrate emmc_therm and msm_therm

oshmoun:
current values seem to be doubled, or have a large offset, since they
reach into the 80 degrees range quite easily and are in the 60 degrees
range when idling.

@kholk pretty sure this is awfully wrong, but the documentation isn't exactly helping...
https://github.com/sonyxperiadev/kernel/blob/aosp/LA.BF64.1.2.2_rb4.7/Documentation/devicetree/bindings/hwmon/qpnp-adc-current.txt#L72


kholk:
Well, this time the documentation is wrong. This change is absolutely valid on Kitakami R2 (I wonder how I forgot those values -- I remember I changed them, must be some commit refactoring gone wrong), but needs to be tested on the R1 devices (ivy, karin).

humberos:
I can test it on my karin.
Sorry, although I do not understand very well what this patch is doing but just let me know what do you expect from me :P

oshmoun:
The thermal readings are inaccurate on some kitakami devices (observed on suzuran and sumire)
You could either read the values directly from sysfs or use an app (aida64)  to check if the values for emmc_therm and msm_therm are normal or not with this commit

humberos:
- Without this patch:
  msm_therm ≃ 47.3
  emmc_therm ≃ 49.3
- Using this patch:
  msm_therm ≃ 24.2
  emmc_therm ≃ 26.9

humberos:
Ok!
According docs we should use qcom,scale-function = <2> which seems fix the problem.
So, just remove:

>   chan@73 { /\* msm_therm _/
>       qcom,scale-function = <13>;
>   };
>   chan@74 { /_ emmc_therm */
>       qcom,scale-function = <13>;
>   };

from **msm8994-kitakami_common.dtsi**
- Now I have these values:
  - msm_therm ≃ 30.0
  - emmc_therm ≃ 31.0

which IMHO are correct when compared with others.

screenshot_20160611-202120

oshmoun:
so follow documentation or stock configuration? I'm confused :(

humberos:
Dunno!
What side is it broken ? qcom or stock ?

oshmoun:
PR for 1.3.3 here
https://github.com/sonyxperiadev/kernel/pull/742
went with Humberos' suggestion, please test on other kitakami device.

PR #734, opened 2016-06-16
arm64: config: disable msm_performance

oshmoun:
with the latest PowerHAL changes, msm_performance shouldn't be needed
and can be put to rest.


kholk:
OK to MERGE :+1: 

rqbalance replaces msm_performance.

PR #736, opened 2016-06-16
Revert "arm64: config: disable msm_performance"

stellirin:
Actually we DO need this, msm_performance is the ONLY reliable way
we have for HMP cluster throttling (max cpufreq limiting).

on HMP all cores of a cluster can go offline so we lose userspace
permissions on the key sysfs. msm_performance never goes 'offline'
so we can set maximum frequencies via this module.

This is therefore needed by thermanager.

https://github.com/sonyxperiadev/device-sony-ivy/blob/master/rootdir/system/etc/thermanager.xml#L60


kholk:
Offline one HMP cluster through rqbalance. No need for conflicting drivers. 

Cpufreq throttling was a working feature on my rqbalance, which was originally developed for Tegra 3: it actually already supports managing frequencies, but it has been crippled because of the lack of support for this in userspace (thermanager and powerhal). Now that we've got support for it in powerhal, and partial support in thermanager, I don't see the need to still stay legacy.
Thermanager support has to be completed to support frequency throttling through rqbalance.

stellirin:
This is not about taking cores offline, this is about setting the maximum scaling frequency.

On older devices we set system perms on `/sys/devices/system/cpu/cpu0/cpufreq/scaling_max_freq` and this is enough. When cpu1-3 is offlined the sysfs is destroyed, so it loses it perms and values. When those cores are onlined it reuses the same perms and values from cpu0

https://github.com/sonyxperiadev/device-sony-honami/blob/master/rootdir/system/etc/thermanager.xml#L18

On HMP this is not possible. Each core takes the perms and values of the first cpu in each **cluster**. This means when cpu4-7 are offline all cpus in that cluster lose the perms and values. The default is root perms which is impossible for us to work with.

This is why msm_performance is critical for thermanager. QCOM uses the `perfd` daemon running as root, and the HALs pass everything through this, to work around the root perms problem.

Another issue is that thermanager only sets scaling _once_. THis means if cpu3-7 all go offline then if one comes back online it will take its default values. However if thermanager had previously set a scaling limit, this will be lost. This situation is actually very common!

Using msm_performance for scaling also solves this issue.

jerpelea:
we need a creative way to work around this issue

PR #739, opened 2016-06-18
[1.2.2] arm: irq: Fix logspam by removing harmless irq messages

humberos:
These kernel messages are harmless so do not populate dmesg with them

Change-Id: Ia2a4af1ac9d55db50df63266ea4b12d93f14f77c


kholk:
We should solve the problem instead of hiding the message.

humberos:
oh, is it a problem ?
I was thinking that it is just a harmless messages.

PR #740, opened 2016-06-26
Fix kernel compilation for honami.

Diewi:
Without these changes, there are missing symbols. Thus, the compilation for Z1 fails. This commit fixes these issues.


net/ipv4/Kconfig
```diff
@@ -485,6 +485,7 @@ config TCP_CONG_WESTWOOD
 config TCP_CONG_HTCP
         tristate "H-TCP"
         default m
+        depends on (IPV6 || IPV6=n)
```
kholk, line 488: Basically, that config option depends on "whatever"?
This line is a huge nop!

Diewi:
EDIT: True, I'll rework the commit tonight. It collapses to some defconfig changes.

# 

As far as I remember, this/these line(s) were needed to make the linking step to operate correctly (although these are truly nops),
I will compile another build tonight without these specific line(s) to ensure that I did not confuse something here and report the results in this place.

Diewi:
Could not push without a merge for whatever reason to update the commit... I'll open a new one...

PR #744, opened 2016-07-02
Updates for Suzu

stellirin:
building the kernel inline is more robust
removes dtb on rebuild
do not build modules if not enabled


alviteri:
+1

PR #745, opened 2016-07-02
[1.3.3] arm: DT: Kitakami: Fix scale function choice in VADC channels

kholk:
Some VADC channels are connected to the 100K pullup thermistor.
The qcom,scale-function number got off-phase after some kernel
drivers update from copyleft releases.

Fix the scale function configuration in DT for global Kitakami
and write small but effective documentation for the values.

### DIFF
--- a/arch/arm/boot/dts/qcom/msm8994-kitakami_common.dtsi
+++ b/arch/arm/boot/dts/qcom/msm8994-kitakami_common.dtsi
@@ -3275,10 +3275,10 @@
 		qcom,fast-avg-setup = <0>;
 	};
 	chan@73 { /* msm_therm */
-		qcom,scale-function = <13>;
+		qcom,scale-function = <18>; /* 100K_PULLUP_DECI */
 	};
 	chan@74 { /* emmc_therm */
-		qcom,scale-function = <13>;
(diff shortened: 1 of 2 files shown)


kholk:
This PR is CRITICAL.

Fixes thermal management on MSM8994 Kitakami.

stellirin:
Tested, good, when merged we can modify thermanager configs to use emmc_therm like stock instead of  pm8994_tz

zachariasmaladroit:
@kholk do we also need this for 1.2.2 kernel ?

https://github.com/sonyxperiadev/kernel/blob/aosp/LA.BF64.1.2.2_rb4.7/arch/arm/boot/dts/qcom/msm8994-kitakami_r2_common.dtsi#L300

https://github.com/sonyxperiadev/kernel/blob/aosp/LA.BF64.1.2.2_rb4.7/arch/arm/boot/dts/qcom/msm8994-kitakami_common.dtsi#L3224
https://github.com/sonyxperiadev/kernel/blob/aosp/LA.BF64.1.2.2_rb4.7/arch/arm/boot/dts/qcom/msm8994-kitakami_common.dtsi#L3227

PR #746, opened 2016-07-05
[1.2.2] Split DSDS and ROW kitakami devices

Myself5:
when Appending both DTBs (DSDS and ROW) the DSDS Variant is falsely selecting the ROW DTB, resulting into a broken SDCard support. Split them as discussed.


Myself5:
As discussed, 1.2.2 is EoL soon, so this PR is no longer needed.

PR #764, opened 2016-07-18
ASoC: wcd9320: Fix the bit width configuration per codec dai

alviteri:
Following commit: 6398d603abcb7cbb9f6256618b55310be3619ae9


alviteri:
@jerpelea this is a port from tomtom wcd9330 ..... needs test .... thanks

kholk:
This will work. The hardware is capable of doing what you're asking it to do. Good job

PR #768, opened 2016-07-20
arm: DT: ehci-host: fix clock

alviteri:
cxo_ehci_host_clk (xo) is used by clock-rpm-8974 for ehci-host

### DIFF
--- a/arch/arm/boot/dts/qcom/msm8974.dtsi
+++ b/arch/arm/boot/dts/qcom/msm8974.dtsi
@@ -1761,6 +1761,7 @@
 		compatible = "qcom,usb-dbm-1p4";
 		reg = <0xf92f8000 0x1000>;
 	};
+
 	ehci: qcom,ehci-host@f9a55000 {
 		compatible = "qcom,ehci-host";
 		status = "disabled";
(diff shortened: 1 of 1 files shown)


alviteri:
@jerpelea @kholk

kholk:
I don't know if the xo clock is even called in ehci, but anyway..... usb2b sleep clock is evidently on gcc
arch/arm/mach-msm/clock-gcc-8974@2095

Moreover, looking at the same file, I can see that (line 2535) there's no xo clock declared for msm_ehci_host. Verify whether there is a call for that XO clock in the ehci host driver for MSM or not: if there's none, remove the declaration for it in the DT.

alviteri:
@kholk 

https://github.com/sonyxperiadev/kernel/blob/aosp/LA.BR.1.3.3_rb2.14/arch/arm/mach-msm/clock-rpm-8974.c#L173

alviteri:
commit updated

alviteri:
@kholk https://github.com/sonyxperiadev/kernel/blob/aosp/LA.BR.1.3.3_rb2.14/drivers/usb/host/ehci-msm2.c#L1369

stellirin:
This fixes the charging issues I had on Sirius **BUT** it introduces new issues!

When you unplug the charger the devices still thinks it is charging (charger icon stays in status bar) and the device cannot shut down.

reproducible:
- Power on
- insert charger
- remove charger
- power off
- try to enter fastboot

The device will not enter fastboot until I use the SIM power button, which means the device did not power off properly.

alviteri:
I can't reproduce this.....

rcstar6696:
same as adam explained here on castor_windy

alviteri:
Seems not solve nothing

PR #773, opened 2016-07-26
msm: mdss: dsi: Fix warnings

Dudulle:
drivers/video/msm/mdss/mdss_dsi.c:132:12: warning: 'mdss_dsi_labibb_vreg_ctrl' defined but not used [-Wunused-function]


jerpelea:
this change is only cosmetic and can be integrated in a real fix

PR #776, opened 2016-07-30
[1.3.3] Fixes for Loire bluetooth

humberos:
Welcome bluetooth and deep sleep.


humberos:
It requires https://github.com/sonyxperiadev/device-sony-loire/pull/15

PR #780, opened 2016-07-31
[1.2.2] staging: android: lowmemorykiller: Remove rbtree implementation.

Dudulle:
The implementation is not complete and causes random reboots.
The overall performance improvement (if any) is most likely not worth fixing and maintaining it.

This reverts commit b246d16d2220bbeac9f70df3eabe6dbad0507f2e and associated commits.

Signed-off-by: Cyril Kaiser cyril.kaiser2@gmail.com

Tested on flamingo. I haven't experienced any reboot since I removed it a few days ago (vs multiple reboots/day with the rbtree enabled)


jerpelea:
it is enough to disable the option in config on the targeted device(s) instead of removing the code
when you want to revert a commit you should use "git revert"

PR #786, opened 2016-08-02
[1.3.3] Fix race condition for rfkill clock voting

humberos:
Fix unbalanced rfkill clock voting and revert unused msm_serial implementation.


rcstar6696:
Seems not to solve problems with bluetooth on shinano but after a while the clocks are up and bluetooth turns on

[  575.173556] msm_hs_tx_empty(): Failed.Clocks are OFF
[  575.177647] msm_hs_stop_rx_locked(): Clocks are off
[  575.182373] msm_hs_tx_empty(): Failed.Clocks are OFF
[  575.187520] msm_hs_set_mctrl_locked(): Clocks are off

humberos:
@rcstar6696 thank you. 
We will get those clocks output when uart port is not ready yet which is normal if you get bluetooth working normally after boot but it is not normal if bluetooth is turning on/off by itself after boot.

PR #787, opened 2016-08-03
[1.2.2] arm: msm: clocks: adapt to actual qcrypto needed clocks

alviteri:
since we are using just one instance for qcrypto there is not reason for have extra clocks


alviteri:
@kholk

kholk:
Test eventual regressions on this prior even thinking of merging it on 1.3.3. In any case, should be a nice improvement if it works as expected. 
Theoretically, the performance of a device with HW disk encryption should be better with two instances of qcrypto, with one doing only disk encryption work (we would lose less time waiting for blocking jobs that can be parallelized).

PR #788, opened 2016-08-03
[1.3.3] arm: msm: clocks: adapt to actual qcrypto needed clocks

alviteri:
since we are using just one instance for qcrypto there is not reason for have extra clocks


alviteri:
@kholk

PR #789, opened 2016-08-04
mmc: Fix for bkops-status read fail on Castor

rcstar6696:
This fixes mmc1: mmc_start_bkops: Failed to read bkops status: 1

### DIFF
--- a/drivers/mmc/core/core.c
+++ b/drivers/mmc/core/core.c
@@ -1519,6 +1519,10 @@ int mmc_read_bkops_status(struct mmc_card *card)
 	}
 
 	mmc_claim_host(card->host);
+	
+#ifdef CONFIG_MACH_SONY_CASTOR
+	err = mmc_send_ext_csd(card, ext_csd);
+#else	
 	if (mmc_card_cmdq(card))
 		err = mmc_cmdq_halt_on_empty_queue(card->host);
 
@@ -1528,6 +1532,7 @@ int mmc_read_bkops_status(struct mmc_card *card)
 		if (mmc_card_cmdq(card))
 			mmc_cmdq_halt(card->host, false);
 	}
+#endif
 	mmc_release_host(card->host);
 	if (err)
 		goto out;


rcstar6696:
ping @kholk

drivers/mmc/core/core.c
```diff
@@ -1519,15 +1519,20 @@ int mmc_read_bkops_status(struct mmc_card *card)
...
+#ifdef CONFIG_MACH_SONY_CASTOR
+	err = mmc_send_ext_csd(card, ext_csd);
+#else	
 	if (mmc_card_cmdq(card))
 		err = mmc_cmdq_halt_on_empty_queue(card->host);
 
 	if (!err) {
 		err = mmc_send_ext_csd(card, ext_csd);
-
+
```
erikcas: oops

kholk:
Looks fine, but should be tested on all devices, and eventually done with DT compatible property only for devices who need it, since the ones that don't need it should be the majority.

rcstar6696:
it seems that only castor needs it noone saw the issue before on 1.3.3
i tested it on sirius too :)

kholk:
Rhine has the same issue

alviteri:
maybe is my fault... but I have grab 3 dmesg in some situations but I can't see this error in honami :|

kholk:
The code is new, only for new eMMCs and should be enabled only for new devices which support the new eMMC standard, which are Dora and Suzu. In this kernel we support only Suzu, so perhaps you should ifdef for suzu and leave the new code disabled for all other devices. It's safe.

jerpelea:
plase add a barier for all 8974 devices (both rhine and shinano)

rancidfrog:
Aries,  8974 has following issue, related?

```
[16384.419460] mmc1: mmc_set_auto_bkops: error in setting auto bkops to 1 (-74)
[16384.426068] mmc1: mmc_init_card: Failed to enable auto-bkops: err: -74
[16384.520883] android_usb gadget: high-speed config #1: fe8050c8.android_usb
[16384.521012] dwc3 f9200000.dwc3: Avail curr from USB = 500
[16384.521500] android_work: android_work: sent uevent USB_STATE=CONNECTED
```

alviteri:
https://github.com/sonyxperiadev/kernel/pull/808

PR #792, opened 2016-08-08
[1.3.3] ashmem: Validate ashmem memory with fops pointer

humberos:
Validate the ashmem memory entry against f_op pointer
rather then comparing its name with path of the dentry.

This is to avoid any invalid access to ashmem area in cases
where some one deliberately set the dentry name to /ashmem.

Change-Id: I74e50cd244f68cb13009cf2355e528485f4de34b
Signed-off-by: Sunil Khatri sunilkh@codeaurora.org

### DIFF
--- a/drivers/staging/android/ashmem.c
+++ b/drivers/staging/android/ashmem.c
@@ -766,11 +766,28 @@ static long compat_ashmem_ioctl(struct file *file, unsigned int cmd, unsigned lo
 }
 #endif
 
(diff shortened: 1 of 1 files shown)


rcstar6696:
Tested on Shinano no problems here

kholk:
Well, it was like querying the USA president to know if your neighbor is at home...... Lmao

The change makes absolute sense, testing is not even required. Merge it!

PR #811, opened 2016-08-18
msm: kgsl: Defer adding the mem entry to a process

scafroglia93:
If we add the mem entry pointer in the process idr and rb tree
too early, other threads can do operations on the entry by
guessing the ID or GPU address before the object gets returned
by the creating operation.

Allocate an ID for the object but don't assign the pointer until
right before the creating function returns ensuring that another
operation can't access it until it is ready.

Bug: 28026365
Bug: 28377352
CRs-Fixed: 1002974
Change-Id: Ic0dedbadc0dd2125bd2a7bcc152972c0555e07f8
Signed-off-by: Jordan Crouse jcrouse@codeaurora.org
Signed-off-by: Sunil Khatri sunilkh@codeaurora.org
Signed-off-by: Santhosh Punugu spunug@codeaurora.org

### DIFF
--- a/drivers/gpu/msm/kgsl.c
+++ b/drivers/gpu/msm/kgsl.c
@@ -1,4 +1,4 @@
-/* Copyright (c) 2008-2015, The Linux Foundation. All rights reserved.
+/* Copyright (c) 2008-2016, The Linux Foundation. All rights reserved.
  *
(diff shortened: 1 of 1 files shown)


mustermaxmueller:
works.
tested on togari 1.2.2 omnirom
according to quadrooter app exploit is fixed

PR #815, opened 2016-08-18
ARM: dts: msm: Update adsp heap memory size

alviteri:
Adding 2M additional to adsp heap size to support
hevc adaptive playback cases. Otherwise, adaptive CTS
test cases can lead to ADSP crash.

Change-Id: I89d9597852cbdc6efce56814752302ba9af0e98f
Signed-off-by: Balamurugan Alagarsamy balaga@codeaurora.org


alviteri:
ping @kholk

kholk:
If you tell me how did you calculate the uA rating per frequency, then I'd say it's acceptable.... 

... But in this moment it's absolutely not

PR #819, opened 2016-08-20
usb: dwc3: only disable vdda33 and vdda18 regulators if enabled enabled

oshmoun:
disabling the regulator without checking state leads to unbalanced
disables.

### DIFF
--- a/drivers/usb/dwc3/dwc3-msm.c
+++ b/drivers/usb/dwc3/dwc3-msm.c
@@ -2479,10 +2479,13 @@ static int dwc3_msm_remove_pulldown(struct dwc3_msm *mdwc, bool rm_pulldown)
 	return 0;
 
 disable_vdda33:
-	ret = regulator_disable(mdwc->vdda33);
-	if (ret)
-		dev_err(mdwc->dev, "Unable to disable vdda33:%d\n", ret);
-
+	// try to disable only if the regulator is enabled
+	// this helps avoid unbalanced disables
+	if (regulator_is_enabled(mdwc->vdda33)) {
+		ret = regulator_disable(mdwc->vdda33);
+		if (ret)
+			dev_err(mdwc->dev, "Unable to disable vdda33:%d\n", ret);
+	}
 unset_vdd33:
 	ret = regulator_set_voltage(mdwc->vdda33, 0, DWC3_3P3_VOL_MAX);
 	if (ret)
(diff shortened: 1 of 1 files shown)


oshmoun:
this is a (bad?) workaround, please don't merge if there's a better fix possible.

kholk:
Would need hardware verification here, but noone is going to measure the thing for real.

The commit is logically perfect, this is the best solution. 
Merge!

drivers/usb/dwc3/dwc3-msm.c
```diff
@@ -2479,10 +2479,13 @@ static int dwc3_msm_remove_pulldown(struct dwc3_msm *mdwc, bool rm_pulldown)
 	return 0;
 
 disable_vdda33:
-	ret = regulator_disable(mdwc->vdda33);
-	if (ret)
-		dev_err(mdwc->dev, "Unable to disable vdda33:%d\n", ret);
-
+	// try to disable only if the regulator is enabled
+	// this helps avoid unbalanced disables
+	if(regulator_is_enabled(mdwc->vdda33)) {
```
kholk: For code style issue, put a space after the if. 
if (regulator_.........

oshmoun:
Hmmm, now another regulator vdda18 is having the same issue, only appeared after a while of using the device and charging...
Guess it's the same fix for that as well!

kholk:
Okay, makes sense. But a research on WHY those regulators aren't enabled at
that point would be good.

Il 20 ago 2016 9:46 PM, "oshmoun" notifications@github.com ha scritto:

oshmoun:
yeah, that's a good idea I guess
at least in dwc3-msm, only the function "dwc3_msm_remove_pulldown" handles regulators, but skips the enabling part if rm_pulldown is false.
The decision if rm_pulldown is false or true happens here:
https://github.com/sonyxperiadev/kernel/blob/aosp/LA.BR.1.3.3_rb2.14/drivers/usb/dwc3/dwc3-msm.c#L2541
The assumption is that case POWER_SUPPLY_DP_DM_DPR_DMR is never reached, for some reason...

Other places where vdda18 and vdda33 regulators are enabled:
phy-msm-ssusb-qmp.c
phy-msm-qusb.c
phy-msm-ssusb.c
phy-msm-hsusb.c

EDIT: removed wrong statement!

oshmoun:
Ok back to this subject.
https://github.com/sonyxperiadev/kernel/commit/0861965af5bfb590e1ba550bf022ff7cf5b3cd94
this is the commit that introduces the new behaviour.
This new behaviour is exactly the opposite of what was previously there! With this commit, dwc3_msm_remove_pulldown() is called with parameter rm_pulldown = true when mdwc->rm_pulldown = false, and with rm_pulldown = false when mdwc->rm_pulldown = true!
Without the commit, dwc3_msm_remove_pulldown() would be called with parameter rm_pulldown = false when mdwc->rm_pulldown = false,and with rm_pulldown = true when mdwc->rm_pulldown = true.
How can this complete reversal of logic be explained?

PR #823, opened 2016-08-21
msm8974: enable FaceDetection engine

alviteri:
camera framework checks for FD HW

Signed-off-by: Alin Jerpelea alin.jerpelea@sonymobile.com


alviteri:
sorry xD

PR #828, opened 2016-08-24
DO NOT MERGE: overwrite current msm_venc.c with CAF version

oshmoun:
this fixes video recording for both back and front cam. Since this is just an overwrite, please don't merge it. This PR is only so more experienced people can find the correct commits/reverts needed.

NOTICE: changes to FPS in media_profiles.xml are required in order to make back cam work (change FPS from 30 to 20).

### DIFF
--- a/drivers/media/platform/msm/vidc/msm_venc.c
+++ b/drivers/media/platform/msm/vidc/msm_venc.c
@@ -2812,18 +2812,6 @@ static int try_set_ctrl(struct msm_vidc_inst *inst, struct v4l2_ctrl *ctrl)
 		property_id = HAL_CONFIG_REALTIME;
 		enable.enable = ctrl->val;
 		pdata = &enable;
-		switch (ctrl->val) {
-		case V4L2_MPEG_VIDC_VIDEO_PRIORITY_REALTIME_DISABLE:
-			inst->flags &= ~VIDC_REALTIME;
-			break;
-		case V4L2_MPEG_VIDC_VIDEO_PRIORITY_REALTIME_ENABLE:
-			inst->flags |= VIDC_REALTIME;
-			break;
-		default:
-			dprintk(VIDC_WARN,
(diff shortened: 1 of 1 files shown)


oshmoun:
https://github.com/sonyxperiadev/kernel/pull/828/commits/52b3a5a19d39c09106740f186edba506b9081859#diff-068a1697fe82c9d4e5fa2354c846308aL3422

this is where the issue was caused before. However,  with this overwrite we lose color space, and result video has switched colors. Best solution would be to adjust the switch block to make it recognize and accept the given color space.

kholk:
I see. 

Looks like a desync between userspace and kernel. 
The call_hfi_op function is calling function venus_hfi_session_set_property as op, which is simply sending a command to the Venus HFI, telling it to use a predetermined color space information to encode the data.

jerpelea:
or better explained d4aa13b98c18fa4e44b5999fda5a9e47a05908fa

oshmoun:
call_hfi_op is never reached, because the kernel is not detecting a valid color space.

It reaches the default case here and exits: https://github.com/sonyxperiadev/kernel/pull/828/commits/52b3a5a19d39c09106740f186edba506b9081859#diff-068a1697fe82c9d4e5fa2354c846308aL3440

dmesg output is something like this:
<6>[ 4118.353148] msm_vidc:  err: Colorspace -225815180 not supported

That negative number looks like the main issue :D
Is the kernel expecting a different data structure?

kholk:
Most likely there is data corruption when userspace passes the data of the structure of pix_mp to the kernel. 
My guess is that userspace has either got parameters of that struct reordered differently or passes a longer struct, which gets misinterpreted when kernel tries to fit it into its one.

kholk:
Better said, check userspace headers!!

oshmoun:
colorspace wasn't set in userspace it seems. Adding it fixes the error.
The colors are stil inaccurate though.
This commit is not needed then.

PR #833, opened 2016-08-29
Improvements MarshMallow

scafroglia93:
- CRC is disable by default to save 30% performance cost. It's possible back to default setting with any kernel app twaeker like Kernel Adiutor
- Add_Random is designed for old HDD so bye bye

Tested with my Z3

### DIFF
--- a/drivers/mmc/core/core.c
+++ b/drivers/mmc/core/core.c
@@ -73,8 +73,8 @@ static const unsigned freqs[] = { 400000, 300000, 200000, 100000 };
  * performance cost, and for other reasons may not always be desired.
  * So we allow it it to be disabled.
  */
-bool use_spi_crc = 1;
-module_param(use_spi_crc, bool, 0);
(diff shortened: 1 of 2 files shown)


kholk:
These patches will make the device unstable on purpose. 
Disabling crc on eMMC is a very bad idea if you don't want to lose your data someday. 

About the add_random patch, this will possibly slow down disk encryption.

scafroglia93:
- CRC: All my device have the crc disable and there's no any kind of problem :)
- add_random is designed to disable random on flash; However, if you tell me that slows down is a big problem

zachariasmaladroit:
Only because you don't value YOUR data doesn't mean that others have to feel the same ;)

(silent) bit corruption is a very very bad thing,

especially with flash storage devices ;)

jerpelea:
this is unsafe
safety should not be ignored for perfomance

PR #834, opened 2016-08-29
defconfig: karin_windy: update to be inline with karin defconfig

stellirin:
Right now karin_windy does not boot because the device repo uses
the karin defconfig. The boot failure is due to the bootloader
not recognising the DT. THis commit updates the windy variant
with all changes to the regular variant, keeping the variant
specific differences, as seen at:

https://github.com/sonyxperiadev/kernel-copyleft/tree/32.2.A.0.xxx/arch/arm64/configs/diffconfig

Specificaly the difference between the two defconfigs is:

-CONFIG_MACH_SONY_KARIN=y
-CONFIG_AD7146=y

+CONFIG_MACH_SONY_KARIN_WINDY=y
+# CONFIG_PMI8994_FLASH is not set
+# CONFIG_SIM_DETECT is not set


jerpelea:
instead of diffing against stock you should use the aosp karin config and add only the relevant change

stellirin:
umm, that's what I did?

PR #840, opened 2016-09-03
[1.3.3] DO NOT MERGE: driver: misc: Update AD7146 driver from 32.1.A.0.x

humberos:


humberos:
we need to check castor and scorpion.

PR #846, opened 2016-09-05
[DO NOT MERGE] drivers: clearpad: Workaround for suzu touchscreen

humberos:
This patch disables wakeup gesture and force incell panel as SP mode
during initialization.

Wakeup gesture feature uses EWU as panel mode which is broken
during suspend/resume.

Also, it moves clearpad hacks to mdss incell driver.

although the panel and touch are not fixed totally but seems
this patch improves the touchscreem functionality a lot.


humberos:
ping @erikcas

humberos:
Thank you @erikcas 
This issue is already solved by using your "dirty" patch.
So, I think that we also could remove clearpad hacks, right ?

PR #848, opened 2016-09-05
Device tree: minor modifications about charging for kitakami

aenglebert:
Some modifications on kitakami DT about charging:
- correct a typo about 9v minimal current
- (add qcom,float-voltage-mv with a value of 4350mV - Removed)


jerpelea:
please remove b6f8707b692547a0311b3ed583debd8f6309c74d because qcom,float-voltage-mv is already defined by
https://github.com/sonyxperiadev/kernel/blob/aosp/LA.BR.1.3.3_rb2.14/arch/arm/boot/dts/qcom/msm-pmi8994.dtsi#L205

aenglebert:
Oups, I didn't see it.
I just removed this commit.

PR #851, opened 2016-09-05
[1.3.3] incell: suzu: fix touchscreen after suspend / resume

erikcas:
on some panels (at least, mine does) touchscreen on suzu does initially work on boot, but after suspend / resume touch functionality is completely absent. This commit fixes that


humberos:
Thank you!

PR #852, opened 2016-09-05
[1.3.3]clearpad: update incell driver from x-performance

erikcas:
taken from 35.0.D.362 kerneldrop. Did not re-apply the (first) blank code, as it seems to be working fine on suzu here in combination with https://github.com/sonyxperiadev/kernel/pull/851

ping @humberos @kholk


humberos:
~~hmmm~~
~~something is breaking wakeup gesture.~~
~~I will check it then report back.~~

Nevermind!
Commit is ok!

PR #857, opened 2016-09-06
[1.3.3][DO NOT MERGE] drivers: clearpad: Workaround for suzu touchscreen

humberos:
This patch forces incell hybrid panels as SP mode
during its initialization.


humberos:
ping @erikcas @kholk 
I think may should I check the FULL/HYBRID panel first ?
Btw, this patch works really nice on my device (Panel 7).

erikcas:
panel 8. Stuck at bootlogo....
Quite strange we get that different results.

I have no problem at all on the updated driver (which was reverted again)

humberos:
Panel 7 is HYBRID and panel 8 is FULL.

erikcas:
still stuck on bootlogo

PR #861, opened 2016-09-08
MSM8974 camera: Updates

alviteri:
remove all FD traces since 8974 has not support


alviteri:
https://github.com/sonyxperiadev/kernel/blob/aosp/LA.BR.1.3.3_rb2.14/arch/arm/boot/dts/qcom/msm8974-v1-iommu.dtsi

https://github.com/sonyxperiadev/kernel/blob/aosp/LA.BR.1.3.3_rb2.14/arch/arm/boot/dts/qcom/msm8974-v2-iommu.dtsi

as you can see only JPEG and VFE IOMMUs available on this platform for CAMERA

alviteri:
ping @kholk

kholk:
Uhm........ Ok!

PR #862, opened 2016-09-08
[1.3.3] drivers: mdss: Call panel post on from specific driver data

humberos:
Just reduce the use of #ifdef in panel_driver.


humberos:
ping @erikcas and @kholk

PR #864, opened 2016-09-10
Fiops I/O Sched

scafroglia93:
Flash based storage has some different characteristics against rotate disk.
1. no I/O seek.
2. read and write I/O cost usually is much different.
3. Time which a request takes depends on request size.
4. High throughput and IOPS, low latency.

Source:https://lkml.org/lkml/2012/1/4/18

### DIFF
--- a/block/Kconfig.iosched
+++ b/block/Kconfig.iosched
@@ -61,6 +61,14 @@ config CFQ_GROUP_IOSCHED
 	---help---
 	  Enable group IO scheduling in CFQ.
 
+config IOSCHED_FIOPS
+	tristate "IOPS based I/O scheduler"
+	default y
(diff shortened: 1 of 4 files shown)


scafroglia93:
Nexus 6P with latest Android 7.0 + Stock Kernel
img_20160911_162308

Add Random is OFF

stellirin:
The copyright header in your patch does not state that this is compatible with this kernel's GPL licence

kholk:
Disabling add_random cripples FS encryption performance. Moreover there is no proof that this very old io Scheduler makes any improvement.

scafroglia93:
Now on Z3 am using ZEN V2 with no problem

PR #865, opened 2016-09-10
ASoC: wcd9330: avoid hardcoding values in tomtom_codec_mclk_enable

oshmoun:
<del>the functions tomtom_codec_mclk_enable() and
tomtom_codec_get_ext_clk_users() were introduced when adding
support to tasha. These calls break sound on kitakami, so ifdef them to
fix the problem.</del>
the values for the fallback to tomtom_codec_ext_clk_en() were hardcoded.
Pass over enable and dapm to avoid that.


zachariasmaladroit:
Awesome !

So this is responsible for those overflow messages and sound stopping after some time ?

I saw those calls and had assumed that they were fine, never underestimate changes :/

oshmoun:
NEVER! :D

PR #875, opened 2016-09-15
[1.3.3] Broadcom BT/FM Combo chip driver

kholk:
This driver allows to enable FM functionality on BT/FM combo chips from Broadcom.
Note that this conflicts with our Qualcomm reference implementation, so needs further examination.


lowtraxx:
Shouldn't N_BRCM_HCI in drivers/bluetooth/broadcom/line_discipline_driver/brcm_hci_uart.h be 25 like defined in libbt?

PR #880, opened 2016-09-17
[1.3.3] drivers: mdss: incell: Fix panel reset sequence

humberos:
Fix reset sequence for panels that are using somc,ewu-rst-seq


humberos:
@erikcas : panel 8 uses somc,ewu-rst-seq so I think this will solve the panel reset problem

humberos:
nevermind... the logic is the same :P

PR #886, opened 2016-09-18
Add FPC1145 platform driver from upstream Sony kernel

EnJens:
Required to control the electrical parts of the fingerprint device. (Communication is handled through TEE QSEECOM driver.)


EnJens:
The DTS part is already in https://github.com/sonyxperiadev/kernel/blob/aosp/LA.BR.1.3.3_rb2.14/arch/arm/boot/dts/qcom/msm8956-loire-common.dtsi#L57 .
Note that this hasn't been tested yet as we don't have the HAL up in AOSP , so merge with caution.

PR #891, opened 2016-09-21
[1.3.3] Upgrade kernel memory management

kholk:
Those upgrades are for not-so-recently discovered problems with memory on Suzuran.
This will improve performance on memory allocation and reclaiming, other than allowing to reclaim more unmovable pages.

### DIFF
--- a/include/trace/events/kmem.h
+++ b/include/trace/events/kmem.h
@@ -268,11 +268,11 @@ TRACE_EVENT(mm_page_alloc_extfrag,
 
 	TP_PROTO(struct page *page,
 		int alloc_order, int fallback_order,
-		int alloc_migratetype, int fallback_migratetype, int new_migratetype),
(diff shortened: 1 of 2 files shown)


kholk:
Please, test on at least two devices per platform. Specifically:
1. One tablet on ARM
2. One tablet on AArch64
3. One compact and one "regular" (3GB RAM) phone on ARM
4. One compact and one "regular" (3GB RAM) phone on AArch64.

humberos:
Should we do a specific procedure/routine during this test ?
I am going to test it on Karin.

kholk:
Not really. Just use the device, possibly do something heavy on page allocation and reclaiming, like opening a lot of apps and then closing them all together.

kholk:
Tests are done.

PR is not making performance better, but makes mm to reclaim more pages that were unreclaimable without those patches, making slightly more memory to be available at no performance cost.

Merge!

PR #895, opened 2016-09-25
Upgrade ZRAM driver and enable it

stellirin:
Android _loves_ using RAM, which is a good thing! If everything is loaded in memory then it runs quicker. Unfortunately Android tries to keep _everything_ in memory, and there are many things that Android says _must_ be kept in memory, even if they are rarely used, if at all.

Fortunately Google has drastically improved the ZRAM driver. It now uses LZ4 algorithm, which is very fast at decompression. This allows those less frequently used services to be pushed into ZRAM but they can be very quickly decompressed so that the user does not notice a delay.

Typical compression rates range from 3:1 to 4:1 which means - if you use a 512MB ZRAM - you effectively gain an extra GB of RAM for Android to play with.

Unfortunately there's not much hard data to be found, but there are many references to Google's internal testings which showed a clear advantage to using ZRAM, even on 'flagship' devices. This is why Angler and bullhead both have ZRAM enabled, and indeed all Sony Stock ROMs have it enabled.

You can see lots of ZRAM statistics at `/sys/block/zram0/` but most important is to see if Google's claims are valid:

```
cat /sys/block/zram0/orig_data_size && \
cat /sys/block/zram0/compr_data_size && \
cat /sys/block/zram0/mem_used_total
```

My testing consisted of installing 30-40 apps from the Play Store on `suzu`, opening them all _three times_ with some minor interaction in each app:

```
 orig_data_size: 129769472 (123.8 MB)
compr_data_size:  32609762 ( 31.1 MB)
 mem_used_total:  36855808 ( 35.5 MB)
```

**Yes Google's claims are valid.**

Here I have a 3.5:1 compression ration, and I never saw any significant difference in app responsiveness.


rcstar6696:
Why do you not add msm8974?  Everytime i see only kitakami

kholk:
After a long private discussion, ZRAM benefits have emerged. Power cost is negligible and so is computational cost in most situations.

This PR can be merged, but please note that the compression algorithm has evolved during time and there were some little performance improvements in newer versions.

Also note that we should try cleancache to reduce ZRAM compressed pages and eventually improve the overall performance, but that shall be done after PR merge.

PR #897, opened 2016-09-26
rhine/shinano: defconfig: Enable Zram for better performance under load

alviteri:
Following: c72a6143f786db41833f1d40d97fd3b61409f403


rcstar6696:
David tested your pr now same as before no zram0 in sys/block 
So zram is not active

ghost:
Doesn't CONFIG_ZSMALLOC=y need to be set as well?

PR #899, opened 2016-09-27
msm: isp: Remove camif cmd from camif_cfg

alviteri:
Enabling CAMIF is done in update_camif_state() at a later stage
then camif_cfg. Remove duplicate configuration of this register
from camif_cfg

Change-Id: I41739a8056be953fb6f3a3a64a4ff6d78c430ce6
Signed-off-by: Harsh Shah harshs@codeaurora.org

### DIFF
--- a/drivers/media/platform/msm/camera_v2/isp/msm_isp40.c
+++ b/drivers/media/platform/msm/camera_v2/isp/msm_isp40.c
@@ -479,11 +479,14 @@ static void msm_vfe40_process_violation_status(
 		return;
 
 	if (violation_status & (1 << 0))
-		pr_err("%s: camif violation\n", __func__);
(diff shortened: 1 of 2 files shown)


kholk:
This commit is basically null. It doesn't actually change any functionality, just removes the interface (input mux) check and replaces it with camif forced mode.
This implies that TESTGEN and external interfaces will never work anymore. But... Apart this... Nothing changed, eh.

alviteri:
the comment is clear when it says removing duplicate config.... but yeah is null for what we are searching for

kholk:
My point is that it is NOT removing any duplicated configuration!!

kholk:
Ah no I've misread, my bad. It removes a register write that was there as paranoid init.

PR #908, opened 2016-10-03
[1.3.3] Do not use MSM_ADRENO_USES_OLD_FIRMWARE as default

humberos:
1. Do not use MSM_ADRENO_USES_OLD_FIRMWARE as default
2. Enable it for Shinano platform.


jerpelea:
all 8974 needs it
please add rhine

PR #910, opened 2016-10-04
kitakami/loire: replace rqbalance with core_ctl

Myself5:


zachariasmaladroit:
In case this comes up again:

```
soc: qcom: msm_perf: prevent cpumask bits to be NULL

    This changes fixes NULL pointer derefence issue for
    cpumask variabls for clusters.
```

should also be included

PR #911, opened 2016-10-05
dts: kitakami: Rename thermal sensor-information child nodes

oshmoun:
this commit makes kitakami follow the renaming that was introduced in
https://github.com/sonyxperiadev/kernel/commit/a884800d19f536ba066b737b38494287fc6a624b
this fixes the issue where reading /sys/module/msm_thermal/sensor_info
would cause kernel to crash.


oshmoun:
I hope I didn't forget a node somewhere. Please check!

stellirin (approved):
Working great!

PR #912, opened 2016-10-05
power: bcl: add special triggering conditions for kitakami

oshmoun:
kitakami devices sometimes draw more power than what is allow by bcl.
this causes bcl to trigger its mitigations, which means limiting
frequencies and number of maximum online cores as soon as drawn current
increases beyond a specific threshold.
this behaviour has been adjusted on kernel-copyleft by means of
modifying bcl conditions. Applying the same conditions from
kernel-copyleft here will help improve performance considerably, as
cores won't get throttled as soon as high workloads appear.

### DIFF
--- a/drivers/power/Kconfig
+++ b/drivers/power/Kconfig
@@ -585,6 +585,12 @@ config MSM_BCL_PERIPHERAL_CTL
 	  provides routines to configure and monitor the BCL
 	  PMIC peripheral.
 
+config MSM_BCL_SOMC_CTL
+	bool "BCL driver to control with new condition"
+	depends on BATTERY_BCL
+	help
+	  Say Y here to enable new condition.
+
 source "drivers/power/reset/Kconfig"
 
 endif # POWER_SUPPLY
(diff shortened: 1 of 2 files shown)


oshmoun:
it's just a reordering in order to keep the ordering like the #else branch:
on kernel-copyleft it is:
(ibat AND soc) OR vph
on 1.3.3 it is:
vph OR (ibat AND soc)

otherwise they're perfectly the same, I hope :D

zachariasmaladroit:
Copying from commits to here:

> Is there a reason why this second occurrence deviates from sony copyleft source ?
> 
> I had to triple-check that I didn't do wrong paste and and this time again (had assumed that meanwhile >changes had slipped in - but it didn't):
> 
> https://github.com/sonyxperiadev/kernel-copyleft/blob/32.2.A.0.xxx/drivers/power/battery_current_limit.c#L329

They are :)

Just noticed that I forgot to modify the condition in my Kernel - for whatever reason I thought we were using # CONFIG_MSM_BCL_SOMC_CTL is not set

:(

PR #920, opened 2016-10-09
Fix silly typo

pablomh:
My bad, sorry


erikcas:
wait a sec for my upcoming pr. It also fixes android dm verity

pablomh:
Will be better addressed by Eric commits

PR #923, opened 2016-10-12
[1.3.3] clk: qcom: mdss-dsi-pll-28hpm: Fix header inclusion for MSM8974

kholk:
We were including MSM8976 clocks spec header on all SoC.
That was wrong!!! We use HPM PLL not only for 8976/8956, but
also for MSM8974, which obviously has its own clock spec header.


alviteri:
Tested on Honami CM14

kholk:
Merge it :)

PR #924, opened 2016-10-14
[CRITICAL] thermal: msm-tsens: Fix TRDY address for MSM8994 tz sensors

kholk:
This was changed when MSM8976 was implemented.
The address is COMPLETELY wrong for MSM8994 and some MDM chips.

We have only 8994 devices, so we are fixing it the FAST way by
just adding an ifdef for them.


kholk:
During porting of kernel 3.18, the issue was noticed.

On our project we have never experienced (luckily!!!) any issue because we run rqbalance, which is helping to keep CPUs cool, and 3 thermal engines, one of which is evaluating ALL sensors, including the ones coming from VADC.
Without this fix, if we remove all thermal engines and rely solely on kernel emergency shutdown on thermal bite, it will simply not fire, allowing hardware permanent damage.

kholk:
Patch has been tested.

Merge!!!!!

PR #926, opened 2016-10-17
arm64: make sys_call_table const

pablomh:
As with x86, mark the sys_call_table const such that it will be placed
in the .rodata section. This will cause attempts to modify the table
(accidental or deliberate) to fail when strict page permissions are in
place. In the absence of strict page permissions, there should be no
functional change.

Signed-off-by: Mark Rutland mark.rutland@arm.com
Acked-by: Will Deacon will.deacon@arm.com
Signed-off-by: Catalin Marinas catalin.marinas@arm.com


alviteri:
git cherry-pick use on this case

pablomh:
I'll do it that way.

PR #932, opened 2016-10-23
[TESTING NEEDED]camera: actuator: use vzalloc instead of kzalloc for init_settings

oshmoun:
On devices with low RAM, high memory usage can lead to depletion of
CMA pages. Those CMA pages aren't reclaimed because compaction isn't
triggered for small allocations. This is exactly what happens when
trying to allocate memory for init_settings. The CMA pages are
depleted, and memory compaction isn't triggered because
size of init_settings is considered too small for that. This leads then
to a failed allocation that makes camera fail.
Avoid allocation errors for init_settings where by using virtually
contiguos memory. Since physically contiguos memory isn't strictly
required here, and vmalloc is largely left unused, this is safe to use.
### loire
- [ ] suzu
- [x] kugo
### kitakmi
- [ ] satsuki
- [ ] satsuki_dsds
- [ ] sumire
- [ ] sumire_dsds
- [x] suzuran
- [ ] ivy
- [ ] ivy_dsds
- [ ] karin
- [ ] karin_windy
### kanuti
- [ ] tulip
### shinano
- [ ] leo
- [ ] aries
- [ ] sirius
- [ ] scorpion
- [ ] scorpion_windy
- [ ] castor
- [ ] castor_windy
### rhine
- [ ] honami
- [ ] amami
- [ ] togari

### DIFF
--- a/drivers/media/platform/msm/camera_v2/sensor/actuator/msm_actuator.c
+++ b/drivers/media/platform/msm/camera_v2/sensor/actuator/msm_actuator.c
@@ -13,6 +13,7 @@
 #define pr_fmt(fmt) "%s:%d " fmt, __func__, __LINE__
 
 #include <linux/module.h>
+#include <linux/vmalloc.h>
 #include "msm_sd.h"
 #include "msm_actuator.h"
 #include "msm_cci.h"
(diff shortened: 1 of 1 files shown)


kholk:
This will lead to obvious performance decrease, because virtually contiguous memory is not necessarily physically contiguous. Also, please note that IOMMU has some issues dealing with not (physically) contiguous memory! 

Probably the performance impact is very negligible even under stress conditions, so I would say that this should be a very good idea to solve this issue.

oshmoun:
Woohoo finally a somewhat usable fix :D
Let me address the points mentioned though. 
About the performance issue:
This allocation is only done once during actuator init, and the memory is freed directly afterwards, so no future access through vmalloc for camera (at least until the next camera start). The performance hit is therefore definitely negligible. 
<del>About contiguity of vmalloc:
Currently, vmalloc is hardly ever used, like at all! Even under heavy usage i only see maximum vmalloc usage of under 100mb,out of about 250mb allotted to vmalloc, at least according to /proc/meminfo. Considering that the size of init_settings is less than a 100kb, it seems pretty reasonable for the assigned vmalloc memory through vzalloc to be contiguous. 
Also, this brings up the question of maybe shrinking vmalloc in order to free up ram, but that's a research for another day ;) </del>
EDIT: ignore that last part, I only read the first three digits and assumed the reading was in bytes. Turns out it's in kB, which makes it dead wrong. Kernel is reporting wrong vmalloc size, so it can't be said for certain there is enough free space there to guarantee a contiguos allocation.

oshmoun:
Small update: The same issue was observed on kugo, a 3GB device. The same fix helps there as well.

kholk:
I wonder why I see no comments from other people. This is an important fix and noone really cared.

Btw, it's valid, merge it!

PR #938, opened 2016-10-27
[WIP]Tulip fixes

erikcas:
some fixes for tulip.
- It seems I have other hardware touchscreen, using cyttsp4, which works with DMA enabled and updated dt
- prima works

but.. still rebooting constantly....... BMS or USB?? Or????


pablomh:
Can we have this in to have a better baseline for 3.18?

@erikcas: "still rebooting constantly" -> do you mean _after_ booting or _while_ booting? I only know one person who has managed to boot his tulip and I'd like to at least have a tree where the phone at least boots... Thx!

erikcas:
fresh build does not fully boot anymore. It reboots on completion.
Later this week when I have more time, I will recheck my local tree with this one

erikcas:
This is no further wip.

PR #943, opened 2016-10-29
[UM 5.5.r1] KGSL patchset for MSM8974, 8976, 8994, 8936

kholk:
This patchset introduces kgsl_legacy, which may be short-lived.

This legacy kgsl driver is very similar to its (currently latest, LA.BR.1.3.6 branches) 3.10 counterpart and works with old blobs made for 3.10 kernels.
It also introduces changes for MSM8994 to boot and MSM8976 (by introducing TBU clock management for its IOMMU) to work completely and basically perfectly.

This driver _may_ be the only way to get MSM8994, MSM8974 and MSM8939 to work on this 3.18 kernel due to the new one requiring a new ringbuffer implementation, which requires new Adreno GPU firmwares, currently not available for older A4xx and A3xx chips.

On the other side, MSM8976 should get soon migrated to the new KGSL featuring the new ringbuffer implementation, since the GPU in that chip is A5xx series, which is the same series as 8996, which has got the new firmware with the new RB.

The patchset is successfully tested on private branch with MSM8936 Tulip, MSM8976 Suzu and MSM8994 Suzuran devices.


mcgi5sr2:
@kholk let me know if/when you want this tested on msm8974. I can do an d5803 aries build

kholk:
This kernel will receive complete MSM8974 support as last platform. Devices with 8996 have precedence, then in order, 8956/76, 8994, 8939, 8974.

Please concentrate your efforts on the current 1.3.3 3.10 kernel to make sure we don't port bugs on the new 3.18 branches. 

Thanks!

mcgi5sr2:
Thanks to all of you guys who keep this development going, it really is amazing. If your ever in the UK beers on me

kholk:
Hehe, beers is the right word! :D

bartcubbins (approved):

PR #946, opened 2016-10-29
[UM 5.5.r1] Performance enhancements patchset

kholk:
This patchset introduces performance improvements for ARM64, ARM and generic kernel infrastructures.

### DIFF
--- a/arch/arm/include/asm/io.h
+++ b/arch/arm/include/asm/io.h
@@ -23,6 +23,7 @@
 
 #ifdef __KERNEL__
 
+#include <linux/string.h>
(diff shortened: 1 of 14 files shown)


stellirin:
AIO is being obsoleted by Google, we may as well drop the patch and disable it on 3.18 already.

https://android-review.googlesource.com/#/c/292158/

> The aio interface adds substantial attack surface for a feature that's
> not being exposed by Android at all. It's unlikely that anyone is using
> the kernel feature directly either. This feature is rarely used even on
> servers.

PR #948, opened 2016-10-30
[DWC3] Add missing return at the end of function dwc3_chg_detect_work

alviteri:


oshmoun:
tested, no regressions on suzuran/kitakami here 👍

PR #955, opened 2016-10-31
dts:kitakami:sumire: overwrite mdss-dsi-on-command for novatek jdi panel

oshmoun:
it has been noted that turning screen on for sumire devices that have a
novatek jdi panel can produce a number of issues ranging from wrong
colors to inverted display.
Noting that ivy uses the same panel, a test was started to see if some
of the different values between sumire and ivy panels would help
alleviate the issue. The conclusion was that using mdss-dsi-on-command
value from ivy fixed the issues happening on sumire.
Thanks to:
kholk@github for suggesting to try out ivy values
Myself5@github for testing on his sumire.


oshmoun:
moved overlay above generic panels as requested.

PR #961, opened 2016-11-01
[LA.UM.5.5.r1] defconfig: add kanuti defconfig

pablomh:
Initial defconfig for Xperia M4 Aqua (tulip)


jerpelea:
there is no suport for this device... yet

pablomh:
I unfortunately know it :(

pablomh:
Is your plan only to add defconfigs for working platforms? I wanted to add it to ease working on it, but I agree with your approach as well and will continue working with forks.

erikcas:
@pablomh the best route imo is to fix tulip in 1.3.3 and with that help @jerpelea and @kholk do the 3.18 magic.

question: do you have recovery fully booting? For example, I have a functining twrp here

kholk:
I have android booting fine and working nice on Tulip 3.18.
I also have the config, so don't care about posting one, I have tested things to push, but that will happen only when ready. 

We are implementing one platform at a time to ensure a hopefully bugless experience as soon as possible. 

Care to fix the remaining bugs on 1.3.3 branch, as this is what I am using to port SoC and Sony boards on the new kernel. If 1.3.3 has a bug on a phone, probably 3.18 will have the same!

PR #969, opened 2016-11-03
[UM 5.5.r1] Add MSM8916, MSM8936, MSM8939 SoC support

kholk:
This patchset introduces support for MSM8916, MSM8936, MSM8939 SoC.

This SoC is used in SoMC Tulip.


jerpelea:
partially cherry-picked

PR #979, opened 2016-11-06
[1.3.3] arm64: loire: suzu: Fix missed file renaming

humberos:
This file was renamed in 2f9fa94c97b21ddb3711293091602803b9f0aab5
But for some reason it still present at sources.

So this patch forces the renaming process and fix the missed file error
during compilation process.

Change-Id: I6a8c554df849240c8c8f42a298414473ee012c98


jerpelea:
your patch is incomplete and you will miss the second adc range. There was a reason I left it like this untill the proper implementation comes.

PR #980, opened 2016-11-06
ion: adjust system heap pool orders

alviteri:
As part of the change to fix unmovable block migration over time, we
need to reduce the orders requested by ion to under order-5
allocations. In a future CL, we can add fixed-size larger-order pools
to improve performance when allocating large buffers.

bug 26916944

Change-Id: I1ca336c8057b984987a5a5c28a86ab36488e08ea


kholk:
From what I remember we are not using that config option, eh! 
Though, the fix for it isn't that bad to have

PR #985, opened 2016-11-10
Add msm id from D6502 (Indian Z2)

alviteri:


kholk (changes requested):
WTF. Really? Is there any specific reason why you didn't add the msm-id to the preexisting dts?!

alviteri:
I said him the same but @jerpelea did the same on other targets like castor see it 

https://github.com/SonyAosp/kernel/commit/995fd70750fe52d17bb84ff1d6d87d0aa6c1ea55

kholk:
Ohhhh..... Right, that's a "hack" to use Google's mkbootimg, appended dtb etc...... Ok

PR #987, opened 2016-11-12
Android 7.0 cam

alviteri:


kholk:
Good for allowing a better description of possible errors coming from unsupported ioctls (also good it is rate limited).

rcstar6696:
tested on castor_windy and sirius seems ok but no noticable change

PR #989, opened 2016-11-14
Fix Panel include after reverting the shinano rework

rcstar6696:
Dtb 2 was no updated


jerpelea:
i will push a new rework soon

rcstar6696:
ok i will close the pr

PR #990, opened 2016-11-15
[1.3.3] drivers: bluetooth: Fix ttyHS0 wakelocks

humberos:
First of all, this change will affect only LOIRE platform.

For some reason bluetooth will not start without clock vote
during rfkill initialization. This is necessary and it works even
with the kernel warning about "unregistered wakeup source".
But it should be called only once, otherwise it will break
"deep sleep" state, consequently causing kernel wakelock.

The clock vote should be done after enabled/disabled bt_reg_on.
Also, the new msm_serial_hs implementation is no necessary anymore.

This patch fixes "deep sleep" state and it improves BT [re]connecting.

Change-Id: Ia6059d6f0cd14a12503d1bae44566417b43dce1c

### DIFF
--- a/drivers/bluetooth/bluesleep.c
+++ b/drivers/bluetooth/bluesleep.c
@@ -91,6 +91,7 @@ enum {
 	DEBUG_SUSPEND = 1U << 1,
 	DEBUG_BTWAKE = 1U << 2,
 	DEBUG_VERBOSE = 1U << 3,
+	DEBUG_BTWRITE = 1U << 4,
 };
 
 static int debug_mask = DEBUG_USER_STATE;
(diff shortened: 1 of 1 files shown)


humberos:
ping @oshmoun @rcstar6696
Loire is fixed.

erikcas:
@humberos I was looking at the same! This should work, will check this out

oshmoun:
NOTE: kitakami and shinano testers need to add the corresonding config to the ifdefs
testing now on kitakami

humberos:
@oshmoun actually the intention is testing the patch as it is. Since the rfkill hack was required for suzu (during bring up process).
If the deep sleep still broken then add platform to config.

oshmoun:
ah, thanks for the hint!
Tested by putting kitakami in ifdef. Deep sleep is still broken.
Testing without changes now.
Update: Deep sleep is still broken on suzuran at least.

humberos:
@erikcas 
deep sleep is still broken when BT is off ... huh ?
although the fix is closer but this patch needs more work yet :P

humberos:
Patch updated.

rcstar6696:
Do we need to add something that the pr works on shinano or kitakami?

humberos:
@rcstar6696 It should be tested as it is.

humberos:
Shinano is broken. Loire is still getting deep sleep issue when BT is turning on/off. So I will leave this change and start new one investigation asap.  Thanks for the feedback!

oshmoun:
for what it's worth, can confirm issue persists on kitakami as well with the patch

PR #994, opened 2016-11-19
[1.3.3] Update Bluesleep driver from 39.0.A.1.x

humberos:
Those patches will ~~fix BT power management and it~~ get all platforms aligned by using the same implementation as stock driver and AOSP similar with.


rcstar6696:
this will fix deep sleep?

humberos:
Yes, it requires https://github.com/sonyxperiadev/device-sony-shinano/pull/319

rcstar6696:
Confirmed :D Deep sleep works on sirius and castor_windy

oshmoun:
that sounds great!
the fact that it works on loire and shinano would highly suggest it will work on kitakami.
still, will test on kitakami when I have the chance.

humberos:
@oshmoun I am building for loire right now. Shinano is the only confirmed platform at moment.

lowtraxx:
The workaround seems to be needed on my shinano scorpion as well. If I try the PR without enabling it, bluetooth only turns on one in ten times and then takes about thirty seconds to do so.

humberos:
@lowtraxx interesting! 
Did you do a clean build ?
Also, did you pick the BT vendor change for shinano ( https://github.com/sonyxperiadev/device-sony-shinano/pull/319 ) ?

oshmoun:
according to the commit message of the workaround, it should apply to all bcm devices.
All bcm devices suffer from the same issue with line discipline after all.

humberos:
@jerpelea 
I think we can ship it as it is.
Since all BCM devices depends on line discipline (not implemented yet) then may the deep sleep state will be broken still.

Bluetooth works as expected.

This patch requires:
https://github.com/sonyxperiadev/device-sony-shinano/pull/319
https://github.com/sonyxperiadev/device-sony-loire/pull/49
https://github.com/sonyxperiadev/device-sony-kitakami/pull/218

PR #995, opened 2016-11-22
[1.3.3] USB work around

mcgi5sr2:
Works on Aries.


rcstar6696:
What was the problem?

mcgi5sr2:
Detection of whether it was charging. 

I have created the pull request from a fix offered by the author on xda. Before patching my phone would struggle to decide if it was charging or not when plugged into a pc.

jerpelea:
you are patching more than is needed.please cleanup your patch

munjeni:
Ok

jerpelea:
in this case the patch should be ifdefined for the RHINE platform only

PR #996, opened 2016-11-23
USB patch for Rhine and Shinano boards

mcgi5sr2:
Author:    munjeni


rancidfrog:
I would do proper test on this.
As on Aries phone detects charge when connected to PC usb.

The only time it did not detect was before bug regarding unresponsiveness after connecting to wall charger.
But, that has been patched already and usb charge works fine, as in gets detected when phone is connected to pc

jerpelea:
The patch should be ifdefined for the RHINE platform only

munjeni:
Just to write a comment...

On rhine usb is broken, charging source detection not working, when connects to the ac and disconnects usb locks and android reports charging when there is no charger and makes usb totaly malfunctional. But enabling CONFIG_SONY_USB_EXTENSIONS things is a lot better and everything working except charging source detection because charging source is always DWC3_CDP_CHARGER whatewer is connected to the phone usb port. Thats because rhine not support get_prop_proprietary_charger, this workaround handles that and detects if we are on ac or we are on usb. Also it fixes all problems which we have on rhine usb (excluding mhl). More info, more tests, more reports and finaly confirmation about our workaround can be seen on the folowing thread: http://forum.xda-developers.com/crossdevice-dev/sony/discussion-development-custom-rom-t3479108 . Yes patch is maily for rhine but can be used without any problem on ather boards which fail in get_prop_proprietary_charger , workaround will go skiped if get_prop_proprietary_charger function is not failed.

PR #1016, opened 2016-12-02
[1.3.3] soc: qcom: spm: Poll for the PMIC_STATE after updating the VCTL register

humberos:
The PMIC_STATE is expected to move to idle after the VCTL register
value is written to the PMIC.

Also update the PMIC_STATE poll timeout to 500us for all the targets.


humberos:
still testing.

humberos:
it is ok.
Sony (38.1.A.0.289F) uses it too.

PR #1018, opened 2016-12-05
[UM.5.5.r1] Bluesleep update

humberos:
Update bluesleep driver based on 39.0.A.1.x


humberos:
opss... I need to update the msm_serial_hs part.

PR #1019, opened 2016-12-05
[1.3.3] FM audio kitakami

alviteri:


jerpelea:
check again
https://github.com/sonyxperiadev/kernel/blob/aosp/LA.BR.1.3.3_rb2.14/arch/arm/mach-msm/Kconfig

PR #1021, opened 2016-12-06
drivers: qpnp-smbcharger: add somc current detection

oshmoun:
taken from kernel-copyleft 34.1.A.1.xxx

Change-Id: I1cbde6ded8d821cbe726bdc5b63d574eb4c1e00c
Signed-off-by: oshmoun <oshmoun100@googlemail.com>

### DIFF
--- a/drivers/power/qpnp-smbcharger.c
+++ b/drivers/power/qpnp-smbcharger.c
@@ -4441,6 +4441,7 @@ static int smbchg_change_usb_supply_type(struct smbchg_chip *chip,
 	 * modes, skip all BC 1.2 current if external typec is supported.
 	 * Note: for SDP supporting current based on USB notifications.
 	 */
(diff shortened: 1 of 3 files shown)


oshmoun:
this fixes wall charging for usb type c devices, like kugo.
The change also affects other devices with CONFIG_QPNP_SMBCHARGER_EXTENSION, so testing is needed. If it breaks other devices, we can ifdef for only usb type c devices.

alviteri:
https://github.com/sonyxperiadev/kernel/pull/1021/commits/5590912820b57a8f7ec861826f86b55091190374#diff-ba1a139305c66bfe5b7d6281e9969d88R4465

Use TABS? to kernel identation

humberos:
Apart from code style discussion, this patch does not affect _SUZU_ (I mean in a bad way).
Wall and USB changer are still working here.

PR #1026, opened 2016-12-07
[1.3.3] arm64: DT: kugo: Fix bluesleep

humberos:
Those DT entries are present in loire-common already.
But the gpios are different a bit.

So, move the common entry to suzu and fix the kugo one.

Change-Id: Ie06527f3c0423dae07aa8fe6857c2afed7bc77cb


oshmoun:
tested on kugo
everything works fine now
bt can be turned on (haven't tested pairing)
fmradio can start, and there is audio output

PR #1032, opened 2016-12-09
[LA.UM.5.5.r1] CVE-2016-3868 - msm-core: debug: Fix the number of arguments for sysfs nodes

pablomh:
Ptable and enable node parses the input arguments incorrectly. Parse the
input message into exact number of arguments that are required for the
respective nodes.

CRs-fixed: 1032875
Change-Id: I881f18217b703a497efa4799288dee39a28ea8ab
Signed-off-by: Archana Sathyakumar <asathyak@codeaurora.org>


pablomh:
Please commit as soon as possible because I think I broke kernel with previous merge nad this fixes it...

PR #1036, opened 2016-12-10
Misc Update

scafroglia93:
If missing a few commits let me know

screenshot_20161210-210937


scafroglia93:
All this commits are tested on my Z3

PR #1043, opened 2016-12-11
Media angler

alviteri:


alviteri:
into this ..... is needed to make intra-refresh work in media-codecs for android 7.0 and newer

PR #1044, opened 2016-12-11
Fix FM Audio path for Shinano

rcstar6696:
Shinano FM Radio sound only works if FM-Audio path is FM_AUDIO_DAC

### DIFF
--- a/drivers/bluetooth/broadcom/v4l2_fm_driver/fmdrv_config.h
+++ b/drivers/bluetooth/broadcom/v4l2_fm_driver/fmdrv_config.h
@@ -46,9 +46,11 @@
 
 /* Set default Audio path */
 #ifndef DEF_V4L2_FM_AUDIO_PATH
+#ifndef CONFIG_MACH_SONY_SHINANO
 #define DEF_V4L2_FM_AUDIO_PATH FM_AUDIO_I2S
 #else
 #define DEF_V4L2_FM_AUDIO_PATH FM_AUDIO_DAC
+#endif
 #endif
 
 /*Make this TRUE if FM I2S audio to be routed over */


rcstar6696:
@humberos @Myself5 @Diewi @lowtraxx

jerpelea:
nope
i have fm

Myself5:
You said you only have a static noise yesterday :/ 
Max' commit is fixing sound.

Also, might sumire also need that in order to work?
I know know if its using i2s or not, mind clearing that up :P?

jerpelea:
loire and kitakami are ok

rcstar6696:
interssting alin are you sure you have full working sound on shinano ? without this patch ?

humberos:
Fm is OK here (shinano Leo}
Actually the problem is the coherence between BT and ldisc, I mean it is
not possible to use Fm if bt is on and vice versa
Also, once bt is on you will need to turn it off, reboot device then Fm
will start normally.

Em dom, 11 de dez de 2016 19:51, Christian Oder <notifications@github.com>
escreveu:

rcstar6696:
hmm ok my sirius and castor only work with dac audio path

rcstar6696:
https://github.com/sonyxperiadev/kernel/blob/c5c05933ddb7982e735033d618fbadf55b10f572/drivers/bluetooth/broadcom/v4l2_fm_driver/fmdrv_config.h#L48 this couldn't be correct if its not defined define it as i2s and if defined as  dac

oshmoun:
Maybe this is only an issue on the first shinano generation (sirius, castor)?

Diewi:
Just syncing a 7.0 pdk according to the build guide to exclude other influences. I think, I can evaluate it tomorrow evening on my sirius.

humberos:
Checking stock sources (23.5.A.1.291)
http://developer.sonymobile.com/downloads/xperia-open-source-archives/open-source-archive-for-23-5-a-1-291/

then we have:

```
ifeq (,$(filter $(SEMC_PLATFORM), shinano))
KBUILD_OPTIONS += CONFIG_SOMC_FM_AUDIO_PATH=FM_AUDIO_I2S
endif
```

Diewi:
Ok, finally finished testing... uff.

So, using the build guide (7.0, of course; due to shinano), neither BT nor FMRadio are working: Dmesg shows an ldisc error (cannot work). I followed the build guide strictly. It makes no difference if I use the precomiled kernel, or compile it from source together with the rest of the pdk.

However, I got a better working build using the device tree + kernel from sonyxperiadev and the hals from sonyaosp toghether with android 7.1.1_r6: BT works & FMRadio finds stations.
Here, I only get audio for the FMRadio with Max' patch, without it it's just silence. I made two builds that only differ by this patch to verify.

alviteri:
@Diewi the problem was due to missing commits for Shinano in n-mr0 branch now is ok

rcstar6696:
David this is right but does not solve the audio issue on sirius and castor 
only if dac is enabled it works i tested it more than once now

Myself5:
So that gets us back to 2/2 sirius and castor users saying that DAC is
mandatory. Can't we just use DAC for those specific devices and let the
others use i2s?

jerpelea:
this should be the best option

alviteri:
@jerpelea only for sirius and castor then ?

alviteri:
https://github.com/sonyxperiadev/kernel/pull/1074

Diewi:
@alviteri Thanks for the update and the fix. Most of us use an abstraction layer on top of the build guide (manifest), but some other people might use it directly, so I wanted to not it here.

Reading the documentation again, audio is seperate. Hence, the change should fix the problem properly. Compared with the copleft kernel, there might have been some renaming?

rcstar6696:
now it should be the fix which is okay 
@jerpelea

alviteri:
@rcstar6696 leo needs it too

rcstar6696:
Leo is added now

alviteri:
probably is needed by all shinano ..... but I am not sure

humberos:
ok guys, I got my leo here then I will be able to check this issue again.
I will report about it asap.

drivers/bluetooth/broadcom/v4l2_fm_driver/fmdrv_config.h
```diff
@@ -46,9 +46,11 @@
 
 /* Set default Audio path */
 #ifndef DEF_V4L2_FM_AUDIO_PATH
```
humberos, line 48: Confirmed!
Shinano uses DAC
So the change would be something like this:

> #ifndef CONFIG_MACH_SONY_SHINANO
>     #define DEF_V4L2_FM_AUDIO_PATH FM_AUDIO_I2S
> #else
>     #define DEF_V4L2_FM_AUDIO_PATH FM_AUDIO_DAC
> #endif

Good work!
Happy 2017!

rcstar6696:
okay i will update it later 
now only the dwc wakelock needs a fix

drivers/bluetooth/broadcom/v4l2_fm_driver/fmdrv_config.h
rcstar6696: fixed

alviteri: why not used the way posted by humberto ? only 1 string need to be changed

drivers/bluetooth/broadcom/v4l2_fm_driver/fmdrv_config.h
```diff
@@ -46,10 +46,12 @@
...
+#ifndef CONFIG_MACH_SONY_SHINANO
 #define DEF_V4L2_FM_AUDIO_PATH FM_AUDIO_I2S
 #else
 #define DEF_V4L2_FM_AUDIO_PATH FM_AUDIO_DAC
 #endif
+#endif
```
humberos, line 54: ooopss! double #endif

rcstar6696: Its correct 
It is the endif or ifndef audio Path

humberos:
IMHO I think you could just replace

> #ifndef DEF_V4L2_FM_AUDIO_PATH
to
> #ifndef CONFIG_MACH_SONY_SHINANO

That's it

PR #1050, opened 2016-12-13
Io improvements

scafroglia93:


scafroglia93:
@rcstar6696

rcstar6696:
Tested on sirius and castor windy without problems

kholk:
The mmc clock scaling commits are useless. I have disabled it on the controller because of hardware issues. 

Also the writeback sync patch is no good: this is a backport from newer kernels which have api changes: the function being called as a replacement is different here and this will break NFS mounts, leading to eventual loss of data on those.

The tag block patch is also useless because we have no such device on our embedded boards and we will never have those, since we don't use raid, lvm and things like that which use tags. 

About the slub patch, I'm a little in doubt. 

The other ones are also backports from kernels with that different API on block/FS. 

The only good one for me seems to be the binfmt_elf patch, which should give a little performance improvement without breaking things.

scafroglia93:
No problem

scafroglia93:
No problem



 Sent with Mailtrack
<https://mailtrack.io/install?source=signature&lang=en&referral=scafroglia57@gmail.com&idSignature=22>

2016-12-27 14:30 GMT+01:00 AngeloGioacchino Del Regno <
notifications@github.com>:

kholk:
Since the binfmt_elf patch seems valid, please send it separately, so that it can be merged.
Thanks

PR #1057, opened 2016-12-15
pstore: Update sources and enable PSTORE on tone

AdrianDC:


jerpelea:
pstore should be imported from 39.2.A.

AdrianDC:
Wasn't aware open sources were released yet (Sony downloads, not copyleft),
will do in a few hours and test once again.

AdrianDC:
Commit description updated.
Sources are identical on 39.2.A.0.327, no commit changes done.

PR #1064, opened 2016-12-15
[DO NOT MERGE][REFERENCE] 1.3.3 spmi

alviteri:
this  fixes the warnings in build like https://github.com/sonyxperiadev/kernel/commit/72adc78d3588760a7f893b6404efd84ef6784515

but needs changes in thermanager.xml
like
https://github.com/sonyxperiadev/device-sony-honami/pull/46


alviteri:
ping @kholk @humberos @jerpelea

kholk:
Close that thing. We have those differencies because we use a patch that sets "More appropriate names" to SPMI devices. The patch is required because otherwise we would have name changes on SPMI devices at every kernel rebuild (almost).

The new QC patch does nothing good for us, it will still change the new "ID" every now and then.

PR #1065, opened 2016-12-16
[LA.UM.5.5.r1] Add cpuquiet driver from 1.3.3@3.10

humberos:
Add cpuquiet driver from 1.3.3@3.10 kernel.


kholk:
Check my GIT. I have a 3.18 cpuquiet framework and rqbalance there.......

PR #1067, opened 2016-12-16
[LA.UM.5.5.r1] Add cpuquiet framework

humberos:
Copied from @kholk 's gitspace

It is enabled for Suzu and Tone platform.


Kali-:
i guess we dont need it on 820

PR #1068, opened 2016-12-17
[LA.UM.5.5.r1] drivers: soc: loire: Fix boot process

humberos:
Add missed DT entry and #ifdef soc driver to match with DT

### DIFF
--- a/arch/arm/boot/dts/qcom/msm8956-loire-common.dtsi
+++ b/arch/arm/boot/dts/qcom/msm8956-loire-common.dtsi
@@ -929,7 +929,7 @@
 
 &pm8950_vadc {
 	chan@13 { /* case_therm */
-		qcom,scale-function = <15>; /* 100K_PULLUP_DECI */
+		qcom,scale-function = <17>; /* 100K_PULLUP_DECI */
 	};
 };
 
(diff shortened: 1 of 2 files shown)


humberos:
ping @kholk

humberos:
Not sure if we can still using 15 (for loire) or update it to 17

oshmoun:
same as on 3.10, the documentation doesn't seem to help :(
https://github.com/sonyxperiadev/kernel/blob/aosp/LA.UM.5.5.r1/Documentation/devicetree/bindings/thermal/qpnp-adc-tm.txt#L75

humberos:
Thanks @oshmoun 
I think 17 is ok since tone platform uses the same.
Also it matches with driver header.
Now we need librilqmiservices.so for RIL then I think radio/data will be ok at 7.1.1
WiFi (mmc sdio) is on my way.

stellirin:
~~According to: https://github.com/sonyxperiadev/kernel/blob/aosp/LA.BR.1.3.3_rb2.14/include/linux/qpnp/qpnp-adc.h#L274~~

~~It should now be 18 on all platforms~~

Wrong branch, 17 looks correct.

humberos:
Check it at 3.18

Em sáb, 17 de dez de 2016 21:44, Adam Farden <notifications@github.com>
escreveu:

PR #1079, opened 2016-12-22
[DO NOT MERGE] [TOCLEANUP] arm: DT: MSM8956: Create and assign MBA fixed memory device

kholk:
Most boards use an old firmware for modem and MBA. This fixed
memory device allows us to specify the address range where to
load the firmwares.


jerpelea:
not needed

PR #1080, opened 2016-12-22
[LA.UM.5.5.r1] drivers: mmc: Updates from Sony 39.2.A.0.361 and 3.10(1.3.3)

humberos:
1. Add support eMMC 5.0 from Sony
2. Add deferred MMC SD resume
3. Disable clock scaling for Sony devices
4. Add mmc low voltage if host has the capability.


kholk:
You should check out my GIT more frequently my friend! This is already done....why wasn't it already merged on the sonyxperiadev HEAD?!? 

BTW, the last commit about bkops enable/disable should be marked as CRITICAL and should be ported to 3.10 ASAP.

humberos:
@kholk 
Thanks!
Although mmc is ok for both platforms (loire and tone) but we still have some problem with sdio on loire platform.

From 3.10@1.3.3: after a long time reading the **CMD52** then mmc2 timing has been mapped with **6** (MMC_TIMING_UHS_SDR104)
> [   18.330969] mmc2: clock 200000000Hz busmode 1 powermode 2 cs 0 Vdd 7 width 2 timing 6
> [   18.349782] mmc2: sdhci_msm_set_uhs_signaling-clock:200000000 uhs mode:6 ctrl_2:0xb

From 3.18: this timing (UHS_SDR104) has not been recognized and it is required for bcmdhd (sdio).
> [   16.201140] mmc2: clock 400000Hz busmode 1 powermode 2 cs 0 Vdd 7 width 0 timing 0

Now that's the current problem with mmc/sdio driver for loire.

Again thank you and happy 2017 ;)

oshmoun:
according to some debugging done, it seems the detection is stopping here
https://github.com/sonyxperiadev/kernel/blob/aosp/LA.UM.5.5.r1/drivers/mmc/core/sdio.c#L1198
mmc_send_io_op_cond() is always returning -ETIMEDOUT, so it doesn't get to do anything concerning voltage, timing etc..

PR #1081, opened 2016-12-22
video: mdss: Color temperature interface using PCC

alviteri:
* MDSS5 supports Polynomial Color Correction. Use this to implement
   a simple sysfs API for adjusting RGB scaling values. This can be
   used to implement color temperature and other controls.
 * Why use this when we have KCAL? This code is dead simple, the
   interface is in the right place, and it allows for 128X accuracy.

Change-Id: Ie17c84ee3c1092ea65804566bdf05326a34a6d4d
Signed-off-by: Chirayu Desai <chirayudesai1@gmail.com>

### DIFF
--- a/drivers/video/msm/mdss/mdss_fb.c
+++ b/drivers/video/msm/mdss/mdss_fb.c
@@ -867,6 +867,68 @@ static ssize_t mdss_fb_get_dfps_mode(struct device *dev,
 	return ret;
 }
 
+static int pcc_r = 32768, pcc_g = 32768, pcc_b = 32768;
+static ssize_t mdss_get_rgb(struct device *dev,
+				struct device_attribute *attr, char *buf)
+{
+	return sprintf(buf, "%d %d %d\n", pcc_r, pcc_g, pcc_b);
+}
+
+/**
+ * simple color temperature interface using polynomial color correction
+ *
+ * input values are r/g/b adjustments from 0-32768 representing 0 -> 1
+ *
+ * example adjustment @ 3500K:
+ * 1.0000 / 0.5515 / 0.2520 = 32768 / 25828 / 17347
+ *
+ * reference chart:
+ * http://www.vendian.org/mncharity/dir3/blackbody/UnstableURLs/bbr_color.html
+ */
+static ssize_t mdss_set_rgb(struct device *dev,
+				struct device_attribute *attr,
+				const char *buf, size_t count)
+{
+	uint32_t r = 0, g = 0, b = 0;
+	struct mdp_pcc_cfg_data pcc_cfg;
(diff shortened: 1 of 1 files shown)


alviteri:
tested on honami with cm-14.1 (livedisplay and white balance regulator) work tested on aosp 7.1.1 with satsuki and Night Display (Google method) work display is still ambar ..

alviteri:
ping @adrianDC  @lowtraxx

stellirin:
I've used this even on yukon in the past, works well.

stellirin (approved):

lowtraxx (approved):
Works on scorpion as well

alviteri:
Have our own kernel fork

alviteri:
tested on kugo and working .... does nothing on aosp but works on LineageOS

lowtraxx:
Tested on scorpion with lineage and AOSP. Does nothing on AOSP but works on lineage.

kholk:
Big warning. You should either provide a way to get back to defaults or something, because whenever you set things through this code, it will DEFEAT the display calibrations provided in DT.

My calibration code uses PCC.

alviteri:
In aosp there is no way to manipulate this ... Only Los system can ... And yes there is a way to restore default settings

kholk:
Then, **on LineageOS**, it makes sense.

Yet, I think that the implementation is wrong. It should have a default setting restore handle in kernel.
IF someone implement a default settings restore handle in kernel, then it's acceptable for me, as Extended Settings in AOSP could be modified to make use of it.

Having it managed all on kernel side will permit literally any userspace to use it, not just AOSP and Android-based userspace.


This is Open Devices, so anything we have in this kernel, IMHO, should be usable by everyone and not specific to one android fork.
Do NOT forget that Open Devices aims to ALSO gives to users (and developers) a way to REPURPOSE devices, so we should make it simple for the final userspace to use new facilities.

rcstar6696:
You can change the values on aosp too you only need to install kernel auditor . the app has an interface for pcc changes too and it works i use it since the commit was mentioned the first time

kholk:
@rcstar6696 read again what I wrote..... :/

rcstar6696:
i read what you wrote :) it was only an info that the setting can changed on aosp too and kernel auditor saves all settings and they can be restored

kholk:
Unfortunately no, unless there is a specific save function in userspace (which I believe exists in LOS), so unless that program has an actual save function, whenever you write a value to pcc, you lose the defaults until reboot. 

Anyway, in AOSP it should be managed by Extended Settings with eventually preset color temperature profiles other than custom values. 

Whatever, the defaults should be restored via a kernel method, not userspace.

PR #1082, opened 2016-12-22
[LA.UM.5.5.r1] ARM: dts: msm: Add hyp log to tz log driver for msm8956

humberos:
Add hypervisor log device tree parameters to tz-log driver
to enable hypervisor log service.


humberos:
Fix:
[    5.683416] tz_log 8600720.tz-log: Hyp log service is not supported

Good to go.

PR #1083, opened 2016-12-22
[LA.UM.5.5.r1] ARM: dts: msm8956: Limit qseecom reserved memory to 32-bit

humberos:
Qseecom reserved memory needs to be in the 32-bit addressable range.
Use the alloc-ranges property to provide the limits.


kholk:
Are you sure?!

PR #1091, opened 2016-12-26
arm64: Kconfig: Increase UI responsiveness

alviteri:
following angler see:
                 https://android.googlesource.com/kernel/msm/+/521ae588c433b7b50b868eb61858d62fd6022998
                 https://android.googlesource.com/kernel/msm/+/9837c5a44098ccdcca58707c16cf39be1f2fbdb1

set 300HZ for ARCH_MSM8916 (tulip - suzu - kugo) and ARCH_MSM8994 (kitakami)

### DIFF
--- a/arch/arm64/Kconfig
+++ b/arch/arm64/Kconfig
@@ -595,6 +595,7 @@ config SWP_EMULATE
 
 config HZ
 	int
+	default 300 if ARCH_MSM8916 || ARCH_MSM8994
 	default 100
 
 config ARCH_HAS_HOLES_MEMORYMODEL


kholk:
We (and angler,  too) use nohz, I really don't understand the reason for tuning the hz parameter.

kholk:
Sorry for double comment, btw, remember that we also use hardware high resolution timers that are actually giving us a pretty darn good wake up granularity on scheduling operations. IMHO, raising the hz will only affect power consumption (Scheduler will work more than needed, software wakeups from WFI loop on CPUs will be triplicated).

alviteri:
probably you don't use the devices to make specific operations like games but sometimes in few games there are explicit lags ... this help alleviate them ...

PR #1094, opened 2016-12-30
[1.3.3] [MSM8974] kernel compression

alviteri:
use XZ for kernel compression
add BZIP2 and LZMA ramdisk compression


alviteri:
boot.img and recovery.img have better size 
also fixes big ramdisk on cm/los recovery.img

ping @AdFad666 @adrianDC

alviteri:
ping @kholk

scafroglia93:
@alviteri Great PR, XZ it's my default kernel kernel compression

scafroglia93:
@jerpelea 

This is my old custom kernel for SonyAosp (Z3) XZ was the default kernel compression

https://github.com/CasualsKernelProject/leo/commit/7ca68c9c10c4b1fcbdb065bf0abdcf177198f95a

Save about 4MB

kholk:
Xz compression is cpu intensive. We have lots of space and, at boot, eMMC is initialized in HS by bootloader. Our bottleneck for boot time from bootloader is CPU, so this patch will make it worse.

scafroglia93:
Yep, you're right about cpu stress for XZ

alviteri:
So good bye custom support like cm recovery and rest close

alviteri:
What about LZMA ? @scafroglia93

stellirin:
Recovery lives in its own partition but it is actually the boot kernel that loads the ramdisk from the recovery partition. The recovery kernel does not need to be the full kernel. You could simply build a stripped down kernel with almost no drivers as the kernel in the recovery partition is never used.

PR #1095, opened 2016-12-30
Fixup Deep Sleep in combination with bluesleep.

Diewi:
Due to the inclusion of FMRadio, the wake mechanism in the userspace
moved from using the procfs to the ioctl assertion mechanism. This
commit makes the bluesleep and the msm uart driver fully aware of the
ioctl assertion mechanism. The basic idea is to trigger the bluesleep
functions that are called during assertion/deassertion via the procfs
from the msm_serial_hs driver whenever an assert/deassert ioctl is
delivered.

NOTE: This change requires modifications to the libbt configuration
to operate correctly (remove the wake via procfs).

In particular, the following changes have been made:

* Create a bluesleep header file to allow the msm_serial_hs driver to
access the bluesleep assert/deassert functions.

* msm_serial_hs: Make the msm_serial_hs driver to call the bluesleep
start/stop functions upon registration/deregistration. Also call the
assert/deassert functions (bluesleep_outgoing_data() /
bluesleep_tx_allow_sleep()) when the corresponding ioctl commands are
received.

* bluesleep: Make the bluesleep_tx_allow_sleep function externally
accessible.

* bluesleep: Introduce a state tracker variable for uart clock on/off
requests and use it in the hsuart_power method. Further, remove any
direct invocations of msm_hs_request_clock_on/off.

* bluesleep: Add a state flag to the bluesleep_start function to signal
the context from which the bluesleep driver is called: 0 indicates that
no clock has been asserted before, while 1 signals the opposite. This
flag is required to correctly set the initial state which enables deep
sleep and FMRadio with the current limitations. This flag should be no
longer needer once the FMRadio driver uses the standard ldisc.

Change-Id: I50eb35b6af93f6bebb2122b3f80061f04b120998

### DIFF
--- a/drivers/bluetooth/bcm43xx.c
+++ b/drivers/bluetooth/bcm43xx.c
@@ -27,6 +27,9 @@
 #include <linux/platform_device.h>
 #include <linux/rfkill.h>
 #include <linux/slab.h>
+#ifdef CONFIG_BT_MSM_SLEEP
+#include <net/bluetooth/bluesleep.h>
+#endif
 
 #define D_BCM_BLUETOOTH_CONFIG_MATCH_TABLE   "bcm,bcm43xx"
 
@@ -60,6 +63,10 @@ static int bcm43xx_bt_rfkill_set_power(void *data, bool blocked)
 			return 0;
 		}
 		gpio_set_value(bcm43xx_my_data->reg_on_gpio, 1);
+
+#ifdef CONFIG_BT_MSM_SLEEP
+		bluesleep_start(1);
+#endif
 	} else {
 		if (!regOnGpio) {
 			pr_debug("Bluetooth device is already power off:%d\n",
@@ -67,6 +74,10 @@ static int bcm43xx_bt_rfkill_set_power(void *data, bool blocked)
 			return 0;
 		}
 		gpio_set_value(bcm43xx_my_data->reg_on_gpio, 0);
+
+#ifdef CONFIG_BT_MSM_SLEEP
+		bluesleep_stop();
+#endif
 	}
 	bt_enabled = !blocked;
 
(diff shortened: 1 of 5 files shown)


kholk:
Patch seems good, but the bluesleep_start function parameter should be a bool and not an int, since only 0 and 1 are used, this will allow less memory to be wasted.

Diewi:
Thanks for you review, I have made the parameter change.

Diewi:
Another update that should solve the problem of broken deep sleep after BT is turned OFF (when no devices are paired) which is discussed here:
https://github.com/sonyxperiadev/device-sony-shinano/pull/336

The update moves the calls to bluetooth_start/stop from the msm_serial_hs driver to the bcm43xx rfkill driver such that bluesleep is correctly (de-)initialized upon BT enable/disabled.

nailyk-fr:
removed, wrong place.

Diewi:
Updated commit such that the msm_serial_hs driver notifies the bluesleep driver of outgoing data if data is transmitted. Although this is obviously required, I missed this one during testing (too short tests before running into the transmit timeout). Pairing and music playback seems reliable now.

Diewi:
Update: Fix pairing bug with some devices.
Instead of scheduling the sleep worker directly when a deassert command from userspace comes in, fire up the tx timer (which then issues the sleep worker if needed). By this, the BT sleep does not kick in during pairing where for some reasons, a deassert is sent from userspace.

nailyk-fr:
Next time I will wrote every problem I encounter :)
Before PS2 I was not able to pair to my BT v4.0 stick but only the old one (v2 iirc). 
Now it pair fine on both computer stick, mouse and a2dp working. I have no other peripherals to test sorry. 

Thanks to you, Humberos and all of the dev who made this kernel for fixing BT :)

PR #1097, opened 2016-12-30
1.3.3 scripts

alviteri:
ping @jerpelea @kholk @Kali- @adrianDC


kholk:
Never tried initramfs generation from kernel scripts. We actually use Android build system's own crap for this..... 

I don't know if this is valid or not, but it is generally a good idea to leave things as they are, especially if it is a forward port from newer kernels.

PR #1102, opened 2017-01-02
drivers: serial_msm_hs: Fix BT FM

humberos:
This should be removed for BT FM until we find out the proper fix for
bluetooth (deep sleep state and BT timing for new paired device).

Change-Id: I57f8cf46be50d4b40ef3c8d064facd6ef72a9515

### DIFF
--- a/drivers/tty/serial/msm_serial_hs.c
+++ b/drivers/tty/serial/msm_serial_hs.c
@@ -1443,11 +1443,6 @@ static void msm_hs_submit_tx_locked(struct uart_port *uport)
 	mod_timer(&tx->tx_timeout_timer,
 		jiffies + msecs_to_jiffies(MSEC_PER_SEC));
 
-	/* Notify the bluesleep driver of outgoing data, if available. */
-#ifdef CONFIG_BT_MSM_SLEEP
-	bluesleep_outgoing_data();
-#endif
-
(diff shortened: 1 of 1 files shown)


humberos:
ping @Diewi

humberos:
unfortunately without this call the deep sleep state issue is back :(
Also, I have noticed some problems during pair process.
Not sure but I am thinking may we should patch libbt-vendor instead of kernel. I mean about ioctl vs procfs.

Diewi:
Just a moment, I am working on a fix for the BT FM issue which maintains the deep sleep functionality. I need to verify the correctness of the solution (tests), but it looks good up to now. I'll keep you informed.

humberos:
awesome!
thanks!

humberos:
I did some tests here and seems the deep sleep is ok.
Now I am checking BT pairing process.

humberos:
the proper fix is coming.
Thanks @Diewi !

humberos:
Seems that it is still required for bluetooth + watch pairing process.
After some time my watch can't sync data with phone and the BT connection is lost from watch.
From phone BT is still working and saying that my watch is still connected ???
It seems a tx timeout problem.
Please hold on since I am testing audio streaming right now.

humberos:
it breaks audio playback.... bye bye!

PR #1105, opened 2017-01-03
drivers: bluesleep: Fix BT FMRadio kernel panic.

Diewi:
Instead of relying on the bluesleep worker function, power up UART
directly from bluesleep_start. This avoids inconsistent states that lead
to kernel panics most times when the FMRadio app is opened.
Further shift the clock power up further below to avoid crashes when
opening FMRadio while BT is still on.

In addition, some following changes have been made to get the FMRadio to
operate close to what is expected. The current limitations are that BT
and FMRadio cannot be activated simultaneously, and a playback pause of
some seconds that occurs when turning the device on again via the power
button while the FMRadio is playing.
In particular, these changes include:
* bluesleep: open_count: invert the open count logic to actually reflect
opening (start) of the bluesleep driver.
* bluesleep: hsuart_power: if turning off the uart clock fails due to
already turned off clocks (via calls from other modules), also decrase
the uart_clock state counter.
* bluesleep: bluesleep_outgoing_data: always reset the tx timer, if
outgoing data is delivered (robustness).
* bluesleep: bluesleep_stop: place the uart state counter correctly to
avoid erroneous behavior when hsuart_power is called.
* brcm_bluesleep: brcm_btsleep_start: return the return value from 
bluesleep_start. Needed for the FMRadio driver.
* brcom_ldisc_sh: brcm_sh_ldisc_unregister: Add a flag to determine
whether the bluesleep module could be started successfully before
unregistering.
* brcom_ldisc_sh: brcm_sh_ldisc_start: Do not initialize the ldisc if
bluesleep is already opened (by BT). Workaround until the migration to
the standard ldisc driver.

Change-Id: I4e534f8b54832914b661b08c7505a958cd363ed2

### DIFF
--- a/drivers/bluetooth/bluesleep.c
+++ b/drivers/bluetooth/bluesleep.c
@@ -141,7 +141,7 @@ static struct platform_device *bluesleep_uart_dev;
 static struct bluesleep_info *bsi;
 
 /* module usage */
-static atomic_t open_count = ATOMIC_INIT(1);
+static atomic_t open_count = ATOMIC_INIT(0);
 
 /*
  * Global variables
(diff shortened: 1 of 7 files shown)


oshmoun:
Just a suggestion for the last point about the workaround. Maybe put the workaround in a seperate commit for easier reverting later when it's not needed?

Diewi:
Valid point, but to my understanding the brcom_ldisc driver would vanish then. This is also where the workaround is placed. Hence, I think that a revert will not be needed then.

PR #1108, opened 2017-01-04
[UM 5.5.r1] bluesleep patches

Diewi:
Port of the two commits from the 1.3.3 kernel that remove the constant bluesleep wakelock and hence allow deep sleep while BT is on.

- The first commit fixes the bluesleep wakeup.
- The second commit fixes fixes crashes of FMRadio that were introduced by the first commit.

Since I do not have a device that is capable of running a 3.18 kernel, this patch should be tested by someone owning such a device.


jerpelea:
cherry-picked

PR #1111, opened 2017-01-05
arm64: enable MSM Performance module

stellirin:
We need msm_performance as it is the ONLY reliable way
we have for HMP cluster throttling (max cpufreq limiting).

on HMP all cores of a cluster can go offline so we lose userspace
permissions on the key sysfs. msm_performance never goes 'offline'
so we can set maximum frequencies via this module.

This is therefore needed by thermanager.

Signed-off-by: Adam Farden <adam@farden.cz>


jerpelea:
cherry-pick and rebase

PR #1112, opened 2017-01-06
[1.3.3] Fix rhine jack

SuperLamic:
Disable enable flags and use values that are marked as "noise dumping jack". It would make sense to use non-noise dumping values and they are actually used, but they are wrong. Detection does not work and all headphones are detected as 4-pole ones which results in not working phone mic (phone tries to use not existing headphones one) and phantom button presses.

Also correct values for headset buttons that are also wrong - all buttons act as one button - you can't change volume.


rancidfrog:
I believe shinano also suffers from phantom button presses, as music starts playing after i pause it.
Z3 compact

SuperLamic:
@rancidfrog may be true, but it must be different reason. https://github.com/sonyxperiadev/kernel-copyleft/blob/23.4.A.0.xxx/sound/soc/msm/msm8974.c#L1882 
plug detection values are correct

PR #1113, opened 2017-01-07
[1.3.3] CFQ Improvements

scafroglia93:
- Compile on arm / arm64 is ok. There's no error
- Introduce a tuned CFQ made by AK for better perf on mobile device

That's all


kholk:
The values are in function of HZ for a very good reason. This commit will make the calculations imprecise. 

The format string mismatch commit should be good.

scafroglia93:
Rebase done

PR #1115, opened 2017-01-07
[1.3.3] Random Improvements

scafroglia93:
- This don't break compile on arm and arm64
- Boot fine on Z3, wait for tester with arm64
- Performance improvements detected with androbench test

### DIFF
--- a/drivers/char/Makefile
+++ b/drivers/char/Makefile
@@ -2,7 +2,7 @@
 # Makefile for the kernel character device drivers.
 #
 
-obj-y				+= mem.o random.o
+obj-y				+= mem.o random.o frandom.o
 obj-$(CONFIG_TTY_PRINTK)	+= ttyprintk.o
 obj-y				+= misc.o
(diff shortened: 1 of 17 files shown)


kholk:
This will expose the system to security risks. The usage of a nonblocking pool in random number generation can get hijacked because you have more or less a fixed memory range in which the number gets calculated (random data is taken from  memory too, hence an user may be able to inject data in a specific memory address to make the random number predictable).

About the latter frandom (imprecise) implementation, it has got the same security risks and then it and why are you hacking the standard random functions in kernel when you then want to use the frandom implementation. 
I would propose to leave the standard kernel implementation in a blocking pool and then making an ifdef OR a sysfs option to make the user to be able to switch between frandom or standard Linux randomly, explicitly saying the security risks involved with the new frandom. User will decide. 


About me, I would personally use frandom most of the times, but it still depends on where I am and what I'm doing.

PR #1119, opened 2017-01-08
Revert "[DWC3] Add missing return at the end of function dwc3_chg_det…

ergoen:
…ect_work"

This reverts commit b27cb0a9d95e7295104cf5499c2da01703c4c0de.

This commit introduced a bug where a wakelock would get stuck on amami (probably all 8974 devices, can't personally test any others though). After booting the phone, connecting it to a charger (either ac or usb) and the disconnecting it resulted in an msm_dwc3 wakelock getting stuck. Reverting it fixes the problem.

### DIFF
--- a/drivers/usb/dwc3/dwc3-msm.c
+++ b/drivers/usb/dwc3/dwc3-msm.c
@@ -1796,7 +1796,6 @@ static void dwc3_chg_detect_work(struct work_struct *w)
 			chg_to_string(mdwc->charger.chg_type));
 		mdwc->charger.notify_detection_complete(mdwc->otg_xceiv->otg,
 								&mdwc->charger);
-								return;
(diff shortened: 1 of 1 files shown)


rcstar6696:
This causes the wakelock but the return is needed for charging detection 
We currently work on it
 its an issue in pm runtime which causes the wakelock

ergoen:
I guess the pm_runtime_put_sync line gets called as well now by falling through to the next case, does that break detection?

rcstar6696:
The pm runtime in general is the issue and was not present before
There are some other commits in otg driver which has exactly an issue with pm runtime and if you put a pm relax before the return the wakelock is gone but other event logs pop up 

The return in general comes from 1.2.2 dwc driver

PR #1122, opened 2017-01-09
[1.3.3] Crypto

scafroglia93:
@kholk 

- Compile arm / arm64 OK
- Tested on Z3
- Improvements detected

For changing the appearance defconfig ok otherwise be wasted time

Wait for your revision boss


scafroglia93:
@rcstar6696

rcstar6696:
It isuseless for arm devices 
its arm64 only
 it depends on arch arm64 cofing
 i only have 32 bit

jerpelea:
with this one we should stick to the default implementation or we can endup with strange issues

PR #1127, opened 2017-01-10
arm64: DT: Enable regulator in VFE node

alviteri:
following commit 953a0b772770ac51fc9474428514ea14ffe842f1
VFE clocks need VFE regulators to be enabled to turn on clocks


alviteri:
ping @kholk

kholk:
Should be good


...wait. You should check the legacy one.

PR #1128, opened 2017-01-11
[1.3.3] Fix

scafroglia93:
Z3 OK

Wait @kholk for review


rcstar6696:
Works but no noticble diffrence

scafroglia93:
They are small changes that together make

PR #1130, opened 2017-01-11
[1.3.3][WIP] MSM8974 DWC3

alviteri:
already working

Rhine ok (tested only honami) otg/usb/charger working NO WAKELOCKS

Shinano (tested only leo) usb/charger ok, OTG kill usb/charger after unplug (reboot fix it) NO WAKELOCKS


nailyk-fr (commented):
Restore deepsleep on my leo too (obviously)

alviteri:
@nailyk-fr we need check why mhl is not stopping otg

hahnjo:
This should also work for amami, I've found the same "fix" locally

PR #1134, opened 2017-01-12
[1.3.3] qdsp6v2 optimize size

scafroglia93:
- Reduce kernel size (Z3 with this is 8.6MB)
- Compile arm / arm64 ok 
- Tested on Z3


scafroglia93:
@rcstar6696 Have you tested this on your device ?

rcstar6696:
Yes i habe but no diffrence im kernel size

jerpelea:
this commit is not having any benefits

scafroglia93:
Just only for info

schermata del 2017-01-15 19-11-39

kholk:
This has no benefits in our specific case.

Moreover, forcing GCC flags in makefiles is the WORST HACK EVER and should be avoided at every cost.

PR #1143, opened 2017-01-15
[1.3.3] acdb (In call audio)

mcgi5sr2:
Required commits for acdb tested on Shinano with the corresponding blobs


alviteri:
these were close before ... see @humberos ' PR

alviteri:
alin want OSS way for this

mcgi5sr2:
Ok i understand OSS may give a better result, as there is still echo on the line using this. However i would live this open until we have a better way. That allows people to have in call volume control on their own current builds.

alviteri:
As I know this doesn't work without acdb blobs

mcgi5sr2:
yes correct without the blobs it doesn't work

PR #1149, opened 2017-01-17
[1.3.3] BLK Upstream

scafroglia93:
Wait for revision on arm64

Z3 Boot fine


scafroglia93:
@rcstar6696 Have you tested this commits ? I use this set of commits for my Nexus 6P and 5X stable kernel

mcgi5sr2:
fine on aries

scafroglia93:
wait for revision

kholk:
I said this a billion times now I think. Random commits from other kernel version are likely to invisibly break things (problems will show in some time). In this specific case, what can happen is some code depending on other code introduced by an apparently unrelated commit that, if not applied, can produce race conditions and data loss.

Issue #1146, opened 2017-01-15
[1.3.3][D6633] Reboot in call


nailyk-fr:
On LineageOS leo experience reboot in call after screen goes dark. (Aries doesn't have this issue. )
This issue doesn't happens while plugged into usb, even after merging https://github.com/sonyxperiadev/kernel/pull/1130. 
If prevent screen going dark (setting timeout at 30min) and using headset for example, phone doesn't reboot until 30 minutes are reached. 

Don't know if this issue is related to https://github.com/sonyxperiadev/kernel/issues/797 or is relevant because of Lineage. 

Issue present on each LOS build. 

Partial output with this [kernel](https://github.com/nailyk-fr/kernel/tree/test) (include #1131 #1130 #1128 #1120): 
```
[  496.498192] Error: returning -512 value
[  496.513756] Unable to handle kernel NULL pointer dereference at virtual address 0000001c
[  496.513769] pgd = c3a78000
[  496.513788] [0000001c] *pgd=00000000
[  496.513810] Internal error: Oops: 5 [#1] PREEMPT SMP ARM
[  496.513839] CPU: 1 PID: 1331 Comm: system_server Not tainted 3.10.84-gaeaeb5a #1
[  496.513858] task: c4411f40 ti: c3ad8000 task.ti: c3ad8000
[  496.513888] PC is at snd_pcm_do_suspend+0x44/0x54
[  496.513907] LR is at snd_pcm_action_single+0x38/0x64
[  496.513929] pc : [<c0752cb0>]    lr : [<c0752910>]    psr: 60000093
[  496.513929] sp : c3ad9cd0  ip : 0000e64b  fp : c0e7bfec
[  496.513943] r10: 00000002  r9 : c0c7822c  r8 : c03a740c
[  496.513958] r7 : c0f9c914  r6 : 00000000  r5 : c0f8b080  r4 : c426ee00
[  496.513972] r3 : c426ee00  r2 : 00000000  r1 : 00000005  r0 : c426ee00
[  496.513993] Flags: nZCv  IRQs off  FIQs on  Mode SVC_32  ISA ARM  Segment user
[  496.514011] Control: 10c5787d  Table: 03a7806a  DAC: 00000015
[  496.514028] Process system_server (pid: 1331, stack limit = 0xc3ad8238)
[...]
[  496.515401] ---[ end trace 749da69a7594b5f2 ]---
[  496.537614] Kernel panic - not syncing: Fatal exception
[...]

PR #1150, opened 2017-01-17
[1.3.3] Fix in call reboot for shinano

nailyk-fr:
* MI2S links are used while in call.
   This prevent kernel enter sleepmode while in call which cause a segfault.

Change-Id: I63be8ecbc55301e464f319622a440d5397cb0113


nailyk-fr:
Related to [this](https://github.com/sonyxperiadev/kernel/issues/1146) issue. 
However I think this is a workaround as device shouldn't try to enter sleep mode while in call.

kholk:
This will avoid sleep also while listening to music etc. Not good at all!

nailyk-fr:
Agree with you, this is not the best solution but make the device usable. Also I'm not able to debug deepsleep. 
Should-I make a uglier hack by preventing sleeping on TX only or adding 'temporary' into the commit message?

PR #1151, opened 2017-01-20
[1.3.3] CPUQuiet & interactive improvement

RyTekk:


scafroglia93:
We need some improvements on cpuquiet, on Z3 drain much battery. Sometimes it's stuck at maxfreq for 5s +

kholk:
The only piece of code that we *may* use in this pr is... One line. 
We are not using any other because we use rqbalance (which, btw, already refuses to put CPU0 offline).

PR #1162, opened 2017-02-13
[WIP] BT/FMRadio using the broadcom ldisc driver

Diewi:
These commits allow us to use the broadcom ldisc driver for BT/FMRadio without any changes to libbt etc. The only additional requirements are some few changes to the device tree:
- "#define HCILP_INCLUDED false" in bdroid_buildcfg.h
- "UartPort = /dev/brcm_bt_dev" in bt_vendor.conf
- "UimUartPort = /dev/ttyHS0" in bt_vendor.conf (for msm_serial_hs)

The PR consists of 2 commits:
- A bugfix.
- The actual commit which intercepts configuration HCI commands from libbt with few related changes.

Bug List:
- Bug (1)
  BT may not be enabled automatically after startup (Toggling the BT switch enables it).
- Bug (2)
  If BT is switched to disabled once, and the device enters the "Deep Sleep" state, FMRadio does not work afterwards, EXCEPT if the FMRadio app was launched after setting BT to disabled before the device enters "Deep Sleep".
- Bug (3)
  See comment 1
- Bug (4)
  Kernel panic when disabling BT after "Deep Sleep" from the lockscreen. See comment 3

### DIFF
--- a/drivers/bluetooth/bcm43xx.c
+++ b/drivers/bluetooth/bcm43xx.c
@@ -64,7 +64,7 @@ static int bcm43xx_bt_rfkill_set_power(void *data, bool blocked)
 		}
 		gpio_set_value(bcm43xx_my_data->reg_on_gpio, 1);
 
-#ifdef CONFIG_BT_MSM_SLEEP
+#if defined(CONFIG_BT_MSM_SLEEP) && !defined(CONFIG_LINE_DISCIPLINE_DRIVER)
 		bluesleep_start(1);
 #endif
 	} else {
@@ -75,7 +75,7 @@ static int bcm43xx_bt_rfkill_set_power(void *data, bool blocked)
 		}
 		gpio_set_value(bcm43xx_my_data->reg_on_gpio, 0);
 
-#ifdef CONFIG_BT_MSM_SLEEP
+#if defined(CONFIG_BT_MSM_SLEEP) && !defined(CONFIG_LINE_DISCIPLINE_DRIVER)
 		bluesleep_stop();
 #endif
 	}
--- a/drivers/bluetooth/bluesleep.c
+++ b/drivers/bluetooth/bluesleep.c
@@ -450,16 +450,17 @@ int bluesleep_start(bool is_clock_enabled)
 	clear_bit(BT_TXDATA, &flags);
 #endif
 
+	clear_bit(BT_ASLEEP, &flags);
+
+	spin_unlock_irqrestore(&rw_lock, irq_flags);
+
 	// For ldisc-controlled BT, the clock is enabled by upper layers, so
 	// make bluesleep aware of this state.
-	clear_bit(BT_ASLEEP, &flags);
 	if(is_clock_enabled) {
 		atomic_set(&uart_is_on, 1);
 	} else {
 		hsuart_power(HS_UART_ON);
 	}
-
-	spin_unlock_irqrestore(&rw_lock, irq_flags);
 
 	enable_wakeup_irq(1);
 	set_bit(BT_PROTO, &flags);
--- a/drivers/bluetooth/broadcom/line_discipline_driver/brcm_sh_ldisc.c
+++ b/drivers/bluetooth/broadcom/line_discipline_driver/brcm_sh_ldisc.c
@@ -187,6 +187,16 @@ static struct nlmsghdr *nlh;
 static struct sk_buff *hcisnoop_skb_out;
 #endif
 
+/* HIC commands emitted by libbt which shall be ignored. */
+#define IGNORE_CMD_SIZE 14
+static const short IGNORE_CMDS[IGNORE_CMD_SIZE] =
+                                      {0x0c03, 0xfc45, 0xfc18, 0x0c14, 0xfc2e,
+                                       0xfc01, 0xfc27, 0xfc1c, 0xfc1e, 0xfc6d,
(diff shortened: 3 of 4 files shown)


humberos:
Hi!
Nice patch set!
First of all, thank you!

I've tested it on suzu and I can use BT and FM Radio simultaneously.
Good job! Awesome!
Apart from known issues I got new one issue that I will describe below:

  1. Restart device with bluetooth disabled
  2. After restart then open the FMRadio App (it works fine).
  3. Enable bluetooth then FMRadio will stop and bluetooth will no turn ON anymore.
  4. Fix: restart device

PS: I am using Pebble watch paired with my device and when it gets the deep sleep status my watch the pairing but device is still showing up the connection ok. This issue is not related directly to this patch set. It is present by using current sources (it is also present by using this new patch set).

I will do more tests then report back if I find something.

Again, thank you!

Diewi:
Hi @humberos, many thanks for testing!
I can forward a large portion of the thanks to @Quarx2k who first implemented the support for combined BT/FMRadio by modifying system_bt and later libbt. What I did was mainly to reduce the needed modifications to kernel and device tree changes.

Regarding the addtional issue: I could reproduce it and have an idea of how to fix it but I need to experiment a bit during the next one or two days. I also added a short list of known issues to the PR message, so we can trace them more easily.

jerpelea:
bug !
leave BT on 
pt the phone to sleep
wake the phone 
on the lockscreen pull the settings slider and press the BT icon 
....reboot

humberos:
@Diewi 
@jerpelea 
I've got BT procfs back then seems that FMRadio and BT are performing better.
But it breaks deep sleep state which I think may is possible to be fixed on kernel space (not sure).

BT procfs patch:
https://gerrit.omnirom.org/#/c/22010/

~~Also this change fixes BT pairing issue (that I've mentioned before).~~
still broken... too fast... :(

humberos:
@Diewi 
I can confirm the bug that @jerpelea have mentioned.

Kernel panic:
https://gist.github.com/humberos/012a1cf163dfc5aa1ada79edc5c99476

Diewi:
@jerpelea @humberos: Thanks for testing and finding the bug. I will make a fresh build tonight to see if it is some rfkill issue (I could not see the issue; probably some messed up build). From the dmesg (thanks, BTW) it is quite obvious that the uart line is shut down which causes the bug and makes me think of being a rfkill issue. I'll drop a note here, when I identified the cause.

@humberos Afaik, the wake mechanism (defined in bt_vendor.conf) should be either via IOCTLs or via the proc interface exclusively (see the usage of the defines in "libbt -> upio_vendor.c" if I rember correctly). Please, correct me if I am wrong.
From my observations, the define "HCILP_INCLUDED FALSE" in the platforms' "bdroid_buildcfg.h" should prevent libbt from emitting wake (de-)asserts. To test this, I called the msm_serial_hs' ioctl function whenever some IOCTL was received by the broadcom driver. A log statement in the msm_serial_hs' function allowed me to conclude what I wrote above.

Quarx2k:
@Diewi: Yea, rfkill can cause this problem. Check this patch https://github.com/LineageOS/android_hardware_broadcom_libbt/commit/4ee210c91d1c711a67b032a1e48b0d7aa8d40ac7  (except BtDrvDev change, moved to uim). This is all what need to get fm_bt via userspace part.  Upio_set_bluetooth_power use rfkill and make everything broken.

Diewi:
@Quarx2k Thanks for the confirmation!

Since Upio_set_bluetooth_power evaluates the property "ro.rfkilldisabled" before triggering the rfkill switch, we can just set this property in the device tree, and modify the UIM service to evaluate a differenty property, e.g., "ro.rfkilldisableduim". Then, bug (4) should be resolved. I will test it this evening.

UPDATE:
Bug (4) is not caused by rfkill, but the HCI command filtering. To test this, I applied the libbt patch (incl. some small modifications -> UartPortUim) and commented out the HCI command filter (in brcm_sh_ldisc.sh): the broadcom ldisc driver is still used due the libbt patch. Then, the device did not crash. (It was a bit disgusting to debug, since the error does not occur always: For instance, if the FMRadio is launched before deep sleep).
I will add some switch such that the filter only applies when BT is strarting up.

BTW, setting the ro.rfkilldisabled worked to prevent libbt to trigger the rfkill switch. Maybe we need it later...

Diewi:
Updated the commit such that the HCI command filter is only active after enabling bluetooth to intercept the configuration commands from libbt. The end of this sequence is detected and the filter is disabled then.

In my tests, this change solved bugs (2), (3), and (4). I used a clean build (sirius) according to the build guide to do the tests. I tested for the mentioned issues and ran the build for one day and did not observe some faulty behaviour (except the problem with watches etc. that @humberos mentioned). Anyways, it would be nice if someone else could test these updated commits, to check if I overlooked something.

Also, I could not observe bug (1) any more. It seems to have correlated with some other issue of the test rom I used at that time. Can someone else observer BT not turing on after startup (if it was enabled before shutdown)?

alviteri:
@Diewi I want kill you .... I have built one hour ago xD xD xD ... let me rebuild ... thanks for this

alviteri:
tested on kugo

turn on device
enable BT
enable FM
disable FM
display off
enters to deep sleep...

open FM again (No Speaker selection - only headset available (don't know if related)

PR #1168, opened 2017-02-20
Xperia XZ/Kagura unresponsive touchscreen fix

hwchong:
This patch fixes Xperia XZ/Kagura unresponsive touchscreen after a device wakes from sleep mode.


jerpelea:
cherry-picked and cleaned ;)

PR #1173, opened 2017-02-22
ARM: dts: msm: Add bus vectors in cpp node

alviteri:
Add bus vector in cpp node for independent
AB/IB voting. Previously this was being done
via isp.

Change-Id: I3804a72d705295645ef5321d995a8dcab6e395a9
Signed-off-by: Krupal Divvela <kdivvela@codeaurora.org>
Signed-off-by: Neelam Abhinav Karthik <aneelam@codeaurora.org>

### DIFF
--- a/arch/arm/boot/dts/qcom/msm8976-camera.dtsi
+++ b/arch/arm/boot/dts/qcom/msm8976-camera.dtsi
@@ -390,7 +390,14 @@
 			"camss_vfe_cpp_clk","micro_iface_clk", "camss_ahb_clk";
 		qcom,clock-rates = <0 240000000 0 0 240000000 0 0>;
 		qcom,min-clock-rate = <160000000>;
-		bus_master = <1>;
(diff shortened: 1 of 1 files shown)


kholk:
As much clever as it sounds, the first commit is, in reality, a really stupid idea.
We can get as many bad things as you can imagine, with the most obvious being a clock mess in init path (example: you init CSI0, then init CSI1 while closing CSI0) with not just "possible" but LIKELY race conditions, producing eventually hardware faults or battery WASTE.

The second commit (bus vectors for CPP) will conflict with my other PR, but it's better than my interpretation of it, so.... remove the csiphy clocks commit and let's merge the bus vectors commit ONLY.

alviteri:
I can see csi1phy_clk rate before setting its rate. Set the rate first!

kholk:
Reopen the PR with the other commit only, I need to verify some things in drivers

alviteri:
updated

PR #1185, opened 2017-02-26
[1.3.3]  Convert kernel drivers to KUID/KGID

kholk:
This makes us able to enable all kernel namespaces support and comply with the strict uidgid type checks.


oshmoun:
build errors it seems
../../../../../../kernel/sony/msm/kernel/cpuset.c: In function 'cpuset_allow_attach':
../../../../../../kernel/sony/msm/kernel/cpuset.c:1432:17: error: request for member 'val' in something not a structure or union
       cred->euid.val != tcred->uid.val &&
                 ^
../../../../../../kernel/sony/msm/kernel/cpuset.c:1432:35: error: request for member 'val' in something not a structure or union
       cred->euid.val != tcred->uid.val &&
                                   ^
../../../../../../kernel/sony/msm/kernel/cpuset.c:1433:17: error: request for member 'val' in something not a structure or union
       cred->euid.val != tcred->suid.val)
                 ^
../../../../../../kernel/sony/msm/kernel/cpuset.c:1433:36: error: request for member 'val' in something not a structure or union
       cred->euid.val != tcred->suid.val)

kholk:
Commits have been tested on Karin and compiles fine there. Doing a mrproper and retrying. Which device are you trying to compile?

OH. Are you trying to compile without the strict uidgid type checks?

oshmoun:
oooh, yeah that might be it, was doing a dirty build
building for kugo
shutting up till after a clean build is done 😅

kholk:
The configuration options that needs to be enabled for this to work are:
CONFIG_NAMESPACES=y
CONFIG_UTS_NS=y
CONFIG_USER_NS=y
CONFIG_PID_NS=y
CONFIG_NET_NS=y
CONFIG_UIDGID_CONVERTED=y
CONFIG_UIDGID_STRICT_TYPE_CHECKS=y


This doesn't mean that the kernel should break without them.................

bartcubbins:
can confirm, have the same errors

oshmoun:
breaks without configs
with configs -> ~~all good~~ also breaks, that's weird
obviously doing something wrong, not sure what though

kholk:
Rechecked locally, without configs it breaks. With configs it doesn't.

Did some other research and there is no reason to keep the configs disabled when the conversion is actually done, so.. let's ignore the breakages WITHOUT those configs for now.

Your breakages may be because of ANDROID_PARANOID_NETWORK which is not yet migrated to KUID/KGID.

jerpelea:
please fix it berore we can merge it.

kholk:
Pretty much obviously, android's paranoid networking code needs to be fixed for this to be merged.
Also this PR will need to be forcefully merged along with another PR (not yet available) that enables the correct configs for all devices, otherwise breakages will happen.
That would be the same situation we had at the very beginning when we made configs the first time.... anyway, will not be a problem, since the config mods would not be device specific, but global.

oshmoun:
after reverting 2cd1c7c and 4c2f847, and adding configs, kugo now builds and boots fine 👍

kholk:
The PR has been rebased, Adreno commits which ended here erroneously have been removed.

If tests went good, IMHO it can be merged safely.

PR #1194, opened 2017-03-03
defconfig: kanuti: enable MSM_AVTIMER

alviteri:
msm8939 uses avtimer
https://github.com/sonyxperiadev/kernel/blob/aosp/LA.BR.1.3.3_rb2.14/arch/arm/boot/dts/qcom/msm8939-common.dtsi#L1692


kholk:
Makes sense

alviteri:
also camera needs it

drivers/media/platform/msm/camera_v2/isp/msm_isp_axi_util.c:#ifdef CONFIG_MSM_AVTIMER
drivers/media/platform/msm/camera_v2/isp/msm_isp_util.c:#ifdef CONFIG_MSM_AVTIMER

kholk:
Indeed it does. Maybe we should check if blobs support it?
They should..... loire should have it, so that pr should probably be merged

PR #1195, opened 2017-03-03
[1.3.3] Fix bluetooth pairing issues

humberos:
BT pariring issues:

I lost BT pairing from my watch but my phone is still recognizing my watch as a paired device. No communication, no data transferring,... nothing. Same issue happens in my car's BT system (Blue&Me).

If I am move so far away from my phone then I will lost the BT pairing but it can't reconnect anymore even if I have it in my hands again. The issue happens on my phone and my car.

Using IOCTL I need to restart bluetooth from my phone to fix the pairing process.

This patch set requires libbt change:
https://github.com/sonyxperiadev/device-sony-loire/pull/97

Tested on suzu!


humberos:
ping @Diewi

humberos:
IMO I think it is worth to patch libbt library to solve problems between platforms and preserve kernel driver.

https://gerrit.omnirom.org/#/q/topic:sony-brcm-bt-fm-7.1

jerpelea:
a proper patch should not patch userspace

kholk:
I have seen the userspace patches and I agree with Alin.

The patching that you're doing on the userspace libbt can be done in the kernel driver: handles can be added and/or ignore ways can be written.

kholk:
I will be even more specific.
On the file bt_vendor_brcm, you are ifdeffing out the whole vendor case BT_VND_OP_POWER_CTRL. This event will either originate in kernel or in some part of userspace, but all the vendor specific operations are then anyway handled in the kernel, so that's.... Easy. 

On hardware.c you are modifying hw_config_start, which is something that should originate in kernel. You are also defeating NULL checks in your userspace patch, which would cause a whole system crash if anything happens........

PR #1200, opened 2017-03-05
1.3.3 offload

alviteri:


zachariasmaladroit:
https://github.com/zachariasmaladroit/android_kernel_sony_msm8994_kitakami_r2/commit/7faeb1ef9d4a0acf74dec5bd18e87775aae86e5a

how many different commits to solve this are there ?

http://stackoverflow.com/questions/30896489/why-is-u8-u16-u32-u64-used-instead-of-unsigned-int-in-kernel-programming

alviteri:
In this kernel the struct is u32

PR #1201, opened 2017-03-05
drivers: bluesleep: Fix UART clk turning on after resume.

Diewi:
In I50eb35b6af93f6bebb2122b3f80061f04b120998, the UART clock turn on
code in the resume function was modified to use the hsuart_power
function to avoid code duplication. The latter method checks the
BT_SUSPEND flag and does nothing if it is not set. Due to the fact that
the flag is cleared only after the invocation of hsuart_power, the clock
is not turned on as it was previously. Hence, move the code to clear the
flag before the invocation of the hsuart_power function to restore the
orignal behavior.

Change-Id: Icaf781433bd3641c52a2a2f98007ea9ba721ba38

### DIFF
--- a/drivers/bluetooth/bluesleep.c
+++ b/drivers/bluetooth/bluesleep.c
@@ -702,13 +702,14 @@ static int bluesleep_resume(struct platform_device *pdev)
 	if (test_bit(BT_SUSPEND, &flags)) {
 		if (debug_mask & DEBUG_SUSPEND)
 			pr_info("bluesleep resuming...\n");
+		clear_bit(BT_SUSPEND, &flags);
+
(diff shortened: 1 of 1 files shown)


Diewi:
Should fix the BT connectivity issues... BT devices now properly wake up the phone.

oshmoun:
Ping @humberos 😀

kholk:
I don't know whether to laugh or cry when I see code like this being fixed.

Good job!!! :D

humberos:
@Diewi 
Thanks again to your effort but sadly I want to say that it doesn't fix the pairing issue that I have related before. Tested on loire suzu.

PR #1206, opened 2017-03-06
phy-msm-usb: also set usb switch port for Suzu when OTG state is manu…

tomty89:
…ally set

As msm_otg_select_usb_switch() is called in msm_id_status_w() when OTG state is set according to the actual ground state of the ID pin, it should be also called when the state is manually set through sysfs.

### DIFF
--- a/drivers/usb/phy/phy-msm-usb.c
+++ b/drivers/usb/phy/phy-msm-usb.c
@@ -5163,6 +5163,9 @@ static int otg_power_set_property_usb(struct power_supply *psy,
 	switch (psp) {
 	case POWER_SUPPLY_PROP_USB_OTG:
 		motg->id_state = val->intval ? USB_ID_GROUND : USB_ID_FLOAT;
+#ifdef CONFIG_MACH_SONY_SUZU
+		msm_otg_select_usb_switch(motg);
+#endif
 		queue_delayed_work(motg->otg_wq, &motg->id_status_work, 0);
 		break;
 	/* PMIC notification for DP DM state */


kholk:
If it's tested, then ok. 
We should probably even disable automatic otg discovering based on pin5 grounding and provide a stock-like solution through our Extended Settings app.

tomty89:
@kholk Nope, it's not tested. The truth is, I am not even sure how exactly it should be tested, even if I were capable of doing it. (I have been using only unmodified stock ROM, and I don't have computer by my side these days.)

`msm_otg_select_usb_switch` is introduced in commit 867d794 (by you) for Suzu specifically, but there was no explanation on what it does exactly, like whether OTG mode would not work without it at all, or if it is for performance boost or something. (maybe in other words, what the two "ports" are exactly and the difference between them)

I can't tell why specific hacks for Suzu are needed either, as in, why they are not needed for other models, or why we can't have a general approach that works for both Suzu and other models instead. I also find it kind of ugly to had the hacks introduced in the same commit where codes for smbcharger extension are introduced.

But since `msm_otg_select_usb_switch` checks `id_state`, so I think it is needed for the `set_property` case where `id_state` is set explicitly, and have it called at a similar point as it is in `msm_id_status_w`.

By the way what is the stock solution you are referring to? On my Xperia X, the stock ROM is relying on the pin for OTG detection.

tomty89:
I just noticed that `id_state` is set in some other places in the same file as well. I wonder if `msm_otg_select_usb_switch` should called after them too.

kholk:
Sorry about the ugly commit. Usually when I port devices I tend to keep things as clean as possible, but it can happen to not have too much time and just rush to get it working. 

The suzu specific parameters are needed to tune the controller parameters, such as drive strength, latencies etc, to ensure compatibility with all USB devices (even the bad Chinese things...... We don't live in a perfect world). 

Regarding your unification question, you have a damn good point here but.... Unfortunately there's no way. I mean, Suzu uses a ChipIdea USB controller (like 8226, 8939), while other devices do have a Designware DWC3 controller (8974, 8994, 8996 and all high end SoCs). 
If you check the DWC3 driver, you'll see that there all is unified and you have (more or less) the same tuning ways (parameters being different for each SoMC platform, DT). 


About the stock solution, I'm sorry, I confused Suzu with the big brothers (XP, XZ). 

The commit needs to be tested, anyway. Someone will do.

tomty89:
I should probably force push a new commit having `msm_otg_select_usb_switch` called wherever `id_state` is set/changed. Unfortunately I am not capable of doing it now. I'll close this for now and create a new PR  later.

PR #1212, opened 2017-03-10
[1.3.3] Import basic exFAT support

pablomh:
From https://github.com/dorimanx/exfat-nofuse.

The code was originally released by Samsung and hosted at http://opensource.samsung.com, but they decided to remove it. Thankfully some people forked it from there and this is where this code comes from.

Current caveats:

- fsck and mkfs support are missing so vold needs to be modified to cope with that at startup time.


tomty89:
Well https://github.com/relan/exfat provides userspace utilities for exFAT (in addition to a FUSE driver module). I doubt this pull request would be merged though, because of the patent issues and so, which is probably the reason Samsung "unhosted" it.

FWIW, the exFAT support in stock ROMs from Sony is provided by Tuxera under a commercial license, while the driver is closed-source, AFAIK.

pablomh:
We don't want to use FUSE being these devices SOCs.

Regarding the patents issue, I'm just redistributing code that was released under the GPL officially by Samsung, so I'll let the developers decide about the issue.

tomty89:
Well, I was suggesting, if the nofuse driver is going to be merged, perhaps the userspace utilities (which does not rely on FUSE) from relan can be adopted as well, to address the caveat you mentioned.

pablomh:
Hi Tom,

I didn't know I could use the userland utilities with the kernel version. I'll try it and will revert the "hack" commit if they work.

pablomh:
In my initial testing of this approach it seems that after running the FUSE fsck, it reports a clean FS and a subsequent kernel mount reports I/O error, whereas a FUSE mount succeeds... so I need to dig deeper and run further testing.

I will keep this on hold as my testing time is limited ("daily driver device running stock").

pablomh:
Please disregard the last two merges as I merged them into the wrong tree...

tomty89:
You can do something like these to undo it:

```
git reset --hard 4513a79 
git push --force
```

Btw, your last commit (4513a79) probably won't work. You made it check `p_fs->vol_type` before it is set by actual mount function (`fat16_mount`/`fat32_mount`/`exfat_mount`), which would be called later in `ffsMountVol`.

See:
https://github.com/pablomh/kernel/blob/4513a79ecfd5d55d58be442fed1b4115b6e8cf33/fs/exfat/exfat_core.c#L4560
https://github.com/pablomh/kernel/blob/4513a79ecfd5d55d58be442fed1b4115b6e8cf33/fs/exfat/exfat_core.c#L4611
https://github.com/pablomh/kernel/blob/4513a79ecfd5d55d58be442fed1b4115b6e8cf33/fs/exfat/exfat_core.c#L4683

You probably want to replace:
```
	if (i < 53) {
		if (GET16(p_pbr->bpb+11)) /* num_fat_sectors */
			ret = fat16_mount(sb, p_pbr);
		else
			ret = fat32_mount(sb, p_pbr);
	} else {
		ret = exfat_mount(sb, p_pbr);
	}
```
with:
```
	if (i < 53) {
		printk("[EXFAT] driver only supports exFAT formatted drives\n");
		return FFS_FORMATERR;
	} else {
		ret = exfat_mount(sb, p_pbr);
	}
```
in `ffsMountVol`.

FWIW, the values of `FAT12`, `FAT16`, `FAT32` and `EXFAT` that are defined in _exfat_api.h_ are practically arbitrary. IMHO it's silly to define them with MBR partition type codes. A simple `enum` would serve better.

tomty89:
Hi @pablomh , so I see that you changed your last commit as per my last comment. Unfortunately I just notice that I still missed something. Before `return FFS_FORMATERR`, you need to call `brelse(tmp_bh)` and `bdev_close(sb)`, which releases the resource used for the boot sector read and closes the block device respectively. You can see that when it does an early return for invalid PBR Signature just above, it has the same procedures done as well.

So instead of:
```
		printk("[EXFAT] driver only supports exFAT formatted drives\n");
		return FFS_FORMATERR;
```
You actually need:
```
		printk("[EXFAT] driver only supports exFAT formatted drives\n");
		brelse(tmp_bh);
		bdev_close(sb);
		return FFS_FORMATERR;
```

pablomh:
Hi Tom:

You're completely right and I've fixed the code as you suggested. I took it into account when I first looked into that part of code during my first patch attempt, but forgot about it later :(

Issue #1183, opened 2017-02-26
hvdcp_max_current not in effect?


tomty89:
In b413135,  support for HVDCP charger has been added to the driver `phy_msm_usb`. Max current drawn from HVDCP charger has been set to 1800mA in the commit as well. I can confirmed that it has been applied to the kernel I am using:

```
u0_a210@F5122:/ $ cat /sys/module/phy_msm_usb/parameters/hvdcp_max_current
1800
```

However, when I check the following uevent file while I am charging my Xperia X, I see that the max current is not 1800mA but 1500mA:

```
u0_a210@F5122:/ $ cat /sys/class/power_supply/usb/uevent
POWER_SUPPLY_NAME=usb
POWER_SUPPLY_HEALTH=Good
POWER_SUPPLY_PRESENT=1
POWER_SUPPLY_ONLINE=1
POWER_SUPPLY_VOLTAGE_MAX=0
POWER_SUPPLY_CURRENT_MAX=1500000
POWER_SUPPLY_INPUT_CURRENT_MAX=0
POWER_SUPPLY_SCOPE=Device
POWER_SUPPLY_TYPE=USB_HVDCP
POWER_SUPPLY_VOLTAGE_NOW=9038620
POWER_SUPPLY_DP_DM=1
POWER_SUPPLY_USB_OTG=0
POWER_SUPPLY_SUB_TYPE=0
POWER_SUPPLY_CHARGER_TYPE=USB_HVDCP_CHARGER
```

Does that mean some necessary changes are missing for proper support for HVDCP?

Btw, when I check the history of the `phy_msm_usb` driver, a recent change (867d794) is that some hacks have been added for Suzu (which is Xperia X, the model I have here, IIRC). So what so special about Suzu that makes it need hacks?

tomty89:
I can see that `hvdcp_max_current` is only used in an `msm_otg_notify_charger` called in `otg_power_set_property_usb`. (Note: compare this with where/how `dcp_max_current` is used)

The thing is, it appears to me that, `otg_power_set_property_usb` is a method that is only used when we write to corresponding sysfs file (`/sys/class/power_supply/usb/type` in this case), but not when the driver detects the actual `psy->type`. (In fact, I don't see it does that at all other than in `otg_power_get_property_usb`, which is probably used only for sysfs files as well.)

[...]

PR #1213, opened 2017-03-10
Fix for Suzu OTG

tomty89:
NOT TESTED. Please help test it if you can. Refer to #1206 or get in touch with @kholk for more details.

### DIFF
--- a/drivers/usb/phy/phy-msm-usb.c
+++ b/drivers/usb/phy/phy-msm-usb.c
@@ -3468,6 +3468,9 @@ static void msm_otg_init_sm(struct msm_otg *motg)
 	}
 	motg->id_state = (test_bit(ID, &motg->inputs)) ? USB_ID_FLOAT :
 							USB_ID_GROUND;
+#ifdef CONFIG_MACH_SONY_SUZU
+	msm_otg_select_usb_switch(motg);
+#endif
 }
 
 static void msm_otg_wait_for_ext_chg_done(struct msm_otg *motg)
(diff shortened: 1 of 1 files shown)


humberos:
Just wondering if it is not a platform requirement. I mean about kugo,
ping @oshmoun

oshmoun:
sure, could test that
just, what is the targeted difference, i.e what is supposed to change in the behaviour? afaik otg works well 😅

tomty89:
@oshmoun tbh no idea. @kholk briefly explained about it in #1183 and #1206 but i am still not sure what exactly it does. i just create this as a supplement/fix for 867d794, as i've told in the commit message

oshmoun:
sorry for the wait
the function is only defined for suzu
https://github.com/sonyxperiadev/kernel/blob/083cfc020ea502f033cfa3d9819a6efd448d92ec/drivers/usb/phy/phy-msm-usb.c#L209
since that was always the case from before, I'm assuming the fix (whatever it's supposed to do) is only targeting suzu
kugo and suzu do differ a bit on the usb hw side, as kugo has usb type c similar to XZ

tomty89:
Yeah I mentioned in the commit message that "In commit 867d794, msm_otg_select_usb_switch was introduced for Suzu to..."

The problem in the original commit was, that function apparently changes the port selection (whatever it is) according to `id_state` (i.e. OTG state), but was only called at one point where `id_state` can get changed, while that can happen at some other points.

So what I am doing here is merely filling back in what was missed programatically/logically. As for whether the function itself was ever proper, or ever needed for Suzu, I assume @kholk should know best...

jerpelea:
your last patch affects an unused platform !

tomty89:
@jerpelea sorry are you talking to me? what last patch? are you referring to 1d80df9 which got mis-auto-tagged/linked to this pull request?

jerpelea:
1d80df9 is for rockchip devices

tomty89:
@jerpelea the commit was never intended to be relevant, it "referred" to this pull request just because this line in its commit message:

```
[ 7226.716722] CPU: 1 PID: 1708 Comm: bash Not tainted 4.12.0-rc6+ #1213
```

which is a line of log that happens to consists of "#1213"

jerpelea:
ok

PR #1223, opened 2017-03-23
March CVE

scafroglia93:
- Compile and booted on Z3
- I don't have arm64 device plz tested and keep me update


danielgusvt:
Works fine on tulip, only stock kernel tested

danielgusvt (approved):

PR #1225, opened 2017-03-24
[1.3.3] [TESTMEFIRST] staging: LMK: Evaluate regular swap space when calculating memory

kholk:
Android lowmemorykiller was evaluating only swapcache pages, but
not "real" linux swap space: this was making it to usually kill
processes anyway when the kernel swappiness parameter was high.

Include swap space(s) evaluation in the vmpressure and shrink
algorithms to allow swap space to be used as a last resort.


oshmoun:
tested on kugo, seems to work ok
haven't compared if lmk triggers less often now, just checked if apps are being kept in memory by rapidly switching through many of them, looks acceptable.

oshmoun:
tested for a day, device didn't develop sentience
👍

kholk:
If noone has problems with this, the pr can be merged

PR #1226, opened 2017-03-25
[UM.5.5.r1] Tone: Bluetooth fixes

humberos:
Tone family does not have BCM BT/FM so remove it.
Use default bluesleep and serial drivers to avoid kernel instability.


humberos:
Related:
https://github.com/sonyxperiadev/device-sony-common/pull/243
https://github.com/sonyxperiadev/device-sony-tone/pull/24

PR #1227, opened 2017-03-26
[UM.5.5.r1] msm8996 DT fixes

humberos:
Those values was extracted from 39.2.A.0.442.


kholk:
Damn. I did the same commits and was about to pr. Lol

PR #1228, opened 2017-03-26
[UM.5.5.r1] GPU reverts

humberos:
GPU performs better with those reverts although we still can't see its scaling yet.

### DIFF
--- a/drivers/gpu/msm/a5xx_reg.h
+++ b/drivers/gpu/msm/a5xx_reg.h
@@ -640,7 +640,6 @@
 
 /* UCHE registers */
 #define A5XX_UCHE_ADDR_MODE_CNTL            0xE80
-#define A5XX_UCHE_MODE_CNTL                 0xE81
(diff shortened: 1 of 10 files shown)


kholk:
Hold these for a little....... 

Check the dmesg, I have a message about an unknown header in gpmu firmware, probably that's what produces the issue (will continue researching today).

humberos:
yeah
[    6.651173] kgsl kgsl-3d0: |_load_regfile| request firmware failed -11, a530v3_seq.fw2
This file is present at firmware folder.

apart from this gpu issue I think that we have something broken on filesystem mouting process.

humberos:
or filesystem driver (sdcardfs maybe ?)

kholk:
No, I wasn't referring to this error. But that's pretty much interesting! 

We should try to update adreno firmwares.

humberos:
I did it from latest kagura stock (.442) files....
unfortunately same result :(

humberos:
https://github.com/sonyxperiadev/kernel/pull/1232

PR #1242, opened 2017-03-31
1.3.3 nvidia power efficiency

RyTekk:
Reduce power usage by using relaxed macro.

(2 latest commits were copied from kernel which was downloaded from here (https://developer.nvidia.com/embedded/linux-tegra-r2421), I can't find a git site for that kernel source)

### DIFF
--- a/arch/arm/include/asm/atomic.h
+++ b/arch/arm/include/asm/atomic.h
@@ -28,6 +28,7 @@
  */
 #define atomic_read(v)	(*(volatile int *)&(v)->counter)
 #define atomic_set(v,i)	(((v)->counter) = (i))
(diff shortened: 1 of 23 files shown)


kholk:
Yes I actually already knew that thing.... The questions I've always made to myself is, exactly, the same that I'm doing to you. 

Is it specific to Tegra? As in, will it be power efficient on Tegra or on most of the SoCs out there? 
Do "most of the SoCs" include MSM from 8974 to 8996?

Is it a good gain? 
Is it so good to not care about diverging that much from caf AND mainline?

RyTekk:
Yah, I don't know if it works but at least for me until now, it hasn't broken anything, device (sumire) seems to be cooler.

PR #1243, opened 2017-03-31
[UM.5.5.r1] drivers: bluetooth: tone: Fix bluesleep wakelock

humberos:
Current driver is causing wakelock on tone devices which will not
able to going to sleep properly.

So, this patch uses the same logic developed for ldisc targets (tx timer)
and now tone devices should be able to sleep as expected.

Change-Id: Ia619bd8d83e049cabb082b2a668101bd6d856799

### DIFF
--- a/drivers/bluetooth/bluesleep.c
+++ b/drivers/bluetooth/bluesleep.c
@@ -153,11 +153,9 @@ static unsigned long flags;
 /** Tasklet to respond to change in hostwake line */
 static struct tasklet_struct hostwake_task;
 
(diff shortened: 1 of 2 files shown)


humberos:
please hold on a bit. I need one day more testing this!

humberos:
@jerpelea 
it is the best approach that we have for tone's BT platform for now.
Although BT wakelocks still present in logs but the BT power management is equivalent as others platforms.
I mean now deep sleep state is ok and also preserved all BT functionalities.
So this patch is good to go!

To do: I will do a research about BCM BT/FM.

PR #1248, opened 2017-04-02
amami: Correct fcc-mah in DT

hahnjo:
Commit ae463afda7d5 ("arm: DT: Rhine: Implement DT battery data") ported the
battery data to DT and accidentally used the same value as honami.
Use correct value from C removed in commit 0af12f52b182 ("arm: mach-msm:
Remove C-form batterydata for Shinano and Rhine").


hahnjo:
However, even with this change, the battery duration is disappointing and much worse than on stock rom :-(

hahnjo:
Ping

PR #1253, opened 2017-04-03
[UM.5.5.r1] drivers: power: reset: Fix adb reboot bootloader/recovery

humberos:
This change is based on OnePlus3 kernel which made
a hack to force a soft reset as this preserves
memory contents, including pstore and other
persistent memory.

In this case we are focing the soft reset only if the bootloader or
recovery is sent as cmd.

Reference: https://github.com/LineageOS/android_kernel_oneplus_msm8996/commit/f0979313e42697b3bc2311f0241cfb045c18eed0

Change-Id: Ibf201de9cdceab8434926d9e35e453c526c14056


pablomh:
- Format nit: can you please fix the C++ comment format?

kholk:
I thought we already discussed about that.... 
If you always force warm reboot, you will make the pmic to draw current even when you shut down the board. 

I tried something like this yesterday on Dora and didn't work, but you added one more line... I will test again. 

The warm reboot has to be set in pmic only when a reboot command is set.
Let's say, give me a couple hours and I'll put a more constructive comment on this :p

humberos:
@pablomh 
Thanks for the tip. Patch updated.

@kholk 
Yeah we've discussed about that (thanks btw).
It works on kagura and I think it should work on dora too.
This patch adds a new restart reason enum item ~~and it also changes the raw_writel code from 0x776655AA to 0x77665501~~.

About warm_reset hack, device will not be able to check the reboot cmd after a hard reset.
but you have the point about power consumption (shutdown cmd) that we should be consider on this change.

humberos:
oh, I think we also could change the logical verification for cmd.

kholk:
If diverging from qualcomm would be a good option, I would have already changed the stupid if else if else if else if infinite thing... that's really stupid.

Anyway, still didn't test on Dora but it doesn't matter, it's ok as long as it works on Kagura.
If Dora has bootloader bugs it's not our problem.


Unfortunately, the thing here is not as clean as I would like it to be but.... well, as I said, I don't want to diverge from QC too much for easy porting.

Merge!

PR #1254, opened 2017-04-04
[UM.5.5.r1] driver: staging: Android Low Memory Killer: consider swap free size

humberos:
Use freeswap as otherfile when watermark is grater than
low_watermark. This helps to make use of zram swap.

Extracted from 39.2.A.0.442


kholk:
That's way more hacky compared to the one that I made on 3.10, but it's also more precise in swapspace accounting.

Let's use this version on 3.18.

PR #1255, opened 2017-04-05
[UM.5.5.r1] mm: Introduce SWAP_CONSIDER_CMA_FREE

humberos:
When swap is enable, CMA page is also swapped out.
But CMA page is not used for movable pages.
So in some cases many CMA page remains free.
This help to avoid this situation
by introducing a threshold allowing CMA page reclaim.

Extracted from 39.2.A.0.442


humberos:
still testing.
~~seems that it breaks bootanimation~~ (it is ok - garbage on my environment).

humberos:
ok, one entire day running this patch without regressions.
tested on kagura.

PR #1260, opened 2017-04-17
arm: irq: drop "IRQx no longer affine to CPUn" print to pr_debug()

danielgusvt:
On devices where hotplug is used for power savings or thermal
mitigation, it's not uncommon for hotplug events to be relatively
frequent and their latency to matter.

If a serial console is enabled, the "IRQ no longer affine to CPU"
informational prints can be one of the most significant contributors
to hotplug latency. For example, on msm8994, 10 lines of this format
are normally printed on each hotplug. At 115200/8/N/1 baud, that
printing takes a minimum of 42.5ms.

On production devices where a serial console is not enabled, the time
takes is negligible (a few microseconds at most), but the prints do
still clutter the logs and are worth silencing since they are
typically of no concern.

Signed-off-by: Matt Wagantall <mwagantall@cyngn.com>

@kholk muted the logspam for arm64, but not for arm, so this commit (partially) fixes this logspam https://github.com/sonyxperiadev/kernel/issues/1259

### DIFF
--- a/arch/arm/kernel/irq.c
+++ b/arch/arm/kernel/irq.c
@@ -183,7 +183,7 @@ void migrate_irqs(void)
 		raw_spin_unlock(&desc->lock);
 
 		if (affinity_broken && printk_ratelimit())
-			pr_warning("IRQ%u no longer affine to CPU%u\n", i,
+			pr_debug("IRQ%u no longer affine to CPU%u\n", i,
 				smp_processor_id());
 	}
 


kholk:
It's the opposite... This commit is in arch/arm, not arm64 :P

Then, while I agree that the logspam is very bad to have, I disagree with this commit, because it doesn't solve the actual problem. You are silencing a kernel warning in a very sensitive piece of code. 
The comment is not the worst thing ever because of course it's true, logspam adds latency to the fine granularity of rqbalance's hotplugging algorithm, but there is a problem that should not be underestimated. 

The message is being spawned whenever we want to migrate an IRQ that has broken cpu affinity: this can lead to random kernel lockups for apparently no reason, which usually happens in case of spurious IRQs, which are a situation to avoid at any cost. 

By keeping the message there we make sure we see that, in case of a random lockup, we see that there is actually some problem. 

If you want to push a fix, then just silencing the warning is not the way to go. You should research about the source of this issue, by checking why the affinity of those interrupts is broken and fixing it properly.

zachariasmaladroit:
@kholk 

yeah, actually seems to be a pretty old & "known" problem:

https://bugs.launchpad.net/linaro-big-little-system/+bug/1097213 Repeatedly hotplugging CPUs crashes the kernel


something like the following

http://lists.infradead.org/pipermail/linux-arm-kernel/2015-September/369701.html [RFC PATCH v2 3/3] arm/arm64: fix a migrating irq bug when hotplug cpu

?

**edit:**

that seems to be the "approved" approach:

http://lists.infradead.org/pipermail/linux-arm-kernel/2014-June/266923.html

newest "incarnation" (arm64 only) of that patch:

https://patchwork.kernel.org/patch/7254771/ [RFC,v6,2/3] arm64: fix a migrating irq bug when hotplug cpu

and the final (?) upstream variant for arm (32bit):

http://lists.infradead.org/pipermail/linux-arm-kernel/2017-January/480557.html arm: Use generic irq_migrate_all_off_this_cpu() for CPU hotplug

zachariasmaladroit:
Looks like there are some more hotplug & IRQ related issues - outlined in:

http://www.spinics.net/lists/arm-kernel/msg525054.html Re: [PATCH] arm/perf: Fix pmu percpu irq handling at hotplug.

kholk:
Yes, this is a very old problem. To fix it, you must find all the commits and dependencies, hopefully less than 50 commits.

rancidfrog:
The main two direct commits are
-  https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable.git/commit/include/linux/irq.h?id=f1e0bb0ad473a32d1b7e6d285ae9f7e47710bb5e | https://github.com/torvalds/linux/commit/f1e0bb0ad473a32d1b7e6d285ae9f7e47710bb5e
- https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable.git/commit/?id=217d453d473c5ddfd140a06bf9d8575218551020 | https://github.com/torvalds/linux/commit/217d453d473c5ddfd140a06bf9d8575218551020

How to find dependencies?
Is there a strategic way?

zachariasmaladroit:
https://lkml.org/lkml/2015/9/24/136 [RFC PATCH v6 0/3] arm/arm64: fix a migrating irq bug when hotplug cpu

is a start, 

~~mind you that GENERIC_PENDING_IRQ was renamed to GENERIC_IRQ_MIGRATION~~

~~[https://patchwork.kernel.org/patch/7130631/]~~

~~so for us it's GENERIC_PENDING_IRQ (at least 3.10 kernels)~~

You can take a look at newer kernels and their relevant files (e.g. those outlined at https://lkml.org/lkml/2015/9/24/136)

e.g. for arch/arm64/kernel/irq.c

via github (the interface is quite intuitive and helps accelerating research quite nicely):

https://github.com/torvalds/linux/commits/master/arch/arm64/kernel/irq.c

that lists all commits/changes for that specific file

locally you can checkout a newer kernel and take a look line-by-line

via git blame, e.g.

git blame arch/arm64/kernel/irq.c

zachariasmaladroit:
Not sure why there are both options available though: https://github.com/torvalds/linux/commit/f1e0bb0ad473a32d1b7e6d285ae9f7e47710bb5e#diff-3ed369de7a86e09204577cacea5c5e76R29

zachariasmaladroit:
I opened up a new PR at https://github.com/sonyxperiadev/kernel/pull/1264

danielgusvt:
Okay:))

PR #1272, opened 2017-04-27
[1.3.3] arm: DT: MSM8939-mdss: Disable MDSS1 intf by default

bartcubbins:
MSM8916 MSM8936 MSM8939 devices usually have only one MDSS interface connected to a display.
Disable the other interface to get faster mdss probing, hence fast boot times, and less errors at boot time.


pablomh:
Do you have any idea why it's not disabled in copyleft?

kholk:
Good idea, wrong implementation. 
Goes to tulip DT.

jerpelea:
please add it to tulip dt

PR #1276, opened 2017-05-05
[LA.BR.1.3.3_rb2.14] Merge BCMHD fixes from angler 3.10 - nougat-mr2

pablomh:
**Please test as I don't have the hardware to do it.**


rcstar6696:
Any reason why not enable the config on all devices like msm 8974 ?

pablomh:
Finer grained control.

I'll leave it up to @jerpelea to decide.

jerpelea:
after the patch will be fixed and tested we can enable it on all devices that use BCMDHD

pablomh:
This one builds in kitakami and loire at least.

PR #1285, opened 2017-05-13
[5.5.r1] Binder updates from android-msm-marlin-3.18-o-preview-1

kholk:
This is the new Android Binder for Android O Preview.
Also works on Android N 7.1.2 apparently flawlessly.

The new binder kernel driver mainly features more performance and some security updates, though, some of the performance features can be used only by new binder userspace in Android O.


kholk:
PR cleaned up

PR #1286, opened 2017-05-17
[UM.5.5.r1] Decrease BT/FM TX timeouts to improve deep sleep ratio.

humberos:
Decrease the tx timeouts in the bluesleep driver and the
broadcom BT/FM driver to 1 second such that the device enters
deep sleep earlier. This can improve the deep sleep ratio of
devices with a connected BT device due to the many keep alive
packets sent around. For each of thes packets the bluesleep
driver would block suspend for 5-10s otherwise (where keep
alive packets are typically sent every 12s).

Change-Id: I5661f25643acbeaf1c326af45ff20bb9f1a2bd2f

### DIFF
--- a/drivers/bluetooth/bluesleep.c
+++ b/drivers/bluetooth/bluesleep.c
@@ -120,8 +120,8 @@ DECLARE_DELAYED_WORK(sleep_workqueue, bluesleep_sleep_work);
 #define bluesleep_rx_idle()     schedule_delayed_work(&sleep_workqueue, 0)
 #define bluesleep_tx_idle()     schedule_delayed_work(&sleep_workqueue, 0)
 
-/* 5 second timeout */
-#define TX_TIMER_INTERVAL  5
+/* 1 second timeout */
+#define TX_TIMER_INTERVAL 1
 
 /* state variable names and bit positions */
 #define BT_PROTO	0x01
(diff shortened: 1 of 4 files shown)


zachariasmaladroit:
👍 

@humberos did you notice significant savings ?

according to your statements it must surely be noticable

Thanks

Diewi:
The savings only apply for short transmission requests. Also, there are other suspend blockers in the kernel, which may just not be visible in statssince they were hidden. There is a saving, but do not expect a wonder ;) However, an increase of deep sleep ratescould be noticed in my tests.

kholk:
Makes a lot of sense. There's only one thing I'd be worried about. 

Keepalive packets must not be missed, this is for BLE/BT4.0. Did anyone try this with a wearable?

Diewi:
Yep, me, using an Android Wear smartwatch with this patch for some time and did not notice any problems. However, a confirmation would be good.

kholk:
I can test it with my MTK watch with a custom system running android wear.... The connection is tricky even with stock ROM drivers but stable. Probably mine would be the best test we can do but, unfortunately, I can't do that anytime soon. 

Probably next week, but I'll try to do it ASAP.

kholk:
In any case, if you have no problems with your AW watch, then it's probably fine!

humberos:
I am using Pebble watch paired with
No problems here

kholk:
Nah, then just merge it, my test would be superfluous.

PR #1309, opened 2017-06-16
[1.3.3] arm/dts: kitakami/sumire: Higher charging limitation

RyTekk:
We can safely increase maximum charging current with QC2.0 to 1800mA without hurting battery and device. Battery get fully charged faster by 15-20 mins, battery temperature always stays below 45 Celcius (tested for months). I also edit other values to ensure that the current doesn't suddenly increase over 1800mA since I got that issue something.


jerpelea:
do not increase charging current because it can increase thermals and damage battery

PR #1325, opened 2017-06-24
[LA.UM.5.5] Increase the CONFIG_HZ value

humberos:
Yes, we already have talked about this before.
Just send it as a reference as I am using it in my builds.

It offers a noticeable performance and power reduction due to decreased RCU overhead.

Official reference:
https://source.android.com/devices/tech/debug/jank_jitter


jerpelea:
we do not accept new features on this kernel since we are now developing 4.4 
please test your feature on the new kernel

PR #1326, opened 2017-06-24
[LA.UM.5.5] kgsl: convert some workqueues to use kthreads

humberos:
adreno_dispatch_work and _kgsl_event_worker are both low-latency
low-runtime functions that are in the critical path of GPU
rendering. Moving them out of workqueues and into a dedicated FIFO
kthread avoids significant jitter.

bug 30342017

Change-Id: I5feb4e829064d422b4b9af2acb449afd1f981899


jerpelea:
we do not accept new features on this kernel since we are now developing 4.4 
please test your feature on the new kernel

PR #1344, opened 2017-07-03
arm: DT: Tone: Fix ramoops configuration

alviteri:
The configuration was entirely wrong. Fix it by using the correct
sizes and by specifying the properties as they should be.


kholk:
I requested the PR. Mergeme :D

PR #1345, opened 2017-07-04
[LA.UM.5.7_r1] ADSPRPC: add support for tone devices by re-introducing SMD calls

oshmoun:
These two reverts allow adsprpcd to function properly on tone.
The revert that brings back SMD calls is needed due to hw.
The revert to remove the DMA coherency flag is needed for unclear reasons (ping @kholk).

### DIFF
--- a/arch/arm/boot/dts/qcom/msm8996.dtsi
+++ b/arch/arm/boot/dts/qcom/msm8996.dtsi
@@ -3612,52 +3612,43 @@
 			compatible = "qcom,msm-fastrpc-compute-cb";
 			label = "adsprpc-smd";
 			iommus = <&lpass_q6_smmu 8>;
-			dma-coherent;
 		};
 		qcom,msm_fastrpc_compute_cb2 {
 			compatible = "qcom,msm-fastrpc-compute-cb";
 			label = "adsprpc-smd";
 			iommus = <&lpass_q6_smmu 9>;
-			dma-coherent;
 		};
 		qcom,msm_fastrpc_compute_cb3 {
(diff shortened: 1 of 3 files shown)


kholk:
The whole thing looks good, apart one small-big-thing: you break everything like that :D
The revert on adsprpc is introducing a property to activate the glink communication, which will make msm8998 and sdm660/630 (which we are currently not using, ok) to use smd, which is obviously wrong. 

I'm fine with it, as long as you do another commit that inverts the behavior of this: the glink communication has to be default and the smd one has to be activated with a property that we will put in our msm8996/8956/8939 DTs. 


About the dma-coherent attribute, this was done for two things:
1. Performance
2. Default behavior on 3.18 appeared to be that! 

In any case, fix what I said, THEN the pr can be merged just fine!
Good job!

oshmoun:
the property is already there for ~~msm8996~~ msm8998
https://github.com/sonyxperiadev/kernel/blob/aosp/LA.UM.5.7.r1/arch/arm/boot/dts/qcom/msm8998.dtsi#L1233
but you're right, legacy behaviour shouldn't be the default!
Will introduce a new property (qcom,fastrpc-smd) and adjust the driver probe accordingly

kholk:
Yes, thanks.
While you are right on the fact that it could also have been done like that, we shall never forget about portability... That's why legacy behavior shall not be default ;)

humberos:
Nice work!
Kagura is ok!

kholk:
It's ok now. MERGEME

PR #1348, opened 2017-07-14
[LA.UM.5.7.r1] drivers: input: touchscreen: Update clearpad driver from 45.0.B.2.95

humberos:
Tested on kagura.
It works fine.

it needs to be tested on:
*dora -> untested
*keyaki -> ok
*kagura -> ok

*maple -> untested

*kugo -> untested
*suzu -> untested

So, ping
@oshmoun 
@alviteri 
@kholk 
@erikcas


oshmoun:
tested on keyaki, no issues found
👍

kholk:
The regoffset thing has to be removed (which is what I did on our clearpad driver). It is useless, dirty and stupid!

oshmoun:
dt2w works fine here now!
Got panel 6, which should be hybrid according to dt
https://github.com/sonyxperiadev/kernel/blob/aosp/LA.UM.5.7.r1/arch/arm/boot/dts/qcom/dsi-panel-somc-kagura.dtsi#L24

Yennhi-1986 (approved):

PR #1349, opened 2017-07-15
[LA.UM.5.7.r1.] Update bcmdhd driver

humberos:
Update from 41.2.A.2.223 then 41.2.A.7.35


oshmoun:
works on keyaki, no issues so far

PR #1353, opened 2017-07-16
[LA.UM.5.7.r1] 4.4 tuning application UI threads

humberos:
It offers a noticeable performance and power reduction due to decreased RCU overhead.

Official reference:
https://source.android.com/devices/tech/debug/jank_jitter


humberos:
ping @kholk 
That's we have talked about some days ago.

kholk:
Please check all drivers used in SoMC platforms to not use raw jiffies, then we can consider to really increase this value.

humberos:
oh I am sorry! Now I got your point!
So it will trigger a lot of changes in drivers which is not a good idea since we are tracking qcom and sony. Both aren't using new HZ value while only one specific aosp target is.
Thank you!

kholk:
You didn't get the point...
....we can use 1000, but you have to fix all the drivers and, once you do, we will have the patchset for it and that will be portable.

Read it better: this time I am NOT DISAGREEING on the change, because 4k on maple sorta needs it (it's complicated, so I will not explain the reasons here and now).
I just need you to get all the drivers fixed prior giving my merge word on this.

kholk:
The conversion, at least on mdss looks COMPLETELY wrong!! On one handler you are even settings msecs_to_jiffies(2000) and I don't think that the intended value is... 2 seconds?! 

And then you are screwing up all the latencies by using a base value of 1000Hz which is completely wrong (as you should adhere to the previous hz value), but even if you used the correct one, you still didn't calculate the right amount of time for those. 

You should basically do a conversion from jiffies to milliseconds at our usual HZ value (which is 100 right now) and then make _any_ HZ setting (300, 500, 1000 or whatever) to adhere to that timing.

kholk:
...just checked also power, OK for it as well.

Go on :D

kholk:
=====NOTE=====
This PR is dependant on
https://github.com/sonyxperiadev/kernel/pull/1358

And shall NOT be merged prior #1358 and prior careful testing.

kholk:
Tested for 2 weeks on SoMC Maple without side issues.

kholk:
Do not merge prior tests on Tone platform.

humberos:
yeah.... I think may loire is the main target here.
tested on kagura which is ok from my side but we still need dora and keyaki confirmations.

alviteri:
Dora tested .... as I can see there is not battery life impact .... kugo on 4.4 I have tested but can't tell about battery life since 8952 is WIP there

alviteri:
@humberos rebase your kernel branch there is one more not needed commit

kholk:
Humberto, please remove the loire 1000hz commit, that is to be tested later when the camera starts because I want to be paranoid on this. 

For me it's SAFE TO MERGE for Tone and Yoshino platforms.

humberos:
totally agreed!
so, patchset was updated
now we have only tone and yoshino as targets.
Thanks!

PR #1356, opened 2017-07-18
[5.7.r1] [TESTMEFIRST] Enable 1080p@120Hz on ALL Tone devices

kholk:
This converts the DT panel configuration to the timings way. Moreover, it adds one more timing to the list, configuring the panel for 120Hz/120FPS.

Panel configuration will anyway assume that we are on 60Hz by default and the 120Hz configuration has to be activated from ExtendedSettings.

### DIFF
--- a/arch/arm/boot/dts/qcom/dsi-panel-somc-synaptics-jdi-1080p-cmd.dtsi
+++ b/arch/arm/boot/dts/qcom/dsi-panel-somc-synaptics-jdi-1080p-cmd.dtsi
@@ -23,18 +23,9 @@
 		qcom,mdss-dsi-panel-controller = <&mdss_dsi0>;
 		qcom,mdss-dsi-panel-destination = "display_1";
 		qcom,mdss-dsi-panel-type = "dsi_cmd_mode";
-		qcom,mdss-dsi-panel-width = <1080>;
-		qcom,mdss-dsi-panel-height = <1920>;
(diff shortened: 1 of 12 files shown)


arch/arm/boot/dts/qcom/dsi-panel-somc-synaptics-jdi-1080p-cmd.dtsi
```diff
@@ -376,5 +342,90 @@
...
+						29 01 00 00 00 00 1B F0 E3 07 73 DF 00 00 00 00 00 00 00 00 E0 E3 00 00 03 14 00 00 20 00 00 00 00 A7
+						39 01 00 00 00 00 02 35 00
+						39 01 00 00 00 00 02 36 00
+						39 01 00 00 00 00 02 3A 77
+						39 01 00 00 00 00 05 2A 00 00 04 37
+						39 01 00 00 00 00 05 2B 00 00 07 7F
+						39 01 00 00 00 00 03 44 00 00
+						39 01 00 00 78 00 01 11];
+				qcom,mdss-dsi-post-panel-on-command = [
+						39 01 00 00 00 00 01 29];
```
oshmoun, line 379: Syntax error!

kholk: Whoops :p

PR #1357, opened 2017-07-19
[5.7.r1] Add 120Hz configurations for Tone, Yoshino devices

kholk:


oshmoun:
confirmed working on keyaki panel 6

humberos:
There is a problem with video playback at 120Hz (kagura as target tested)
This can be reproduced by using Youtube.
Ping @oshmoun

oshmoun:
can confirm, there is major frame skipping happening!
Can't seem to find errors in logcat nor dmesg

kholk:
Oki let's check!
Does it happen only with video playback?

Vidc may be crapping out because it's not supposed to run at that rate... Though all videos you see will be at a max of 60fps.
In any case, I call it strange. 
Maple has zero problems on 120, but well.. That's a different soc and display

PR #1358, opened 2017-07-20
[LA.UM.5.7.r1] 4.4 HZ adaptation

humberos:
Make the various timeout values HZ agnostic by using the proper macros and values instead.


kholk:
This PR should be ready, but has to be carefully tested because of huge changes in timers.

Testing has to be done on devices known to the developer and evaluation has to be done on:
1. Possible unstabilities produced by eventually different timings
2. Anomalous battery consumption
3. System responsivity changes.


The code in this PR wants to fix those 3 problems on configurations with a HZ value different from the standard of 100, although this has to be currently tested on HZ=100 because it shall not break anything for this standard, default value.
After verifying that there's no side effect on HZ=100, this PR has to be merged and THEN PR #1353 shall be carefully tested.
The validity of the code in this PR has been verified by me.

oshmoun:
all looks ok so far, will keep running with this pr and #1353 for a couple days

kholk:
Same here on Maple, standard 100Hz.

kholk:
After at least 16 days of running this with no stability nor battery problems I can say it's safe to merge.

MERGEME

PR #1366, opened 2017-07-27
[LA.UM.5.7.r1] loire: Enable REGULATOR_ONSEMI_NCP6335D

humberos:
This driver supports the OnSemi NCP6335D switching voltage regulator
(buck convertor). The regulator is controlled using an I2C interface
and supports a programmable voltage range from 0.6V to 1.4V in steps
of 6.25mV.


jerpelea:
this regulator is used by msm8926 and msm8939 and should be removed from loire

PR #1368, opened 2017-07-28
[LA.UM.5.7.r1] loire | tone: Enable qcom,msm-lsuart

humberos:
Select this module to enable MSM high speed UART driver for loire and tone platforms.


humberos:
@jerpelea 
could you check uart again.
DT was missed in last PR.

oshmoun:
testing doesn't show any regressions concerning bt

PR #1378, opened 2017-08-11
[57r1] Introduce refactored ci13xxx driver for MSM

kholk:
This driver has been refactored almost completely, on the base
of the mainline (k4.4) chipidea driver.
Now we correctly (mostly) use the new UDC APIs as found in
modern kernels, plus we support Multi-TD, Isochronous and
interrupt transfers, and much, much more.


kholk:
Opening PR from another branch including all the required updates for Loire USB.

PR #1379, opened 2017-08-11
[57r1] Loire USB: PHY, CI13xxx, smbcharger

kholk:
This patchset includes all the commits to start USB on Loire.

Please note that there's still one disconnection bug that will make Android "think" that the cable is still attached, but battery is not charging. The DISCONNECT uevent is getting sent fine.
This breaks deep sleep.


kholk:
USB cable disconnection kernel panic bug fixed.

kholk:
Critical SPMI fix pushed in this PR

PR #1380, opened 2017-08-13
[57r1] Update msm_audio_ion for QCOM IOMMU v1

kholk:
When IOMMU assisted allocations are enabled, we need to get a
handle to an IOMMU master (specified in DT): this is being
actually done by calling arm_iommu_attach_device and that's ok.

The driver is not accounting for clean IOMMU drivers: it does
expect the IOMMU driver to auto-register its masters at probe
time which is not a common practice, infact the only driver
doing this is arm-smmu.
We need to use the msm_audio_ion driver with a new, clean and
Linux API adhering IOMMU driver (msm_iommu-v1) that is adding
the masters through the of_xlate IOMMU API callback: this
requires setting up the DMA-API mappings properly, which will
also configure the IOMMU through the appropriate standard API
calls.

### DIFF
--- a/arch/arm/boot/dts/qcom/msm8956.dtsi
+++ b/arch/arm/boot/dts/qcom/msm8956.dtsi
@@ -2830,6 +2830,7 @@
 		compatible = "qcom,msm-audio-ion";
 		qcom,smmu-version = <2>;
 		qcom,smmu-enabled;
+		qcom,smmu-force-sid = <0x0 0x1>; /* u64 */
 		iommus = <&apps_smmu 14>;
 	};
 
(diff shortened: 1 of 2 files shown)


kholk:
MSM8956 tested. Driver probing, IOMMU attach OK.

This PR updates the msm_audio_ion driver to properly use the DMA-API unconditionally (because it's the right thing to do!!!!) and, although there's no reason for them to break, functionality of SoCs using ARM SMMU drivers (8996, 8998) has to be verified.

In the event of ARM SMMU powered SoC breakage, the eventual issues will have to solved in the ARM SMMU driver and NOT BY IFDEFS in there.

oshmoun:
msm8996
audio output and input still work

kholk:
ARM-SMMU verified.

kholk:
WAIT TO MERGE --- Commit has an unwanted debugging line.

kholk:
New version of the whole thing. Adds more commits, including a necessary quirk for old IOMMUv1 (v1, v2 and mmu-500) where we have static domains and SIDs.
Now audio starts FINE.

kholk:
MSM8956 DT changes pushed.
Patchset is complete.

Audio is working on Loire!

MERGEME.

PR #1394, opened 2017-08-18
[LA.UM.5.7.r1] Android Security Bulletin - July 2017

pablomh:
Patches for kernel related CVEs.

- Compile tested for teyaki (tone platform).


bartcubbins:
typo in description

PR #1395, opened 2017-08-18
Upstream kernel patch v3.10

bananafunction:
- Patched incrementally upstream kernel patches starting from v3.10.85 until v3.10.107 from kernel.org
- Tested compile after last patch (v3.10.107)
- fixed build errors
- basic functional testing on device suzuran
- found sound issue -> reverted one change in file sound/soc/soc-compress.c


kholk:
I don't mean to discourage this, but basic functional testing is not enough to merge a patchset like this. 
You should carefully test it and make sure that you have no overhead, no increased battery consumption and that the kernel stays rock solid. 
Take all the time you need to test and then report back. 


Thanks!

bananafunction:
@kholk: Thank you for your feedback. Obviously, you are right. At this point, there is at least one question remaining:
Is there a defined test procedure with described steps for all sorts of testing (like component tests, integration tests and, of course, subsystem and/or systemlevel tests)? If there is a procedure, I could do the testing on my PC and device suzuran only.
So basically, what I want to know is, what should be done to fulfill your request :-)

kholk:
You should use the device for a week or so as normal and everything should work as expected.
Stress test it with some benchmarks and then do heavy multitasking, if it doesn't crash or performance didn't degrade, then perfect.

If you want to do serious stability testing (I'd appreciate it), then build and run these tests:
https://source.codeaurora.org/quic/la/platform/vendor/qcom-opensource/kernel/kernel-tests

P.S.: Be sure of picking the right branch!


About the battery consumption test, of course I have no specific test for you, but well, first of all, you can check how much battery consumption you get in music playback, standard browsing (there are some apps for that if I recall correctly) etc.
That test won't be a precise one (unless you wanna spend 3 weeks only for this), but we actually just need a rough one to say that we didn't get any degradation of more than 3-5% in power usage.


That's basically it!

bananafunction:
Ok :-) I'll give it a try. Once again, thank you for the given information.

scafroglia93:
There's a lot of kang inside your PR. Please fix the entire autorship

danielgusvt:
Yes, I agree with @scafroglia93
I would like to see something like this https://gist.github.com/nathanchance/1737edb19f7d46127b8775218d603fa6
You can use this guide for reference https://forum.xda-developers.com/android/software-hacking/reference-how-to-upstream-android-kernel-t3626913

bananafunction:
Hi @scafroglia93 and @dani020110,
thank you for your feedback. I understand your request and you are right. As I merged the upstream kernel changes offline into the codebase and not via git, the authorship of this pull request is not correct. I'd like to say that it was not my intention to "adorn myself with borrowed plumes". Sorry for that. I will close this pull request and try again later with less "formal" issues.
Thank you all.

PR #1398, opened 2017-09-02
power: qpnp-smbcharger: release HVDCP wakelock upon usb removal

oshmoun:
When usb is disconnected, HVDCP detection work is canceled, but the wakelock is not released.
Release it to let the device sleep again.


kholk:
MERGEME

PR #1399, opened 2017-09-07
[LA.UM.5.7] drivers: tty: serial: Fix bluetooth serial communication

humberos:
It is not required anymore.

This patch fixes serial communication for Broadcom BT targets.

Change-Id: I4707b5d514a500a005b212487ca2a163b5153bb6


humberos:
Tested on kagura.
Requires loire confirmation.

humberos:
Seems that deep sleep state is broken by using this patch.

humberos:
ok, I've found the problem... will upload new PR.

PR #1407, opened 2017-09-11
defconfig: kanuti: cleanup

danielgusvt:
Works fine


erikcas:
not nessecary afaik. On tulip, cyttsp4 and cyttsp5 panels are seen afaik. If 4 init fails, 5 inits or vice versa

bartcubbins:
yes we don't use QPNP_LINEAR_CHARGER and QPNP_VM_BMS:
[    7.713061] BMS: bms_get_adc: vadc not found - defer probe rc=-517
[    7.719101] BMS: qpnp_vm_bms_probe: Failed to get adc rc=-517
[    7.724814] spmi 00-qcom,vmbms: Driver qcom,qpnp-vm-bms requests probe deferral
[    7.732239] CHG: qpnp_lbc_main_probe: Failed to read DT properties rc=-517
[    7.738985] spmi 00-qcom,charger: Driver qcom,qpnp-linear-charger requests probe deferral

For USB we use only ci13xxx so USB_DWC3_MSM can be disabled too.
About TOUCHSCREEN_CYPRESS_CYTTSP5_DEBUG. Disabling can be useful on our Tulip devices with CYTTSP5 because we get a lot of debugging stuff in kernel logs.
@jerpelea so I think it can be merged:)

PR #1417, opened 2017-09-15
[LA.UM.5.7.r1] Update tone config

humberos:
ping @oshmoun @alviteri


oshmoun:
did you notice any differences with ALLOC_BUFFERS_IN_4K_CHUNKS? tested that before but saw no changes.
About USB_DWC3_MSM_ID_POLL and Enable DUAL_ROLE_USB_INTF, would those affect charging?

kholk:
The ID polling should be deprecated.
The PMIC found in msm8996 and the other one in msm8998 supports over current protection (OCP) and short circuit protection (SCP, referenced in code as SC) so, even in the event of water contamination of the usb port while in host mode, the PMIC will be self protected.

About the dual role, that shouldn't be touching any of the code that we are currently executing.

The 4K bugger chunks then is a debugging feature and it's not intended for production use.

kholk:
Whoops.
*buffer

Cannot edit from phone!

oshmoun:
ID polling is enabled for yoshino though
https://github.com/sonyxperiadev/kernel/blob/aosp/LA.UM.5.7.r1/arch/arm64/configs/aosp_yoshino_maple_defconfig#L445

kholk:
Yeah, it's a fancy way to recognize other proprietary (and non-proprietary) accessories, other than a paranoid third protection to avoid to even trigger host mode when not necessary.

humberos:
Thanks for the explanation

PR #1425, opened 2017-09-23
don't export eventpoll.h to userspace

alviteri:
* Just conflicts with userspace headers and isn't needed anyway

fixes error in build with inline kernel headers

/obj/KERNEL_OBJ/usr/include/linux/eventpoll.h:75:8: note: previous definition is here
struct epoll_event {
       ^
1 error generated.


humberos:
Thanks!

jerpelea:
please remove the line instead of comment!

kholk:
This breaks GNU libc usages of this entire kernel tree, hence is wrong. 

The solution is to use an ifndef for Android in the header.

PR #1426, opened 2017-09-23
	headers: don't export eventpoll.h on Android builds

Myself5:
* Just conflicts with userspace headers and isn't needed anyway


ItzLevvie:
@Myself5 https://github.com/sonyxperiadev/kernel/pull/1425?

Myself5:
@kholk @jerpelea

kholk:
CONFIG_ANDROID is used for Android specific drivers. This will break GNU libc builds containing Android specific drivers.

The Android build system should set the __ANDROID__ variable

Myself5:
#1431 is a better approach to that.

PR #1435, opened 2017-09-29
[57r1] power: reset: Fix reboot to FOTAKernel recovery on some Sony devices

artem:
Sony uses non-standard way to enter recovery on devices with S1 bootloader (all?)
Make kernel do the same thing for "reboot recovery" as it does for "reboot oem-46". At least for loire and tone.


kholk:
Please check if yoshino needs the same.

artem:
Tone: ok

artem:
@kholk this change should be applied only for devices with S1 bootloader (pre-yoshino)

alviteri:
what about this ? sorry ... on sonyaosp the following steps work

adb reboot recovery
power menu - reboot recovery (own implementation)

all is working without this ......

PR #1436, opened 2017-09-29
[LA.UM.5.7.r1] Avoid loire logspam by skipping unsupported regulator op

oshmoun:
on loire devices, gdsc does not support drms, but calls to regulator_set_load are still made.
This is especially problematic in mdss_io_util, where calls are frequent and cause logspam.
Fix it by avoiding the unsupported call if a property in dt is declared.

### DIFF
--- a/drivers/clk/msm/mdss/mdss-pll-util.c
+++ b/drivers/clk/msm/mdss/mdss-pll-util.c
@@ -260,6 +260,9 @@ static int mdss_pll_util_parse_dt_supply(struct platform_device *pdev,
 
 		mp->vreg_config[i].post_off_sleep = (!rc ? tmp : 0);
 
+		mp->vreg_config[i].drms_unsupported = of_property_read_bool(supply_node,
+					"qcom,supply-drms-unsupported");
+
 		pr_debug("%s min=%d, max=%d, enable=%d, disable=%d, preonsleep=%d, postonsleep=%d, preoffsleep=%d, postoffsleep=%d\n",
 					mp->vreg_config[i].vreg_name,
 					mp->vreg_config[i].min_voltage,
--- a/include/linux/mdss_io_util.h
+++ b/include/linux/mdss_io_util.h
@@ -58,6 +58,7 @@ struct dss_vreg {
 	int post_on_sleep;
 	int pre_off_sleep;
 	int post_off_sleep;
+	bool drms_unsupported;
 };
 
 struct dss_gpio {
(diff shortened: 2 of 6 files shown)


jerpelea:
if you use a dt property why do you also ifdef ?
it is enough to check if property is declared!

oshmoun:
True
I tried to minimize the performance hit on other platforms, but that was probably a bit too overzealous. Especially since the addition is a simple boolean check that shouldn't affect performance. 
Will remove the ifdefs.

kholk:
Can't we check if the regulator is a gdsc instead?!

oshmoun:
That should be possible. The dt parsing logic can be replaced with a strcmp call.
We'll still need the new field in the struct this way though.
Also, we'll need to add an ifdef for loire,  since there are gdsc nodes in tone dt.  Afterall, we don't want this change to apply to a platform that isn't affected.

kholk:
AFAIK, there is no GDSC capable of DRMS..... So... Well, it's curious. 
Let me take a look at this, please hold the PR open for a day or two!

oshmoun:
just for reference
here are the gdsc nodes for 8996
https://github.com/sonyxperiadev/kernel/blob/aosp/LA.UM.5.7.r1/arch/arm/boot/dts/qcom/msm8996-mdss-pll.dtsi

include/linux/mdss_io_util.h
```diff
@@ -58,6 +58,7 @@ struct dss_vreg {
 	int post_on_sleep;
 	int pre_off_sleep;
 	int post_off_sleep;
+	int drms_unsupported;
```
kholk: bool drms_unsupported

kholk (changes requested):

aolexiy:
Instead of introducing yet another property we could pick from mainline, see https://github.com/sonyxperiadev/kernel/pull/1441

The change introduces regulator_ops_is_valid() function to be used as a generic validator and it'll silently skip drms_uA_update call if REGULATOR_CHANGE_DRMS is not supported.

oshmoun:
The difference is that this approach doesn't modify core Linux API, only the affected drivers.
But the question about how tone isn't affected remains open. Maybe answering that can give an insight as to what's the main cause.

drivers/clk/msm/mdss/mdss-pll-util.c
```diff
@@ -260,6 +260,9 @@ static int mdss_pll_util_parse_dt_supply(struct platform_device *pdev,
 
 		mp->vreg_config[i].post_off_sleep = (!rc ? tmp : 0);
 
+		mp->vreg_config[i].drms_unsupported = of_property_read_bool(supply_node,
```
kholk, line 263: if regulator rdesc.name contains "gdsc", then drms_unsupported = true;

That's the right way of doing it.

kholk (changes requested):

drivers/clk/msm/mdss/mdss-pll-util.c
kholk: P.S.: The regulator name is always "gdsc_xxxx"

PR #1440, opened 2017-10-04
ASoC: wsa881x: Fix wsa881x speaker damage

alviteri:
wsa881x speaker damage has been reported due to DC at
wsa output. Add a register setting which will reset
the soundwire when there are any
DC or synchronization issues.

CRs-Fixed: 2086505
Change-Id: I929d2eb7d096005e5089f3bfabded6901d81439e
Signed-off-by: Vatsal Bucha <vbucha@codeaurora.org>


kholk:
should be fine

PR #1441, opened 2017-10-05
regulator: refactor valid_ops_mask checking code

aolexiy:
To make the code more compat and centralized, this patch add a
unified function - regulator_ops_is_valid. So we can add
some extra checking code easily later.

Signed-off-by: WEN Pingbo <pingbo.wen@linaro.org>
Signed-off-by: Mark Brown <broonie@kernel.org>
(cherry picked from commit 8a34e979f684aa13e6c4bf23b394cca9dfabf4a9)

### DIFF
--- a/drivers/regulator/core.c
+++ b/drivers/regulator/core.c
@@ -135,6 +135,19 @@ static bool have_full_constraints(void)
 	return has_full_constraints || of_have_populated_dt();
 }
 
+static bool regulator_ops_is_valid(struct regulator_dev *rdev, int ops)
+{
+	if (!rdev->constraints) {
+		rdev_err(rdev, "no constraints\n");
+		return false;
+	}
+
+	if (rdev->constraints->valid_ops_mask & ops)
+		return true;
+
+	return false;
+}
+
 static inline struct regulator_dev *rdev_get_supply(struct regulator_dev *rdev)
 {
 	if (rdev && rdev->supply)
@@ -201,28 +214,13 @@ static struct device_node *of_get_regulator(struct device *dev, const char *supp
 	return regnode;
 }
 
-static int _regulator_can_change_status(struct regulator_dev *rdev)
-{
-	if (!rdev->constraints)
-		return 0;
-
-	if (rdev->constraints->valid_ops_mask & REGULATOR_CHANGE_STATUS)
-		return 1;
-	else
-		return 0;
-}
-
 /* Platform voltage constraint check */
 static int regulator_check_voltage(struct regulator_dev *rdev,
 				   int *min_uV, int *max_uV)
 {
 	BUG_ON(*min_uV > *max_uV);
 
-	if (!rdev->constraints) {
-		rdev_err(rdev, "no constraints\n");
-		return -ENODEV;
-	}
-	if (!(rdev->constraints->valid_ops_mask & REGULATOR_CHANGE_VOLTAGE)) {
+	if (!regulator_ops_is_valid(rdev, REGULATOR_CHANGE_VOLTAGE)) {
 		rdev_err(rdev, "voltage operation not allowed\n");
 		return -EPERM;
 	}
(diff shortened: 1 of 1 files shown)


aolexiy:
This fixes an annoying debugging spam from regulator_check_drms()  for gdsc_mdss regulator

kholk:
that's not a fix.....

aolexiy:
this doesn't address the question of why current mdss code is doing this for 8956, yes, but this is better than introducing yet another property to stop the spam

the change is effectively very similar to #1436

kholk:
Well, no, not really.
While the solution stops introducing new properties etc (yes I absolutely get your point), this commit is still not valid: when you try to do an unsupported operation on a critical part involving critical hardware and eventually critical software, it's good to throw an error to say "hey, something is wrong", because it's most probably a human mistake.

Trying to do an unsupported operation on a regulator (like the drms) may imply a big mistake in either your driver implementation or in what you think you have to do to make something to work.
Bad regulators (or clocks, eh!) implementation will possibly lead to _really_ random errors in your driver because you may be relying on what "something" (PBL? SBL? whatever else?) has pre-set for you, while you may need to scale accordingly to a certain voltage, not only to get better power efficiency, but also to get the optimal conditions for full stability of your hardware.

Now, what we are dealing with is a driver trying to call DRMS on a GDSC: that is (as we know) an unsupported operation because of the nature of the GDSC itself.
GDSC may or may not be a regulator: it's used to enable "small" (in amperage and/or voltage) SoC internal rails (powered by an always-on supply or an on-demand one coming, from example, from the PMIC)... and it's also used (properly or improperly, that's not the point and I don't want to open that discussion) as a switch to enable or disable clocks.

The GDSC is Qualcomm specific, so it's SoC specific, so if we've got a problem that is specific to _us_ the first rule is to NEVER touch the Linux kernel global APIs, especially on things that are so much important and critical such as regulators.



In the PR #1436 we are adding a particular exclusion from DT to not permit DRMS to be called.
I suppose you want to know why the current MDSS code calls that and it's simple: this kernel tag is not made for legacy SoCs and Qualcomm usually breaks compatibility simply because they... don't care.
The newer SoCs, like MSM8996 and MSM8998 (and others), are using a real regulator (a rail coming from the PMIC), for which is allowed to do a DRMS operation for two reasons: overcurrent protection (OCP) and power efficiency.


This is the reason why your PR is not a valid fix.
In case you have any more doubts, feel free to ask.

aolexiy:
Why are you saying it's an API change? This patch improves the internals of regulators core, functionality  wise nothing is changed. Side-effect is that it shuts down that debugging spam noise.

Since it's mainline change and you'll get it anyways with 4.9 to me it's better than to add local changes which are doing the same. By "the same" here I mean that you will skip the set_load call in both cases and avoid polluting logs.

PS: please don't think of this proposal as of mdss fix

kholk:
Uhm, I've misread the regulator_check_drms usage. Apparently, even if it returns -EPERM, the drms_uA_update function is still returning zero even if the regulator_check_drms returns an error code and effectively the commit uses the same logic, so yes, my bad, sorry for saying that it's an API change.

The commit can be merged safely as the logic is the same.

About the latter thing, that's most probably just a way to think: I personally think that, in any case, the MDSS (camera also had a similar problem on the cpp gdsc) or any other driver for that matter, should not even _try_ to do an unsupported operation on a regulator.
As I said, I see the regulators as a critical part.

My main (and only) concern is that in the future things may change, QC may update the GDSC driver  and we may trigger some unexpected behavior by doing DRMS on a GDSC: this is why I think that the best option is to _prevent_ any driver to do this.
Yes you may say I'm paranoid. Yes, on critical things I am... for good reasons that I've learnt at my expense.

Anyway, when I extensively reply to something, I always try to cover all the cases and any doubt anyone can have... and yeah, I tend to think about the worst reasons a person may have when committing something because of various experiences I had and please don't think of my reply/replies as me wanting to be rude or unpleasant with you (or anyone else)!

aolexiy:
@kholk no worries, thanks for the review. Btw, I'm fine with different kinds of language so if you'll spot anything and use strong words no offence will be taken :)

kholk:
I appreciate that. Thanks!

PR #1431, opened 2017-09-28
[LA.UM.5.7] headers: Do not export eventpoll.h on android build system

humberos:
ANDROID_BUILD_TOP is a global variable for android build system

https://android.googlesource.com/platform/build/+/master/envsetup.sh#297

It can be used to avoid conflicts with userspace headers.

Signed-off-by: Humberto Borba <humberos@omnirom.org>
Change-Id: Ib596fce6abfd0f23f21878c65ddd0ecf187d97f0

PR #1444, opened 2017-10-05
epoll: Drop struct epoll_event definition

alviteri:
Add kernel header sanitizer check to drop struct epoll_event.
This struct epoll_event is not following the POSIX standard
and defining element in opaque data type.

Change-Id: I1eecef1b64eb1fb8f0dd6999263093a453e7b9fb
Signed-off-by: Prasad Sodagudi <psodagud@codeaurora.org>

### DIFF
--- a/include/uapi/linux/eventpoll.h
+++ b/include/uapi/linux/eventpoll.h
@@ -56,6 +56,7 @@
 #define EPOLL_PACKED
 #endif
 
+#ifdef __KERNEL__
 struct epoll_event {
 	__u32 events;
 	__u64 data;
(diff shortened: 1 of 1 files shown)


kholk:
I suppose it works. My solution, however, required more than just this.....
Please verify with an inline android build.

alviteri:
already tested on oreo..... nougat should be priority for this

humberos:
oh I think it is the best approach and also it gets things isolated in kernel build.
So we have a master's decision coming between this one and https://github.com/sonyxperiadev/kernel/pull/1431

PR #1448, opened 2017-10-09
[57r1] Adreno performance fixes

kholk:


oshmoun:
the header change should probably also be applied in common-headers.

PR #1454, opened 2017-10-10
Recheck if CPU is balanced if cpufreq stays constant

djselbeck:
This fixes an edge case where no re-evaluation of CPU balance
is performed. This could happen with the following situation:

3 cores online, core_0 and core_1 at around 100 load and max cpufreq.
 * core_3 clocks down below idle_bottom_freq
 * rqbalance_work_func will result in shutdown of core_3
 * repeat: rqbalance_work_func stays in rqbalance_state == DOWN
         => No re-evaluation of CPU load balance until load
            is reduced and therefore cpufreq changes

### DIFF
--- a/drivers/cpuquiet/governors/rqbalance.c
+++ b/drivers/cpuquiet/governors/rqbalance.c
@@ -563,8 +563,8 @@ static void rqbalance_work_func(struct work_struct *work)
 		else if(num_online_cpus() == nr_cpu_ids)
 			stop_load_timer();
 
-			queue_delayed_work(rqbalance_wq,
-						 &rqbalance_work, up_delay);
+		queue_delayed_work(
+			rqbalance_wq, &rqbalance_work, up_delay);
 		break;
 	case UP:
 		balance = balanced_speed_balance();
--- a/drivers/cpuquiet/governors/rqbalance.c
+++ b/drivers/cpuquiet/governors/rqbalance.c
@@ -555,8 +555,11 @@ static void rqbalance_work_func(struct work_struct *work)
 		break;
 	case DOWN:
 		cpu = get_slowest_cpu_n();
-		if (cpu < nr_cpu_ids)
+		if (cpu < nr_cpu_ids) {
 			up = false;
+			// Next time recheck if CPU is balanced
+			rqbalance_state = UP;
+		}
 		else
 			stop_load_timer();
 
--- a/drivers/cpuquiet/governors/rqbalance.c
+++ b/drivers/cpuquiet/governors/rqbalance.c
@@ -560,7 +560,7 @@ static void rqbalance_work_func(struct work_struct *work)
 			// Next time recheck if CPU is balanced
 			rqbalance_state = UP;
 		}
-		else
+		else if(num_online_cpus() == nr_cpu_ids)
 			stop_load_timer();
 
 			queue_delayed_work(rqbalance_wq,


Bitti09:
@djselbeck I'll test and report tomorrow

_Sent from my Xperia XZ (AOSP) using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

kholk:
Makes some sense, but needs careful testing.

Bitti09:
first short test with   360° youtube and pinout game : much less stuttering than before.

alviteri:
@kholk first test pass on dora seeing galeazzi's videos on 1080p60fps on youtube ..... device didn't freeze on android oreo ...... will report other things

kholk:
That's good guys, thank you!
Now please test this on a daily driver device and please compare heat and power consumption.

I will test it in 1 to 2 days. Please test for at least 2 days.

alviteri:
Mobile Legends game ok on tone

djselbeck:
I've been running it on my daily driver since opening this PR on kagura. 

Also I've tested if it powers down to 1 core which happens but very rarely because the core 0 and 1 are tied together by cpufreq and the cpufreq based rqbalancing kicks in often.

kholk:
Don't misunderstand me, the logic IMHO is completely fine and the change is also good.

Being this a very important driver, common between *all* devices, I am paranoid about anything that can happen due to this change. This is why I'm asking you all to thoroughly test it.

djselbeck:
Of course, I can completely understand this ;).

alviteri:
Probably is not related but poplar is a disaster with this .... It lags as hell .... Need more test

Bitti09:
@alviteri for me there are no lags with this PR

_Sent from my Xperia XZ (AOSP) using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

alviteri:
@Bitti09 as angelo said needs to be tested on all

kholk:
Interesting. The change actually tends to increase the computational power, not to decrease it, as it makes the governor more reactive to upcore decision. 

Please be sure that the lags are related to this pr.

drivers/cpuquiet/governors/rqbalance.c
```diff
@@ -555,13 +555,18 @@ static void rqbalance_work_func(struct work_struct *work)
...
+			rqbalance_state = UP;
+		}
 		else
 			stop_load_timer();
 
-			queue_delayed_work(rqbalance_wq,
-						 &rqbalance_work, up_delay);
+
+
+		queue_delayed_work(rqbalance_wq,
```
alviteri: @djselbeck can you update this ..... string 558 - 562 are the commit string 566-569 is useless

kholk: I agree about the useless whitelines

djselbeck: Yes the lines should be fixed, sadly I noticed them to late. But the indentation should be fixed I guess?

oshmoun:
would it make sense to make the recheck happen less frequently?
would something like this pseudo code work?
```
if (cpu < nr_cpu_ids) {
    if(all_online_cores_are_at_max_freq()) {
        // force a recheck because cpufreq has reached its upper limit
         rqbalance_state = UP;
   }
}
```
or would such code produce even more overhead than forcing the recheck every time?

drivers/cpuquiet/governors/rqbalance.c
kholk: I don't see any indentation problem

kholk:
What you propose introduces unnecessary overhead.......
Probably we should just restore rqbalance's functionality as per my original design.

drivers/cpuquiet/governors/rqbalance.c
alviteri: you should rebase ... this is useless commit

alviteri:
https://github.com/SonyAosp/kernel/commit/6a91672dbfb2e401dde3aac87e190ec12d1b9493
https://github.com/SonyAosp/kernel/commit/f4b9b6561d7255eb9e2b61771c35aaebe7fbf94f

djselbeck:
The code should be ok now?

Bitti09:
any updates on this ?

RyTekk:
It should be closed, doesn't it?

kholk:
Yep. Another solution was applied.

djselbeck:
Yep, close this. Although I can still trigger the old behaviour on kagura. If I've time i'll check if I can find the issue with the new solution.

PR #1456, opened 2017-10-12
power: qcom: qpnp-smbcharger: Fix for slow charging 

Bitti09:
If power supply type is POWER_SUPPLY_TYPE_UNKNOWN at
bootup, heathd, fails to parse power supply properties
correctly -> slow charging.

Setting the power supply type
during bootup to USB when the charger is not plugged seems
to fix this issue

Adapted from here: https://review.lineageos.org/#/c/191247/

 Not tested at the moment (newbuild in progress)

---
This change is [Reviewable](https://reviewable.io/reviews/sonyxperiadev/kernel/1456)

### DIFF
--- a/drivers/power/supply/qcom/qpnp-smbcharger.c
+++ b/drivers/power/supply/qcom/qpnp-smbcharger.c
@@ -4924,11 +4924,11 @@ static void handle_usb_removal(struct smbchg_chip *chip)
 		chip->typec_current_ma = 0;
 #ifdef CONFIG_QPNP_SMBCHARGER_EXTENSION
 	chip->somc_params.chg_det.settled_not_hvdcp = false;
-	chip->somc_params.chg_det.sub_type = POWER_SUPPLY_SUB_TYPE_UNKNOWN;
+	chip->somc_params.chg_det.sub_type = POWER_SUPPLY_SUB_TYPE_FLOATED;
 	smbchg_relax(chip, PM_DETECT_HVDCP);
 	cancel_delayed_work_sync(&chip->hvdcp_det_work); // CHECKME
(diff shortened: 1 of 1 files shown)


oshmoun:
this seems to work, but the cause must be found out!

zachariasmaladroit:
https://github.com/zachariasmaladroit/android_kernel_sony_msm8994_kitakami_r2/commit/efe5c3e551838463daaada64f72f5d15e30576c8 Set default power supply type as USB

that change is in drivers/usb/dwc3/dwc3-msm.c

Bitti09:
@oshmoun do you think the first line change is ok? I see now that I have changed that from sub type  unknownto type USB which doesn't exist.

_Sent from my Xperia XZ (AOSP) using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

kholk:
Test it. If the test succeeds on ***_Loire_*** and Tone, then it's safe to merge, but remember to FIRST squash the two commits.

Also the commit has to be formatted properly.

Good job!

oshmoun:
just by checking battery uevent, which reports about -1300mA current, this seems to be working on loire.
also left the device charging, and it seems to recharge at a 1%/minute rate.
looks good to go

EDIT: corrected the unit of measurement for recharge rate!

kholk:
Fine. Critical PR: MERGEME!!

Bitti09:
@oshmoun  what does dmesg says for you when putting in an wall charger?
 Avail curr from USB = ?     
for normal pc it says  500 for me.

kholk:
It's normal, pc USB draws 500mA max as per spec.

PR #1459, opened 2017-10-14
power: reset: exclude keyaki device from special recovery handling

oshmoun:
keyaki misbehaves when trying to enter recovery using oem-46.
The device enters a weird mode, that can only be exited by using fastboot or flash mode.
Even trying to force shutdown the device does not work.
This has been tested and confirmed occuring on older and the latest bootloader.
Therefore, disable reboot to recovery on keyaki using the oem code.

Change-Id: I0f8e8bd013dbc96538590c0dd8e3fd19db3d8428


oshmoun:
not needed anymore after the latest bugfixes concerning reboot reasons. Now reboot to recovery works fine.

PR #1466, opened 2017-10-21
[57r1] [PR 1/2] msm-poweroff cleanup to restore pstore functionality

kholk:
This is the first of a two-parts PR, because the second part is not ready yet.

This first part starts a necessary cleanup of the PMIC reboot reasons, so that the pstore functionality is restored and the reboot is finally completely clean.

The second part will fix the reboot reasons for all platforms to be able to normally reboot to recovery by correctly informing the bootloader about which boot partition to choose in the warm reboot.


kholk:
Build problems on LOIRE. Will fix and reopen when part 2 will be done.

PR #1467, opened 2017-10-21
[57r1] msm-poweroff: Fix for kernel panic and reboot reasons

kholk:
This PR refactors the handling of reboot reasons for SoMC devices.
Now the code is divided by bootloader (S1BOOT or XBOOT) and not by platform anymore.

Now the reboot reasons are okay and we inform the bootloader correctly: the device can cleanly reboot to fastboot, recovery and back to system.
Moreover, the pstore functionality is restored, as now we are correctly detecting kernel panic conditions and choosing to warm reboot the board, so that RAM contents are preserved.

### DIFF
--- a/drivers/power/reset/msm-poweroff.c
+++ b/drivers/power/reset/msm-poweroff.c
@@ -198,6 +198,9 @@ static int dload_set(const char *val, struct kernel_param *kp)
 #else
 static void set_dload_mode(int on)
 {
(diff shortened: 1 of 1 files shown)


Bitti09:
for some reason after this commit an 
adb reboot recovery
does a  normal reboot instead of recovery start on my kagura

PR #1474, opened 2017-10-24
ASoC: msm: Add hwdep node for NOIRQ FE

alviteri:
ALSA lacks support to accept shared memory supplied
from userspace. But some usecases need to get a shareable
file descriptor for the dma buffer. As an alternate
to adding custom ioctls to the ALSA framework, use
a hwdep node to get a file descriptor to the underlying dma buffer.
Maintain compat mode as well.

CRs-Fixed: 2041151
Change-Id: Id783bb84e8ef59b28e42d982903971625577b9a3
Signed-off-by: Haynes Mathew George <hgeorge@codeaurora.org>

### DIFF
--- a/include/uapi/sound/devdep_params.h
+++ b/include/uapi/sound/devdep_params.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2013-2015, The Linux Foundation. All rights reserved.
+ * Copyright (c) 2013-2015,2017, The Linux Foundation. All rights reserved.
  *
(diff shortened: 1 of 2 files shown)


alviteri:
this is needed now by audioHAL android-8.0.0_r21 tag

https://android.googlesource.com/platform/hardware/qcom/audio/+/96483a2f0d29db67d9246a66c6e4cb4874185742%5E%21/#F0

PR #1475, opened 2017-10-25
[57r1] Loire camera joy

kholk:


arch/arm/boot/dts/qcom/msm8956.dtsi
```diff
@@ -1830,6 +1830,8 @@
                 interrupt-names = "ipa-irq", "bam-irq";
                 qcom,ipa-hw-ver = <6>; /* IPA core version = IPAv2.6L */
                 qcom,ipa-hw-mode = <0>; /* IPA hw type = Normal */
+		qcom,wan-rx-ring-size = <192>; /* IPA WAN-rx-ring-size*/
```
oshmoun: indentation!

kholk: I know. I've indented it badly on purpose.
My mind is strange, don't try to understand the reasons... :P :P :P

kholk:
Whitespaces fixed on both ipa and rmnet-ipa nodes.

PR #1476, opened 2017-10-25
[57r1] Loire updates + driver upgrades

kholk:
Tested on Kugo, Suzu, Poplar

### DIFF
--- a/arch/arm/boot/dts/qcom/msm8956-gpu.dtsi
+++ b/arch/arm/boot/dts/qcom/msm8956-gpu.dtsi
@@ -118,6 +118,15 @@
 		coresight-child-list = <&funnel_in0>;
 		coresight-child-ports = <4>;
 
+		/* Enable context aware freq. scaling */
+		qcom,enable-ca-jump;
+
+		/* Context aware jump busy penalty in us */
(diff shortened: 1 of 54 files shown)


oshmoun:
causes build errors on keyaki in the current state
```
../../../../../../kernel/sony/msm/drivers/media/platform/msm/sde/rotator/sde_rotator_dev.c:469:20: warning: passing argument 1 of 'cancel_work_sync' from incompatible pointer type 
error, forbidden warning: sde_rotator_dev.c:469
```

alviteri:
last commit needs an update in cameraHAL too

kholk:
the warning has to be fixed.
the commit that should need an update to camera hal is refactored and marked as PORT, so that it does NOT require any hal upgrade.

alviteri:
https://github.com/SonyAosp/platform_hardware_qcom_camera/commits/clock

kholk:
Don't commit this on hal. I want to stay compatible with our caf tag and that's why I modified the commit.

kholk:
Alternatively, I will just remove the modified commit and we will evaluate the benefits. After that we'll see.

PR #1486, opened 2017-10-30
4.4 binder from wahoo kernel

alviteri:


kholk:
Well, it makes sense, but.... Can't we avoid the full reverting of binder?

alviteri:
Like ?

PR #1487, opened 2017-11-01
[57r1] SDIO fixes for main topic BCMDHD

kholk:
There are some SDIO cards that don't support plug in/out in
a standard way: once they are plugged in, it will not be
possible to plug them out and retain functionality through
suspend and resume.
This patch flags the card as NONREMOVABLE only after it has
been detected (otherwise, the detection handler will not
work).

In the example case of a BCM4356 SDIO card, the system may
start with no/unpowered card. We powerup the card and the
kernel recognizes it, then the specific driver for the chip
loads a firmware into its RAM: starting from this moment,
if we power down the card, we will not be able to power it
up again, since it will lose the firmware stored in RAM
and will never boot up anymore.

To make things worst, the whole BCM chip may be powered
down when the SDIO card gets a power cut, producing a
functionality breakage: the chip will be disconnected from
the WiFi network and the connection loss will be permanent,
as no sdiowakeup interrupt will fire: the chip is down.

Such cards do _not_ support any change in voltage once the
card setup is done, otherwise Bad Things Happen (TM), hence
the only way to get them working properly is to suspend the
host controller while keeping the card completely alive and
relying on its wireless driver to set it to low power mode.

Please note that, about power consumption, mileage may vary:
some wireless chips support a very very low power mode with
connection keepalive, some others do not, but this case is
not covered in this patch and will never be in this place,
as this is intended to only address problems coming from
the Linux MMC API, and whatever else happens out of it must
be managed out of it.

### DIFF
--- a/drivers/mmc/core/sdio.c
+++ b/drivers/mmc/core/sdio.c
@@ -1168,9 +1168,11 @@ out:
 static int mmc_sdio_runtime_suspend(struct mmc_host *host)
 {
 	/* No references to the card, cut the power to it. */
-	mmc_claim_host(host);
-	mmc_power_off(host);
-	mmc_release_host(host);
+	if (!mmc_card_keep_power(host)) {
+		mmc_claim_host(host);
+		mmc_power_off(host);
+		mmc_release_host(host);
+	}
 
 	return 0;
 }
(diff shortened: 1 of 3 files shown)


aolexiy:
Please hold this one a bit, I'd like to have a close look at "mmc: host: sdhci-msm: Set sdiowakeup irq to low trigger for BCM", this seems to work with xperia "official" kernel which uses the same default high level trigger.

kholk:
Yes I was expecting that... for sure you will also find the reasons why it works on copylefts ;)

kholk:
After 24 hours, no slowdown, deep sleep OK, wifi wakeup OK on suzu

kholk:
Works on Sony copylefts because of an interrupt handling bug in QC 3.10 branches.

PR #1488, opened 2017-11-01
[LA.UM.5.7.r1] Loire: Enable HZ_1000

bartcubbins:


arch/arm64/configs/aosp_loire_kugo_defconfig
```diff
@@ -43,7 +43,8 @@ CONFIG_SCHED_MC=y
 CONFIG_NR_CPUS=6
 CONFIG_QCOM_TLB_EL2_HANDLER=y
 CONFIG_PREEMPT=y
-CONFIG_HZ_100=y
+CONFIG_HZ_1000=y
+CONFIG_HZ=1000
```
ab123321: This shouldn't be set manually, CONFIG_HZ_1000=y will do the whole job

artem, line 47: This shouldn't be set manually, CONFIG_HZ_1000=y will do the whole job

kholk:
It makes sense, but needs a lil testing.

arch/arm64/configs/aosp_loire_kugo_defconfig
bartcubbins: @ab123321 then it must be removed from tone and yoshino

PR #1489, opened 2017-11-01
drivers: bcmdhd: apply changes to dhd_sdio from 4.4 bcmdhd

oshmoun:
These changes are taken from Samsung's bcmdhd driver, which is used on its 4.4 devices that have bcm wifi.
The changes include:
- a new iovar for suspend
- moving codeblocks that handle sdmode and sdclk further up in the code flow

These are required to remain compatible with the 4.4 API.

Change-Id: I810f6102f01bf8341d27e16af3a62678e89b5b8f


kholk:
MERGEME with #1487

PR #1490, opened 2017-11-01
4.4 gapless

alviteri:


alviteri:
devices tested kugo dora poplar maple and lilac

PR #1491, opened 2017-11-02
dts: msm8956: correct the scaling factor for tsens sensors

oshmoun:
tsens does not need scaling on msm8956, since the values reported are only the integer part of the temperature.
That means there is no need to scale in order to remove the tenths from the reported value.
This fixes the issue where msm_thermal would trigger vdd restriction, since it received a low temperature from tsens.

Change-Id: I3fcf6efdccd66c0173436c4efac21e6f050a4219


kholk:
Correct, mergeme

PR #1492, opened 2017-11-02
dts: msm8956: add disabled-pin pinctrl attribute

aolexiy:
This prevents an access to these when reading gpio status from debugfs,
otherwise it hangs the device.

### DIFF
--- a/arch/arm/boot/dts/qcom/msm8956-loire-common.dtsi
+++ b/arch/arm/boot/dts/qcom/msm8956-loire-common.dtsi
@@ -49,6 +49,10 @@
 };
 
 &soc {
+	pinctrl@1000000 {
+		disabled-pins = <0 1 2 3>;
+	};
+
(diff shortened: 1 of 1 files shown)


aolexiy:
@jerpelea, @kholk - a tiny fixup for the annoying hangup on loire, please have a look

kholk:
The idea is valid -- and yes, it indeed works -- but this is not a standard Qualcomm property, infact it's implemented by additional code from SoMC.
The change belongs to msm8956-loire_pinctrl.dtsi for the explained reason: please put the property there instead.

aolexiy:
@kholk agree, makes sense to use a SoMC platform file. placed into msm8956-loire-common.dtsi instead of msm8956-loire_pinctrl.dtsi to keep the same as on tone & yoshino

kholk:
Thanks for the change.
MERGEME :))))

PR #1494, opened 2017-11-04
[57r1] MSM8956 sanification

kholk:


kholk:
Audio buzzing with deep buffer is finally gone.

kholk:
Forcepushed and removed currently useless and dangerous stuff.

PR #1496, opened 2017-11-04
[DONOTMERGE] drivers: bcmdhd: further updates from 4.4 bcmdhd

oshmoun:
These updates add support for deep sleep, and also update some LINUX_KERNEL_CODE ifdefs.
Also adjusts some headers to include hang detection logic, but that is probably ported incompletely.

Change-Id: Ie41821a3ccbf9cd43a34d06554bc1d8dcb2d5df3


jerpelea:
please split the patch 
you have changes that are kernel version dependent and the deep sleep functionality

oshmoun:
Yup will do that soon

PR #1499, opened 2017-11-06
[LA.UM.5.7.r1] Tone/Yoshino: Remove not needed debug_region memory allocation

bartcubbins:
This will be removed on Loire platform in this commit: https://github.com/sonyxperiadev/kernel/pull/1494/commits/5a18d4485975ab08514e25479ee3430d5bb472be so let's follow it on all our platforms

### DIFF
--- a/arch/arm/boot/dts/qcom/msm8996-tone-common.dtsi
+++ b/arch/arm/boot/dts/qcom/msm8996-tone-common.dtsi
@@ -22,13 +22,6 @@
 	};
 
 	reserved-memory {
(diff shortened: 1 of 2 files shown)


jerpelea:
fix the patches.
you add a script then you remove it

bartcubbins:
@jerpelea updated

jerpelea:
this patch has to be carefully tested on the new loaders

bartcubbins:
Was tested on maple device with 47.1.A.8.49 firmware. Pstore is OK. Any regressions.
@kholk your opinion?:)

PR #1502, opened 2017-11-06
[57r1] MSM8956 VIDC cleanup, new tables, low latency implementation

kholk:


kholk:
PR is valid, @Bitti09 tested on Tone, all working.

My Suzu test is fine. MERGEME.

PR #1509, opened 2017-11-12
Revert "PORT: msm: camera2_v2: Initialize buf_time to boottime"

oshmoun:
This commit causes video recording to break after a certain period of
time, because the implementation is not complete.
Revert it for now, until the complete implementation is ready.

This reverts commit 6498522f9bfb58602f205030cd8f03371295e186.

Change-Id: Ic28be6d3203d1a42ddd7af6ffd1ee0190a7ca7b8


alviteri:
Not revert I will push cam hal commits later

oshmoun:
@kholk says it also requires further kernel changes, so it won't work quite well even with the hal commits

alviteri:
We will see

kholk:
We will get this feature when it will be time to.
MERGEME NOW. CRITICAL FIX.

PR #1514, opened 2017-11-14
arm: DT: Reduce the calibration on somc-poplar-id9

stellirin:
The stock calibration is noticeably yellow. However
this calibration changed this to noticeably blue.
Split the difference for a happy medium.

These values are simply the midpoint between the old
and new calibration, rounded to the nearest 0x80 as
per all other values.

With this change, (my) poplar has a panel calibration
that closely matches other calibrated devices in the
SODP project, instead of being noticeably blue.


kholk:
This change is screwing up all the whitepoints which were measured with proper equipment, hence it's wrong.

Please use the Extended Settings app to switch your panel whitepoint if you're not happy with the default calibration.

kholk:
For your reference, this is the basic color output from the Poplar ID9 display using the default 6000K calibration:

Raw R:15512 G:17945 B:18485 C:50871
Reduced Raw R:60 G:70 B:72 C:198 IR:2
IR:535 CRATIO:0.01 Sat:65535 Sat75:65535
CPL:29.61 Max lux:737.69
Compensated R:14977 G:17410 B:17950 C:50336
Reduced Compensated R:76 G:88 B:91 C:196
Lux:387.57 CT:5957.30K

stellirin:
Whatever the readings taken on **your** device, they simply do not match reality on **my** device. If there's no common ground to be found on this then simply revert to stock calibration and let users deal with it.

Telling others to change to another setting in Extended Settings is laughable, making the problem in fact worse.

kholk:
_My_ sensing device has got a good calibration and by "good", I mean that it has a tolerance of +/- 50K in the CT readings.
I don't know what do you mean by _reality_: you didn't show me that my calibration device is giving me wrong values, infact, you're doing this commit based on what _your eyes_ see, which is not indicating a precise value, as everyone's eyes will react differently to the white color temperatures depending on various factors including, but not limited to, the current ambient lighting.

Telling people to personalize their devices to make it fit to their needs through a given feature that we have working in Extended Settings is the only thing that makes sense, since we are giving them multiple choices for the white point.

Screwing up all the calibrations to reflect an unknown white point value, which people will believe that it is the one that they can read on ExtendedSettings is not laughable at all: that would be a severe issue, since people would be misleaded by fake temperature settings.

stellirin:
I don't doubt your ability to calibrate your own devices. The issue is that your calibration is not compatible with other devices.

BTW I'm not looking at the poplar in isolation, I have it compared to several devices. I see, `white|white|white|blue`. If it was correct I would see `white|white|white|white`.

I'm not talking about insignificant `white|slightly-yellow-white|slightly-red-white|slightly-blue-white`. I'm talking about BLUE. It's crazily blue. Enough so that you can immediately see it even on the android logo at boot. It's a major distraction.

This change causes the colour to be insignificantly different from the rest. I'm sorry that you don't like that I didn't use a tool to read the values. You already did that. Your values don't work for this device. My values work for this device since they are average of the pre/post calibation values.

Please tell me now if there can be no compromise on this topic so I can stop working on this device.

kholk:
I have checked what you say already, my Poplar is not showing the same differences as you say, which makes me think that we are experiencing an issue with either my specific unit or yours.

erikcas:

PR #1515, opened 2017-11-14
[57r1] Add z3fold compressed pages allocator

artem:
This driver is extracted from Yoshino copyleft and is being used at least on Tone too.


jerpelea:
zRAM is not the best way forward

PR #1518, opened 2017-11-16
[57r1] input: misc: fpc1145: Implement miscdevice, IOCTLs, PM, wakelocks, redo regulators

kholk:
This commit implements a miscdevice and IOCTLs to replace the
current ugly sysfs handling.
The sysfs implementation will be removed once the userspace
will be migrated to the new IOCTLs management.

-------------------------------------
TEST: Maple OK
All devices will have no reason to produce badnesses, since the old sysfs implementation is still there and fully functional.
The legacy implementation will be removed once the userspace will be fully converted to IOCTLs.

### DIFF
--- a/drivers/input/misc/fpc1145_platform.c
+++ b/drivers/input/misc/fpc1145_platform.c
@@ -513,7 +513,7 @@ static int fpc1145_probe(struct platform_device *pdev)
 	if (rc)
 		goto exit;
 
-	irqf = IRQF_TRIGGER_RISING | IRQF_ONESHOT;
+	irqf = IRQF_TRIGGER_HIGH | IRQF_ONESHOT;
 	mutex_init(&fpc1145->lock);
 
 	wake_lock_init(&fpc1145->wakelock, WAKE_LOCK_SUSPEND, "fpc_wake");
--- a/drivers/input/misc/fpc1145_platform.c
+++ b/drivers/input/misc/fpc1145_platform.c
@@ -8,9 +8,6 @@
  * clocks, controlling GPIOs such as SPI chip select, sensor reset line, sensor
  * IRQ line, MISO and MOSI lines.
  *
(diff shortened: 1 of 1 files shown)


kholk:
PR updated with deprecation of sysfs, interrupt handler fixes, partial refactoring of the regulators handling, introduction of wakelocks for interrupt conditions to ensure TZ availability, device power management.

P.S.: Now REQUIRES HAL upgrades.

kholk:
Forcepushed: Fixed the regulator fix (in the end I went your way @oshmoun)

Ready to merge.

oshmoun:
some (now useless) explanation:
The failure to get the vcc_spi regulator was not "real". The devm_regulator_get call indeed succeeded.
The issue was with the error checking:
https://github.com/sonyxperiadev/kernel/pull/1518/commits/3a7989e387125629ca8cb22b5be86ad4a6c98a7e#diff-61d40d626ce922f06be403d177809b3bR445
This was seemingly being optimized away, as the compiler detected that this always evaluates to true, for some reason that is related to the fact that the number of iterations in the for loop exceeded the number of elements in the vref_conf array.
Now, replacing that error check with something else causes the compiler to not optimize it away, and leads to having the expected segfault due to the second iteration going out of the bounds of array.
This is the codeblock that, when added as a replacement for the error checking block mentioned above, will lead to the expected segfault:
```
		if (fpc1145->vreg[i] > 0) {
			dev_err(dev, "got %s regulator, address 0x%p.\n", vreg_conf[i].name, fpc1145->vreg[i]);
			continue;
		} else {
			dev_err(dev, "CRITICAL: Cannot get %s regulator.\n",
				vreg_conf[i].name);
			return -EINVAL;
		}
```
Of course, one also needs to remove the ifndef from the enum fpc_rails_t, in order to achieve the mismatch that used to happen previously.

PR #1529, opened 2017-12-08
[TESTINGNEEDED] fbdev: msm: somc_panel: Use B sequence as poweroff reset sequence

oshmoun:
The stock panel driver uses the B sequence as a normal poweroff reset sequence, if found.
https://github.com/sonyxperiadev/kernel-copyleft/blob/41.2.A.0.xxx/drivers/video/msm/mdss/mdss_dsi_panel_driver.c#L1109-L1121
This is not the case in our driver, and leads to issues with dt2w on hybrid incell panels.
The reason for that is that the devicetree nodes of hybrid incell panels only declare the B sequences (somc,pw-off-rst-b-seq).
The missing normal poweroff reset sequence causes the driver to always report a fatal error upon turning off the panel:
[  967.464295] FATAL: poweroff reset sequence is NULL!!

The solution is to add a check to make sure that there's always a poweroff reset sequence set, by falling back to the B
sequence. This makes the driver behaviour similar to that on stock.

This needs testing on all devices that have hybrid incell panels!
If you want to report your tests, please include the panel id along with the device name!
Currently tested on:
kugo with panel id 8 (works fine)

### DIFF
--- a/drivers/video/fbdev/msm/somc_panel/panel_driver.c
+++ b/drivers/video/fbdev/msm/somc_panel/panel_driver.c
@@ -3266,6 +3266,9 @@ int mdss_panel_parse_dt(struct device_node *np,
 		"somc,pw-off-rst-b-seq",
 		(u32 **)&spec_pdata->off_seq.rst_b_seq,
 		&spec_pdata->off_seq.seq_b_num);
+	// if no poweroff reset sequence is declared, but B sequence is, use it
+	if(spec_pdata->off_seq.rst_b_seq && !spec_pdata->off_seq.rst_seq)
+		spec_pdata->off_seq.rst_seq = spec_pdata->off_seq.rst_b_seq;
 	rc = of_property_read_u32(np, "somc,pw-down-period", &tmp);
 	spec_pdata->down_period = !rc ? tmp : 0;
 


kholk:
https://github.com/oshmoun/kernel/blob/d0c510ee2765bd066e37a9a125e4cfb2c41cd4f3/drivers/video/fbdev/msm/somc_panel/panel_driver.c#L370

Your commit is wrong. The reset sequence B is processed already at line 370.
The only wrong thing about the current somc_panel behavior is the pr_err... it shouldn't get printed if we have already executed the B sequence instead of the "standard" one.

To be clear, the difference between the A and B sequences are simply that the standard A sequence is executed AFTER touching (both fixed and not) regulators, while the B sequence is executed PRIOR to that.
The copyleft driver that you've linked is also showing the same behavior.

oshmoun:
I'm not sure how the behavior of somc_panel can be the same as on stock if we have no standard sequence. Let me try to describe in another way:

stock overwrites standard sequence with B sequence if found in dt, as can be seen here:
https://github.com/sonyxperiadev/kernel-copyleft/blob/41.2.A.0.xxx/drivers/video/msm/mdss/mdss_dsi_panel_driver.c#L1113-L1114

somc_panel does not do that, and designates a field for the B sequence, which is of course the better way, but that leaves us with a NULL standard sequence. This means the behavior of the driver is different than that of stock. This PR merely replicates the behavior exhibited by the stock driver.

Now if you're saying that replicating the stock behavior is unnecessary, because processing the B sequence once is enough, then we need to find out why this is fixing dt2w on hybrid incell panels. That has been my observation thus far.

oshmoun:
After checking the stock kernel code more throughly, this PR is indeed wrong. Still unclear how dt2w is fixed with this, if at all. Needs further checking.

PR #1532, opened 2017-12-19
[KERNEL] Port Loire, Tone, Yoshino on new branch LA.UM.6.4.r1

kholk:
Say welcome to the Xperia devices on LA.UM.6.4.r1 - Kernel version 4.4.78.

This is a porting of all 5.7.r1 platforms.

Status:
- SoMC Loire is working and tested. Misses MSM8956 specific Venus/VIDC hacks, TBD.
- SoMC Tone has new configurations, required a bad hack, but should work fine
- SoMC Yoshino is untested. Misses the charger extensions from SoMC and kernel configuration fixes (TBD, use Tone defconfig update as guideline)

Speaking about the bad hack, it's related to regulators: HW constraints do not fit!
This is a serious bug which solution should be only earlier in the bootchain (bootloader!!!).

Due to the lack of time, you will notice a treewide build fix. I'm sorry for that, it's the only way to push the kernel right now.

Enjoy, folks!


bartcubbins:
@kholk good job! Let's see what we can do;)

PR #1534, opened 2017-12-20
[5.7.r1] ANDROID: Sdcardfs: Move gid derivation under flag

oshmoun:
This moves the code to adjust the gid/uid of lower filesystem
files under the mount flag derive_gid.

Signed-off-by: Daniel Rosenberg <drosen@google.com>
Change-Id: I44eaad4ef67c7fcfda3b6ea3502afab94442610c
Bug: 63245673

### DIFF
--- a/fs/sdcardfs/derived_perm.c
+++ b/fs/sdcardfs/derived_perm.c
@@ -176,6 +176,9 @@ void fixup_lower_ownership(struct dentry *dentry, const char *name)
 	gid_t gid = sbi->options.fs_low_gid;
 	struct iattr newattrs;
 
+	if (!sbi->options.gid_derivation)
+		return;
+
 	info = SDCARDFS_I(d_inode(dentry));
 	info_d = info->data;
 	perm = info_d->perm;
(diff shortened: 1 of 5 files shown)


kholk:
Check for functionality break on 8.0!

oshmoun:
will do!

oshmoun:
tested! device still boots! io to internal and external storage looks ok!
if anyone else wants to test, feel free!

kholk:
Oki, then close this pr and merge only the 6.4.r1 one.

bartcubbins:
In my opinion it must be merged

jerpelea:
in my opinion it should be properly tested on both Android 7.1 and 8.0 before we decide

bartcubbins:
already tested on 8.0. any regressions

kholk:
Remember: this is an android 8.1 specific feature. 8.0 will never have it, most probably neither in a security patch. Having  it in a system where it is unused and never managed in any way *may* possibly introduce security  issues.

humberos:
IMHO I think it will be easier to manage changes and their impacts if we have branches for android releases here too.
I mean like devices have n-mr0, n-mr1, o-mr0, ...
but who am I to define things? btw it is just an idea ;)
and yes, using branches in kernel like devices maybe solve this specific situation but will demands more work for replication for sure.
so this is what I could say about this situation.

jerpelea:
please note that:
- the 5.7 branch is used for 8.0
- the 6.4 branch is used for 8.1
I think that we should keep each version separate

PR #1539, opened 2018-01-09
[6.4.r1] Update temperature threshold limits for EA for 8996

artem:
Current limits are causing frequent wakeups at lower temperature.
Update higher and lower limits such that msm-core energy-aware driver
will not get too many notifications to avoid unnecessary wakeups.

### DIFF
--- a/arch/arm/boot/dts/qcom/msm8956.dtsi
+++ b/arch/arm/boot/dts/qcom/msm8956.dtsi
@@ -1240,8 +1240,8 @@
 	qcom,msm-core@a0000 {
 		compatible = "qcom,apss-core-ea";
 		reg = <0xa0000 0x1000>;
-		qcom,low-hyst-temp = <10>;
-		qcom,high-hyst-temp = <5>;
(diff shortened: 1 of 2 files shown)


kholk:
Smells like burnt hands......

artem:
MSM8998 has the same change: https://source.codeaurora.org/quic/la/kernel/msm-4.4/commit/arch/arm/boot/dts/qcom/msm8998.dtsi?id=260a1b12fb91aefd706dbbffedaa805fc15e5f53
For MSM8996 it is present on 3.18 but is missing on 4.4: https://source.codeaurora.org/quic/la/kernel/msm-3.18/commit/arch/arm/boot/dts/qcom/msm8996.dtsi?id=3e7ca625780f36a22d9a1b2b6ec31bbd80737236

kholk:
Ok then!

oshmoun:
Can you please also apply the same for msm8956?

jerpelea:
please link the original caf commit for each platform

PR #1542, opened 2018-01-11
[6.4.r1] Enable VP9 Decoder and VIDC clock scaling on MSM8956

kholk:
This patchset enables Venus on MSM8956.

Needs testing.


oshmoun:
can't seem to notice a functional difference concerning venus, since youtube, twitch and mxplayer seem to work fine without this pr. any error in dmesg to look out for?

kholk:
I recall it not being working but ok!
This will at least implement Venus clock/voltage scaling functionality and VP9 HWdec support, so it's safe to merge.

PR #1545, opened 2018-01-13
[6.4.r1] arm: defconfig: kugo: Change CONFIG_HZ value

tomgus1:
This is to be the same as for bullhead

Discussed with @kholk 
This won't have as big an effect on kugo as it is 720p this will mean that 120Hz functionality cannot be used but due to 720p this isn't supported anyway.


danielgusvt:
This is not a dt change but a defconfig change!

tomgus1:
Sorry, my bad. Didn't sleep much last night.

jerpelea:
changes are applied / platform

jerpelea:
in sync for all platforms

kholk:
This makes sense only on Kugo, since HZ=1000 is useful only for 1080p and platforms that can achieve 120Hz refresh rate.
Kugo's panel does NOT support 120Hz.

jerpelea:
then all platforms should stay on ths same config since there is a small benefit

kholk:
Oki, then close it

PR #1546, opened 2018-01-13
[TESTINGNEEDED] drivers: bcmdhd: update wl_set_key_mgmt

oshmoun:
Apply changes to this function to make it similar to the 4.4 version of the bcmdhd driver.
This resolves the random disconnects that would happen due to failing to set mfp.

Change-Id: Id2891fdd536bb9345fc42e8d5d20e09b7bee1efb


oshmoun:
@notz could I ask you to test if this resolves the random wifi issues you had reported before? At least on my device that seems to be the case

kholk:
The change looks like being sane, from a code sanity perspective.

PR #1547, opened 2018-01-14
[6.4.r1] iommu: qcom: Implement context bank HUPCF and Context Fault CFG attrs

kholk:
Some users (for example, Qualcomm's camera_v2 driver) will take
actions that will expectedly generate context faults: in this
case, the driver may request to disable the Hit Under Previous
Context Fault (HUPCF) bit, so that the IOMMU will process subsequent
transactions independently from an outstanding context fault,
instead of stalling (or terminating) them.

Implement handling of DOMAIN_ATTR_CB_STALL_DISABLE: if this
attribute is set, set the HUPCF bit to 1 on the newly set up
context bank to avoid cb stall on fault.


jerpelea:
it looks good but needs more testing

PR #1548, opened 2018-01-19
[6.4.r1] Fix display and let Tone and Yoshino platform boot

kholk:
Remove the cont-splash-memory from the framebuffer nodes to
avoid IOMMU related crashes at display open from userspace.
This is    necessary after    a sane QC fix on mdss_mdp_splash_logo
appeared, as commit hash 245031160bff81703cf109e473329c3f01d77304:
previously, the    PA to VA mapping was completely    broken but now,
being it sane, we need the related IOMMU domain    to be set as   
UNSECURE and correctly released    by the bootloader.

This is a HACK and will have the side effect of displaying  
corruption for one second prior the Android bootanimation to
kick in.

### DIFF
--- a/arch/arm/boot/dts/qcom/msm8996-tone-common.dtsi
+++ b/arch/arm/boot/dts/qcom/msm8996-tone-common.dtsi
@@ -3144,6 +3144,14 @@
 	status = "disabled";
 };
 
+&mdss_fb0 {
+	/delete-node/ qcom,cont-splash-memory;
+};
+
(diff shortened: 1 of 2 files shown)


oshmoun:
This partially solves tone's boot issue described in https://github.com/sonyxperiadev/bug_tracker/issues/68
Basically, now there is no more graphical corruption prohibiting the device from functioning properly, but verification still randomly fails, causing device to shutdown.

PR #1549, opened 2018-01-22
[LA.UM.6.4.r1] Merge tag 'LA.UM.6.4.r1-05900-8x98.0'

tomgus1:
Merged latest CAF tag into SODP Kernel


danielgusvt:
Needs testing of course!

humberos:
FPC fixed on kagura (by using this tag) so as I told before the fix is on kernel space.
ping @oshmoun

tomgus1:
This is fixed on Loire (suzu) also.

humberos:
yeah,... I can confirm that!

humberos (approved):
Although I don't know what's the proper way merging tags but we really need this tag. No regressions for kagura and suzu.

humberos:
~~I did 20 reboots manually (on kagura) without the verity problem. So I think may we have another embedded fix.~~

NVM,... still broken.

humberos:
FPC is working on suzu, but the workaround (mentioned in bugtracker) is not working anymore.
We will lost FPC after reboot cycle or if we use the workaround^.

> 01-22 00:37:28.913  5562  5562 E AOSP FPC HAL: Error loading user database: -6
> 01-22 00:37:28.913  5562  5562 E android.hardware.biometrics.fingerprint@2.1-service: An unknown error returned from fingerprint vendor library: -6
> 
which means that FPC HAL needs some love :(

Btw Kagura is ok by using the workaround.

oshmoun:
Woah nice find! 
This proves me wrong about fpc then 😅
@humberos did you flash O ftf on your kagura? I.e does your kagura have O keymaster firmware?

jerpelea:
camera  ok

PR #1559, opened 2018-01-30
defconfig: update yoshino defconfig

alviteri:
https://cateee.net/lkddb/web-lkddb/ARM64_SW_TTBR0_PAN.html
https://cateee.net/lkddb/web-lkddb/SETEND_EMULATION.html
https://cateee.net/lkddb/web-lkddb/CP15_BARRIER_EMULATION.html

following wahoo defconfig also this avoid some apps crashes like ATM Milano app SimCity game app


jerpelea:
you found a nice bug in you some apps that use legacy instructions
I like the way you documented the patches the issue but I am not shure that a bulk patch is the best way forward. 
I think that there should be reverts on all platforms for each config change and in each revert you should document the reason for the revert like you specified here.
this will help us identify the patch later when we migrate to a new kernel

alviteri:
Actually those are enabled on loire and tone as defconfig:update defconfig without any kind of document so the thing is let these app still crash on yoshino only or merge this anyway that is the way I found for now with the limited free time

jerpelea:
you did a good job and we have to document it properly.

kholk:
Actually SETEND and CP15 are 99.9% always enabled or disabled in bulk, as they provide (as you know) compatibility with old ARM instructions.
The only split that I would do is the SW Privileged Access Never. The reason why this one is enabled on Loire and not on Tone/Yoshino is that from what I knew (I may have been wrong about this.... Or not!) there is HW (arch) support for that on Kryo equipped SoCs.

alviteri:
Angelo tone has these flags as I can see also wahoo check pixel defconfig

kholk:
Yeah I'm not saying to not merge, it's the opposite really.
Just if you can split the SW PAN config from the other two that'd be great.

jerpelea:
@alviteri  is right .
Good work !

PR #1561, opened 2018-02-03
[6.4.r1] Incremental upgrades to Linux 4.4.114

kholk:
This patchset contains incremental upgrades to Linux 4.4.114.
The upgrades have been applied by using incremental kernel patches found on kernel.org.

Tested on SoMC Suzu, Maple
BOOT OK
Apparently no stability issues.

To be _obviously_ carefully tested on all platforms.


humberos (approved):
No regressions here (kagura)

oshmoun:
kugo looks fine here.
Just a small nitpick, why are 101 and 102 squashed together?
EDIT: just checked, the change is too minimal, so that is the reason I reckon.
A small sidenote: .115 is out

kholk:
Because 102 consists in one commit that was already pre-applied to the kernel, hence i didn't feel like doing another commit only for a version update in makefile... That's it

kholk:
About 115, just... Ignore it for now, let's finish testing and merge this BIG PR as is, then, we will see... A small increment should be very easy to test, likely less than an hour of testing should be enough.

oshmoun:
was able to crash kugo with a 3dmark vulkan run
that was possible previously too, and pstore would show messages about kgsl errors. This time, however, it seems to show errors about swap
https://hastebin.com/yonevayage.vbs

humberos:
@oshmoun 
btw, I've pushed a patch (merged) that got vulkan disabled for loire.

oshmoun:
@humberos yup, saw it
for some reason I can still use vulkan here, and it also was working fine on stock, so I doubt that msm8956 really is incapable of running vulkan.
Maybe the patch only disallowed the system from using certain vulkan features?
Anyway, besides this type of crashes, kugo seems stable :D

kholk:
MSM8956 supports vulkan and should normally be able to use those APIs just fine. This, however, is off topic for this PR.

bartcubbins:
@kholk no regressions on my suzu.
good job! I think we must test it during the week

PR #1563, opened 2018-02-08
[loire] fix mp3 playback on internal speakers

oshmoun:
This PR fixes a couple issues on loire devices that break audio playback using offload on internal speakers.
1. The commit https://github.com/sonyxperiadev/kernel/commit/da60e6ac20006f2e5e350b0a75066ed5c2a461a4 removed a dt node required to register some ctls. This node is re-added.
2. The commit https://github.com/sonyxperiadev/kernel/commit/7357058c09a8a923f6698cf045e9ab5c677bb32b reserved MultiMedia16 for NOIRQ, which isn't supported on loire. An ifdef for loire is added there to use the older configuration.

### DIFF
--- a/arch/arm/boot/dts/qcom/msm8956.dtsi
+++ b/arch/arm/boot/dts/qcom/msm8956.dtsi
@@ -2652,8 +2652,12 @@
 		compatible = "qcom,msm-pcm-lpa";
 	};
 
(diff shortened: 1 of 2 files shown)


jerpelea:
@kholk

PR #1564, opened 2018-02-12
Address Castor USB OTG issues

aolexiy:
- CONFIG_SONY_USB_EXTENSIONS can't be supported by platform, disabled for Castor only to avoid testing on other devices
- MHL driver build fixes for the case when CONFIG_SONY_USB_EXTENSIONS is disabled (just stubbed issues, this can be done in a different manner)
- dwc3_otg_set_power() skipping call to power_supply_set_supply_type() fixed

These should address USB OTG issue on Castor, when you can only connect the USB device to the phone just one time.


aolexiy:
Jeez, wrong branch

PR #1570, opened 2018-02-18
[6.4.r1] rqbalance: Fix CPU resources lockdown (and some lagging)

kholk:
This patchset aims to address some lagging issues in some rather rare conditions on some devices, less rare on some others. Also, some cleanup has been done.

When the decider sets the RQBalance state to downcore, we may
still have a very CPU intensive process that is locking CPUFreq
to maximum frequency: this will make our frequency-based
algorithm kickstarter to just do nothing, while the userspace
may be spawning lots of other processes, resulting in a complete
hog of the current CPU available resources without being unable
to unlock more.

Of course, directly running a full evaluation right after that
will just waste CPU cycles and resources, so also introduce a
new worker delay variable "recheck_delay" that wants to be "an
high amount of time", preset to 2*down_delay: this will make us
able to wait enough time so that, if we have a single thread
that is very CPU intensive, we give it time to finish its job
without putting even more load on our system.

So:
If we want to forcefully downcore, and
- We haven't onlined all cores
- Our CPU is intensively used (perhaps by just one thread);
Then, run the worker again after recheck_delay millseconds
and prefer doing an upcore full evaluation to eventually
unlock more resources.

This PR appears to be similar to #1454, but:
1. Fixes the performance regressions introduced by that one
2. Fixes the power regressions
3. Decides based to the real current system load, instead of blindly temporarily disabling the downcore decision

### DIFF
--- a/drivers/cpuquiet/governors/rqbalance.c
+++ b/drivers/cpuquiet/governors/rqbalance.c
@@ -57,6 +57,7 @@
 /* Don't put any CPU down prior at least 3x scheduling samplings */
 #define DNCORE_DELAY_MS		(3 * (RQ_SAMPLE_TIME_NS / NSEC_PER_MSEC))
 #define UPCORE_DELAY_MS		((RQ_SAMPLE_TIME_NS / NSEC_PER_MSEC) / 2)
+#define RECHECK_DELAY_MS	(6 * (RQ_SAMPLE_TIME_NS / NSEC_PER_MSEC))
 
 typedef enum {
 	/* Brings one more CPU core on-line */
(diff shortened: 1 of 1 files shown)


kholk:
Please test on ESPECIALLY Tone.
I won't cry if anyone tests on Loire, too :)

Tested on Maple.

humberos:
Awesome!
My kagura just got a new life!
Good job!

humberos (approved):
It is ok on kagura.

humberos (approved):
Suzu is also ok!

jerpelea:
all 3 platforms are ok

kholk:
Merge!

djselbeck:
After some days of testing I can say, that with this patch applied lacking got back (on kagura). Especially when playing embedded videos in firefox after a while the device locks on one online and three offline CPUs. Like the old behaviour. I think it took longer to happen.

- Kagura
- Running on battery
- Android 8.1

Playing video in firefox:
$ uname -a
Linux localhost 4.4.114-g967945836df8-dirty #1 SMP PREEMPT Sun Feb 18 10:25:43 CET 2018 aarch64
$ cat /sys/devices/system/cpu/offline                                    
1-3

PR #1573, opened 2018-02-19
[LA.UM.6.4.r1] loire: Configure TLMM register space as strongly ordered

humberos:
The MSM8953 register setting stuff is valid for Loire as well:
use it to solve CPUSS related unstabilities.


humberos:
ping @kholk

kholk:
Well, in reality I missed that on purpose.
Also, I don't see any CPUSS unstability on 6.4.r1 :D

oshmoun:
the remaining causes of instability, so far observed, are utilization of the vulkan api, and having an sdcard inserted. So, newbie question:
Would this affect either of those?

kholk:
Fairly sure it wouldn't, as this is affecting the CPUSS block (so addresses hotplugging and instabilities related to cpu sleep).
Though, well, trying won't kill you so... Just... Try it and tell me I'm right! :p

PR #1575, opened 2018-02-21
[6.4.r1] [TONE] Fix USB OTG, disable bcmdhd logspam

kholk:
Tested on SoMC Kagura RoW.


artem:
Is Dora neccesary in 6504f2bffb561f6f0a31a7be0e16ac2abc214a20?
And since keyaki uses the same hw as kagura, I think that f0afee97980dc41381e2ca81c2c8fa8ddb86e049 should be applied to it too

kholk:
I know what I did... :)

Check the code and get yourself a reply :)))

PR #1579, opened 2018-02-26
[Staging] LA.UM.6.4.r1-06900-8x98.0

humberos:
New CAF's tag


oshmoun:
boots on both kugo and keyaki
no regressions discovered so far

kholk:
Obi, try to make a phone call. The SSC VMID change looks suspicious.

oshmoun:
good catch, especially for loire!
tested on kugo and keyaki (earpiece, speaker, bt), and all works fine.

kholk:
Then I'm 99% sure that the weektest will succeed :)))

humberos:
No issues here (kagura and suzu).
Not sure about yoshino.
Could someone report it ?

PR #1580, opened 2018-02-26
Loire bt pinctrl: fix incall and recording usecases

oshmoun:
This PR fixes two issues that prohibited the usage of a BT headset for normal calling or audio recording.
First issue is the missing function declarations in DT nodes related to AUX PCM.
Second issue is the incorrect assignment of the pins 123 and 134 to wrong functions that don't seem to exist on loire.
Adding the correct function declarations in DT, and assigning the pins 123 and 134 the correct functions in pinctrl driver makes the aforementioned usecases work.


kholk:
MERGE! :D

PR #1581, opened 2018-03-02
[6.4.r1] Upgrade to Linux 4.4.120

bartcubbins:
source kernel.org
https://cdn.kernel.org/pub/linux/kernel/v4.x/incr/


bartcubbins:
tested on suzu

kholk:
The irq-gic-v3 change is a bit odd. There the smp_wmb is exchanged with wmb.... In a place where the comment says that we need to be sure of the visibility to other CPUs.

Please guys, test Tone and Yoshino, I have bad suspects about that one.

PR #1582, opened 2018-03-02
[6.4.r1] somc_panel: Remove legacy panel support

bartcubbins:
There is no more reason to support legacy display panel driver.
Tulip had a chance to get 4.4 kernel, but because a4xx does not have an updated firmware, it's impossible.


bartcubbins:
ping @kholk

bartcubbins:
ping @kholk

kholk:
I agree, but... let's just wait a little more. Probably the best idea will be to remove that in the next kernel porting.

pablomh:
As sad as I am to accept it, I think it's the way to go :(

kholk:
The legacy panel driver is still needed.
Please, close this PR.

PR #1588, opened 2018-03-06
[6.4.r1] Add unfortunate hack to solve Loire sensor wakelock issues

kholk:
Add a very poor hack that converts the sensor_ind wakelock to
a timed one (with a 5 seconds timeout) on-the-fly.
This is unfortunately necessary, as it is the only solution for
a battery drainage issue, for broken DSP firmwares on old
SoCs combined with proprietary sensor binaries which aren't
accounting for that (anymore).
​
This change will somewhat retain the functionality that these
sensor binaries are aiming to give on a system featuring an
Android userspace.

Of course, this hack gets enabled with CONFIG_PM_WAKELOCKS_SENSOR_IND_HACK.
In this patchset, it is enabled for all currently supported Loire devices.

P.S.: This code is fresh and it is CURRENTLY UNDERGOING TESTING. Allow at least 2 days prior merging.


oshmoun:
the 5 seconds timeout might indeed be too much.
after about 3 hours of uptime, here are some stats about sensor_ind:
count: 753 times
total time: 253ms
average wakelock time per instance: 0.34ms
it seems to be a really short wakelock, usually.
might i suggest a 20ms timeout? it's the same used for the irq handler in fpc, so that waking up the device through movement or through fpc have the same incurred wakeup time.

kholk:
If you want to relate it to the fpc, then you generally don't want the same amount of time.... But since it doesn't look like being related to that one, at all, at this point it seems just better to allow just 5 to 10ms.

But.... Now that you've seen that..... At this point, let me just search a thing. In the meanwhile, go on with testing this PR :)))

humberos (commented):
Apart from decision about timeout increasing/decreasing but this patch is really necessary to solve battery drains (bring back deep sleep state). 
Good job!

humberos (approved):

oshmoun:
can confirm that the mitigation works
count: 508
total time: 5172
so one instance of the wakelock got stuck, but was terminated due to the timeout

jerpelea:
I would preffer a relaxed time out since we don't really know what triggers the  wakelock

PR #1590, opened 2018-03-10
drivers:gpu:msm: adjust drawobj timeout and memstore size for less performant platforms

oshmoun:
platforms such as msm8956, sdm630 and sdm660 need more relaxed values to avoid gpu hangs and possibly other general instabilities.
Taken from kernel-copyleft 50.1.A.3.xxx:
https://github.com/sonyxperiadev/kernel-copyleft/blame/50.1.A.3.xxx/drivers/gpu/msm

Change-Id: I6516fa6b0a972690e600341e7c89cdfa9588cbe1


kholk:
MERGEME!!!!!!!!

PR #1594, opened 2018-03-14
[6.4.r1] arm64: configs: Yoshino: Enable wlan packets offload

kholk:
This gives a good 18-20% throughput increase on wifi DL and
at the same time also sensibly reduces CPU load.

Tested on the new QCA CLD tag LA.UM.6.4.r1-06900-8x98.0
Tested on SoMC Yoshino Lilac


alviteri:
angelo enable this on nile

PR #1597, opened 2018-03-16
[64r1] Enhance Yoshino camera user experience

kholk:
We have a nice frequency of 384MHz that we can use for certain
resolutions and refresh rates on our cameras.
This frequency ensures we are not wasting battery unnecessarily,
reducing heat and that we have enough performance for a snappy
user experience.

Also, solve an user experience issue on Lilac front camera.


knightazura:
Will this improvement released on stock ROM in the next update?

jerpelea:
Stock and Open Devices are 2 separate projects with different code and functionality

PR #1598, opened 2018-03-17
Enable fcp1145 driver for nile devices

EnJens:


kholk:
Jens, please, arm64: configs: Nile: Enable (blah)

PR #1602, opened 2018-03-24
[LA.UM.6.4.r1] Merge tag 'LA.UM.6.4.r1-07600-8x98.0' 

tomgus1:
Merged latest CAF tag into SODP Kernel


kholk:
Half of the commits in there are REALLY interesting. Let's give this a good test, should bring some good improvements!

oshmoun:
especially those concerning kgsl!

kholk:
Pinched! :p

oshmoun:
https://github.com/sonyxperiadev/kernel/pull/1602/commits/507f372acebdbfa2105b0241dbf939c03f231ec1
this commit conflicts with the linux upstream merges. specifically with https://patchwork.kernel.org/patch/10218657/

Both commits seem to have the same goal, so the question is whether to overwrite the linux upstream update with a caf change, or ignore the caf update. caf over linux or linux over caf?

oshmoun:
so, to get this merge working, the following commits can be reverted:
https://github.com/sonyxperiadev/kernel/commit/507f372acebdbfa2105b0241dbf939c03f231ec1
https://github.com/sonyxperiadev/kernel/pull/1602/commits/97f857f556d5a141ea947326ca9f40f4821533ef
The first one has been explained in the above comment.
The second one breaks boot on loire and tone

jerpelea:
kernel.org has priority

tomgus1:
Reverted both commits

kholk:
No no no no no no no no don't revert the second one. Check if it works on the other platforms and, if it does, then add a build barrier for loire on top.
If that's the case, on the next kernel we will find the same commit and we will apply the same barrier.

tomgus1:
Okay, only first reverted. Can others check the remaining platforms?

kholk:
Thanks!

tomgus1:
Sooo, without the revert for 97f857f as @oshmoun stated, boot on Loire is broken with Magisk installed as the boot fails verity.

oshmoun:
@tomgus1 pretty sure magisk is unrelated. Device will crash due to failing verification either way.

tomgus1:
@oshmoun The device booted okay without the revert but if I was to install magisk the device would fail verification. Applying the revert allows me to pass verification if I installed magisk.

tomgus1:
So, further testing concludes that Reverting 97f857f is required 100% on Loire. Yes it can boot but activating WiFi causes the device to crash and reboot.

oshmoun:
the commit can then probably be ifndeffed with CONFIG_BCMDHD

tomgus1:
Done, builds perfectly now

tomgus1:
Loire and Tone both confirmed to be working no issues. Can someone with Yoshino and Nile test please?

PR #1607, opened 2018-03-29
[6.4.r1] tty: serial: Correct IPC_LOGGING ifdef

bartcubbins:
ipc_log_context_create() is function used to create a ipc log context.
All code that uses this function must be placed under IPC_LOGGING ifdef.
In our case this function is placed before ifdef. Hence, fix this.


jerpelea:
commit message please

bartcubbins:
done

PR #1614, opened 2018-04-06
[5.7.r1] cpuidle: lpm-levels: ifdef non-PSCI code

bartcubbins:
There is missing ifdef for non-PSCI code which used only on
MSM8916 class SoCs. This code should be compiled only for
non-PSCI SoCs.


bartcubbins:
I'm afraid about loire, so I test this PR on it, then I let you know

PR #1624, opened 2018-04-14
[DONOTMERGE]input: misc: fpc1145: remove all wake logic

oshmoun:
The wakelock is not needed anymore with the new hidl hal implementation.
Tested additionally with fingerprint gestures enabled, which is known to cause device crashes without the wakelock, and there were no crashes observed.

Change-Id: Ibacff4b4afb7d167191da5aef9b0b219d3a87e18

### DIFF
--- a/drivers/input/misc/fpc1145_platform.c
+++ b/drivers/input/misc/fpc1145_platform.c
@@ -47,7 +47,6 @@
 #include <linux/of.h>
 #include <linux/of_gpio.h>
 #include <linux/uaccess.h>
-#include <linux/wakelock.h>
 #include <linux/regulator/consumer.h>
 #include <linux/platform_device.h>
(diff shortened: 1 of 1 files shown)


oshmoun:
testing needed on all platforms for extended periods of time with fingerprint authentication enabled.

jerpelea:
@BuzzBumbleBee @EnJens

EnJens:
@oshmoun the irq wake removal seems a bit too much? Won't that prevent the fpc from waking up the device from sleep?

oshmoun:
according to tests done here on kugo, that irq wake has no effect. device is unable to wake from sleep with and without it.

jerpelea:
irq wake is tested on yoshino and confirmed working

PR #1625, opened 2018-04-18
[6.4] upstream commits for clang compatibility

Myself5:


kholk:
Most of the commits in there are total crap.

There are some good ones that I would really like to have (even for gcc, actually), some are partially good, some are to refactor and some are totally unacceptable. 

Will eventually talk about them in private, as this is gonna be a veeeeeeery long conversation :p
... But I think it's worth it to PROPERLY fix all the possible commits so that people will be able to build with clang.

nathanchance:
I'd be very curious to see which commits you think are "crap" because 75% are directly from mainline. The ones labelled "hack" really aren't THAT bad and in fact, most are fixed with an upgraded version of LLVM/Clang.

kholk (changes requested):
First, a patch saying that "When %a0 is changed to [%x0] it uncovered a GCC bug"...
https://github.com/sonyxperiadev/kernel/pull/1625/commits/a00c7fa2c28e357aeb3536b72c71fd1aa9685ede

...then a patch that doesn't care about it.

Really?

kholk (approved):

kholk (approved):

kholk (changes requested):
I seriously hope that this patch is an april fools.

kholk (changes requested):
Some versions of GCC, depending on their configuration at compile time, will throw a warning for each file when an unknown warning option is set.

C'mon guys.....

kholk (changes requested):
I don't wanna say bs here, but aren't we using pci/probe.c functions also for PCIe?

....Because seriously, PCIe can use 64-bit addressing as well, what is going on here?

kholk (approved):

kholk:
Oh, I HATE when github doesn't point to the commits where you approve or suggest changes :)))

kholk:
Let's write them here for reference....

REVIEWS:
Rx (WHAT) --- R = Review - x= Review No. -- WHAT = APPROVE/REJECT
R1 (REJECT): https://github.com/sonyxperiadev/kernel/pull/1625/commits/cddd996fd5922d7adad3c76855b61fa28e2204a9
R2 (APPROVE): https://github.com/sonyxperiadev/kernel/pull/1625/commits/f1a01e44dc750201e2c472292286ebbe20b904cf
R3 (APPROVE): https://github.com/sonyxperiadev/kernel/pull/1625/commits/ed96715827fd1d03c891f3e55a894372a3bc0f5a
R4 (crying REJECT): https://github.com/sonyxperiadev/kernel/pull/1625/commits/d667937353f693ae485465594314c82e96f1fe19
R5: (REJECT): https://github.com/sonyxperiadev/kernel/pull/1625/commits/ca1a8f63a256ea9dd75414afbbabb6cb333aaf98
R6 (REJECT): https://github.com/sonyxperiadev/kernel/pull/1625/commits/4ae772227131b4c1b0bb7d36e2ddfaba029c6fa3


Sorry if any confusion was created.
There is a lot more to be said for this PR, but at this point I think I'll just cherry-pick the really good patches and then we'll go on for the rest.

nathanchance:
R1: So only support LLVM 6.0+ and you don't need it.

R4: Yeah, not exactly pretty but the whole idea is it is improving undefined behavior (https://youtu.be/6l4DtR5exwo?t=27m56s). It also got added to stable as https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable.git/commit/?id=1b94a87cf9888ad9675b65138a33608abd8fa067 in 4.4.127 so if it was really as bad as you say, why did it get added?

R5: I'm not sure what the issue is with that patch, again, just bringing the two compilers in line to do the same thing.

R6: I guess the PCI maintainer is an idiot for accepting that patch upstream lol. It will be in 4.4.127 as https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable.git/commit/?id=01f4db3cd85ad0dc9fdd838f276abd33427e107f

If you really think a lot of these are a problem, why not submit the "proper" fixes to mainline, since anything with a `BACKPORT`, `UPSTREAM`, or `FROMLIST` suffix came from there? Do as you see fit, it is your project but this work is going into mainline so you'll see it in your kernels eventually.

erikcas:

PR #1634, opened 2018-04-26
[6.4.r1] Fix GCC 7.3.0 warnings [1/2]

nathanchance:
Part 2 is in [qcacld-3.0](https://github.com/sonyxperiadev/vendor-qcom-opensource-wlan-qcacld-3.0/pulls)

1. Move from gcc-wrapper to a pure `-Werror`, removing any Python dependency.

2. Fixes all GCC 7.3.0 warnings. This has been build tested on all `aosp_*` defconfigs with both [AOSP's GCC 4.9.4](https://android.googlesource.com/platform/prebuilts/gcc/linux-x86/aarch64/aarch64-linux-android-4.9/) and [Bootlin's GCC 7.3.0](https://toolchains.bootlin.com). The warnings rage from stylistic errors to "wow that is actually broken" (like https://github.com/nathanchance/kernel/commit/fb247dd0b7fcab63a44af78651b39e5a820f8975 for example). See the commit messages for further details.

3. Remove the trailing space in the compiler version (looks better in Settings -> About phone).

NOTE: It is entirely possible the device with fail to boot without https://github.com/nathanchance/op5/commit/7c848abaf621cf181db335729e7339fcb83f5aec; if so, please let me know so I can add it to the patch set.


kholk:
Good job Nathan!

Only two "objections": first, please don't modify that kconfig commit by setting it default y. While I agree to the fact that it's sane to actually enable it, the best idea is to update all the devices configurations to have it =y instead of setting it as default. 
It's true that this kernel is basically being used only for MSM, but still, let's stay in line with mainline style....

And second objection, about that tautological warning on BCMDHD, I'd like to examine it more deeply and understand why that variable is normally undefined. I trust your judgment about just putting an ifdef, but since that driver is seriously messed up and has its other issues, 4 eyes are better than one. 

In the meanwhile, can you please update your PR with the removal of the last commit? (enable cc_werror globally) 


Thanks!

nathanchance:
Done! I figured it would be sane to enable it globally and made more sense to just do it via the default rather than enabling it in every single defconfig but I have no strong opinion either way!

As for the tautological warning, my guess is maybe the bcmdhd driver had something like qcacld-3.0's option to build out of tree at some point in time, where that was a configurable option. The history of the driver makes that a little hard to uncover...

Here are the other occurrences of BCMBUSTYPE:
```bash
nathan@flashbox ~/kernels/others/xperia/kernel (gcc7-warnings) $ ag -s BCMBUSTYPE
drivers/net/wireless/bcmdhd/siutils.c:722:#if !defined(BCMBUSTYPE) || (BCMBUSTYPE == SI_BUS)
drivers/net/wireless/bcmdhd/siutils.c:724:#endif	/* !BCMBUSTYPE || (BCMBUSTYPE == SI_BUS) */
drivers/net/wireless/bcmdhd/siutils.c:727:#if !defined(BCMBUSTYPE) || (BCMBUSTYPE == SI_BUS)
drivers/net/wireless/bcmdhd/siutils.c:729:#endif	/* !BCMBUSTYPE || (BCMBUSTYPE == SI_BUS) */
drivers/net/wireless/bcmdhd/include/bcmdefs.h:114:#ifdef BCMBUSTYPE
drivers/net/wireless/bcmdhd/include/bcmdefs.h:115:#define BUSTYPE(bus) 	(BCMBUSTYPE)
```

jerpelea:
good work!

PR #1645, opened 2018-05-05
Misc updates

EyeCantCU:


drivers/block/zram/zram_drv.c
```diff
@@ -1021,7 +1021,7 @@ static void zram_reset_device(struct zram *zram)
 	/* Reset stats */
 	memset(&zram->stats, 0, sizeof(zram->stats));
 	zram->disksize = 0;
-	zram->max_comp_streams = 1;
+	zram->max_comp_streams = CONFIG_NR_CPUS/2;
```
artem: It's a dirty hack that relies on defconfig value

mm/vmscan.c
```diff
@@ -146,7 +146,7 @@ struct scan_control {
 /*
  * From 0 .. 100.  Higher means more swappy.
  */
-int vm_swappiness = 60;
+int vm_swappiness = 15;
```
artem: I don't think it plays well with our 3GB RAM devices. Anyway, changing default value directly in kernel used by range of different devices is not a very wise decision

fs/sync.c
```diff
@@ -18,7 +18,7 @@
 #include <linux/backing-dev.h>
 #include "internal.h"
 
-bool fsync_enabled = true;
+bool fsync_enabled = false;
```
artem: We need to make sure that memory chips on all supported devices don't corrupt data with fsync disabled

nathanchance:
* eaeb83d968e25a0180de08aa2ce9c90a972ddd17 was rejected upstream in 2011 (see the long discussion [here](https://lkml.org/lkml/2011/5/11/228)), I disagree with carrying it forward, especially since there's no evidence that it is relevant anymore.

* ea7c7e08aedecfc1223ef7c4e593eb143ec2f498 did not make it into mainline either (although I can't find a solid thread for why it was rejected) so it should be taken with absolute scrutiny. For what it's worth, none of my devices have had issues with entropy being deleted randomly, I suspect the same will be said here.

* As pointed out by @ab123321, both 16a692a76b593e0085ed13facd1e766ccb7362c3 and a699f8d6ac0b5edec9c82dafcd47186c16f3465c should be done in init scripts echoing the desired value to sysfs. Hardcoding changed values into the kernel is a bad idea.

* I personally would edit the commit messages of 2446f94effec22cb45cf9e60bc699c3be109738f and 9d29b198ddd92da6d4d69cd0e6c91973c389a053 to note that they both came from mainline (0f376c9cd86c23f37312d37748b233660ef9d9af and d0563a039c9d35305f4cb2405665130de596d86a respectively).

* c39a01007dcf73cee0a5f0afbba1ee4e18f07062 seems worrisome as there aren't just MMC devices supported here, 835 devices are UFS SCSI devices. If it can be proven disabling fsync won't have disastrous consequences for those devices, then I suppose it makes sense to disable it.

EyeCantCU:
Thanks @nathanchance and @ab123321. I'll adjust these accordingly

EyeCantCU:
I addressed all of the issues with the previous patchset. Are there any more concerns? I'm going to look into disabling fsync on UFS devices. I wasn't thinking of those at the time. But I'd rather do that as a separate pull request as I just remembered another interesting commit that could coincide it

nathanchance:
Another thing I forgot to mention...

https://github.com/andip71/boeffla-kernel-oos-oneplus5/commit/5c605cc252cf53e0194b45345b4e50a78e640d01

https://github.com/andip71/boeffla-kernel-oos-oneplus5/commit/dc5d5d0eebe1b9522f8f4f012906baa5184156b9

Boeffla claimed [on Twitter](https://twitter.com/andip71/status/975640051777458177) that 8ac39bb1d2f74b5d435c982c00a50e65b97665c1 can cause data corruption on the OnePlus 5. No idea if that is the case here and I never took the time to see if he was right because I never added the initial commit but just in case you weren't aware of it.

EyeCantCU:
@nathanchance I'll investigate further. That doesn't sound right to me. I've used that commit on the OnePlus X & 2, in addition to Berkeley (the Honor V10) now for a long while without that happening. I'll look into it though and report back

EyeCantCU:
We can accomplish the same thing, if proven useful, by mounting the filesystem with the noatime option in fstab. I'll get rid of that commit. Better safe than sorry

bartcubbins:
Also about https://github.com/sonyxperiadev/kernel/pull/1645/commits/a78fdf3315875ae54d2c7903b0935b26c9768d37 commit.. Please update commit message. Original commit: https://android.googlesource.com/kernel/msm/+/f4f91b946ce2af74d5fb3c8938a328cfec11d69e

EyeCantCU:
Both commits are by Tim Murray respectively.... Nevermind. Thought I saw something about authorship. I just woke up

bartcubbins:
This commit was Signed-of by Franco but there are no changes from him.

nathanchance:
@bartcubbins what's the problem with that? It's in accordance with [the developer's certificate of origin](https://developercertificate.org) and almost every patch in this series has an additional signoff.

bartcubbins:
then ok, np

kholk:
I'm sorry, I am trying to be as nice as possible, but... There is no commit in there that is worth to be considered.

Between cache line alignment for some specific SoC, the removal of debugging that would help to solve issues, scheduler behavior changes that will screw up the logic of all the drivers engineered with that hiccup in mind (cpu down counted as idle) and the other stuff....... To me, it looks like a random cherry pick of stuff. 

I'd recommend to the owner of this PR to make some good research about what I just said, to understand what all of this does. 
By the way, in case your research can't come to an exhaustive explaination of what the code that you've pushed actually does, feel free to ask to get any clarification.

EyeCantCU:
@kholk I understand. Half of these are from the mainline Android kernel. If, however you feel as though this is useless you probably have more experience than I do in this field.

PR #1646, opened 2018-05-06
[6.4.r1] Make BCL power mitigation less harsh for MSM8996

artem:
When battery level hits qcom,soc-low-threshold BCL hotplugs out
range of CPUs and limits max CPU frequency to qcom,mitigation-freq-khz.
Unfortunately, this makes our MSM8996 devices hardly usable in low
battery state.
Disable BCL hotplugging and increase frequency cap to ~1.1 GHz

Signed-off-by: Artem Labazov <123321artyom@gmail.com>

### DIFF
--- a/arch/arm/boot/dts/qcom/msm8996.dtsi
+++ b/arch/arm/boot/dts/qcom/msm8996.dtsi
@@ -2933,12 +2933,10 @@
 		qcom,bcl-enable;
 		qcom,bcl-framework-interface;
 		qcom,bcl-freq-control-list = <&CPU2 &CPU3>;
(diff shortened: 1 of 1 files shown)


jerpelea:
@kholk

jerpelea:
-1 it will deplete faster a battery that is already almost empty and we may endup with an emergency situation without a way to do it

artem:
@jerpelea We have RQBalance, so it isn't the case for us. HAL should hotplug needed cores off.
As for freq limit, keeping it super low for two cores is overkill

jerpelea:
even with only 2 cores we can save some power

PR #1648, opened 2018-05-11
[6.4.r1] Forwardport msm8996 device tree patches from msm-3.18 kernel

artem:
This patchset contains forwardports of several device tree commits that are missing in 4.4 kernel and includes some OEM-specific updates from Sony open source copyleft releases


jerpelea:
@kholk

PR #1653, opened 2018-05-14
Revert CONFIG_RANDOMIZE_BASE for tone and loire

oshmoun:
This config makes tone and loire device unbootable.


jerpelea:
lets see why it crashes before we decide to revert

oshmoun:
i had assumed it's because the bootloader doesn't support to randomized base, but not really sure bootloader has anything to do with this feature

jerpelea:
not the best thing to do but......a way forward

PR #1655, opened 2018-05-16
fbdev: msm: somc_panel: correct machine barrier to not allow incompat…

oshmoun:
…ible devices

The logic of the if barrier was allowing all devices in except those that weren't compatible with all three given values.
AND the seperate statements instead of ORing to receive the needed result.


oshmoun:
@kholk

kholk:
Yeah, of course, bad logic mistake!

Merge!

PR #1656, opened 2018-05-16
[6.4.r1] treewide: Linux 4.4.132

nathanchance:
source kernel.org
https://cdn.kernel.org/pub/linux/kernel/v4.x/incr/patch-4.4.131-132.xz


pablomh:
Running on pioneer (nile platform) since you PRd it with no apparent regressions.

PR #1657, opened 2018-05-18
[LA.UM.6.4.r1] Merge tag 'LA.UM.6.4.r1-08500-8x98.0'

tomgus1:


nathanchance:
You have conflicts in `drivers/android/binder.c` and I don't think the resolution of `net/packet/af_packet.c` is correct.

See https://github.com/android-linux-stable/msm-4.4/commit/eac7fd6c1ad300e548b3beda0818c8b8ffca4a74#diff-e936270d5d8fefaf66cb5ded38a638fc

tomgus1:
Yeah, I'll look at those. Should really do this at a PC and not on phone.

nathanchance:
Just a heads up, build log with `aosp_loire_blanc_defconfig`: https://del.dog/raw/ogokogujak

net/packet/af_packet.c
```diff
@@ -1713,19 +1713,17 @@ static int fanout_add(struct sock *sk, u16 id, u16 type_flags)
 	    match->prot_hook.type == po->prot_hook.type &&
 	    match->prot_hook.dev == po->prot_hook.dev) {
 		err = -ENOSPC;
-		if (atomic_read(&match->sk_ref) < PACKET_FANOUT_MAX) {
+		if (refcount_read(&match->sk_ref) < PACKET_FANOUT_MAX) {
 			__dev_remove_pack(&po->prot_hook);
 			po->fanout = match;
-			po->rollover = rollover;
-			rollover = NULL;
-			atomic_inc(&match->sk_ref);
```
nathanchance: The two statements above this line should stay to match the diff of the patch: 171588e9ad56705b9bd1b1af80c0467063937d0d

This also matched mainline: https://github.com/torvalds/linux/blob/master/net/packet/af_packet.c#L1752-L1760

drivers/android/binder.c
```diff
@@ -1432,6 +1432,26 @@ static void binder_send_failed_reply(struct binder_transaction *t,
 	}
 }
 
+/**
+ * binder_cleanup_transaction() - cleans up undelivered transaction
+ * @t:		transaction that needs to be cleaned up
+ * @reason:	reason the transaction wasn't delivered
+ * @error_code:	error to return to caller (if synchronous call)
+ */
+static void binder_cleanup_transaction(struct binder_transaction *t,
```
nathanchance: This causes an unused function warning because it's not used unlike in the original patch: f743ad82a49afae201ae8b24530c5832d979af4e

I guess you guys switched over to the Pixel 2's Oreo binder stack even though Google is moving back to kernel/common in P: https://github.com/aosp-mirror/kernel_msm/commit/29d196199717c5b49a236f2f0ced0f25bb0d03ed

net/packet/af_packet.c
tomgus1: Fixed this one now

nathanchance (approved):
All conflicts look proper now, just needs on device testing!

tomgus1:
Been running for the past few weeks on Suzu with no regressions.

PR #1659, opened 2018-05-26
[6.4.r1] treewide: Linux 4.4.133

nathanchance:
source kernel.org
https://cdn.kernel.org/pub/linux/kernel/v4.x/incr/patch-4.4.132-133.xz


pablomh:
Working fine with no apparent regressions on pioneer (nile platform) since 5 days ago.

PR #1660, opened 2018-06-02
[6.4.r1] Linux 4.4.134 and 4.4.135

nathanchance:
4.4.134 announcement: https://lwn.net/Articles/755964/

4.4.134 announcement: https://lwn.net/Articles/756052/


pablomh:
It's been running with not apparent regressions on pioneer (nile platform) since ~6 days ago.

PR #1664, opened 2018-06-11
defconfig: nile/yoshino: enable CONFIG_HZ_1000 instead of CONFIG_HZ_100

pablomh:
This way all the supported devices will run with CONFIG_HZ_1000.


pablomh:
I will be running this diff on pioneer (nile platform) and will report any regression, but some yoshino tests will be needed as well.

bartcubbins:
I think it's better to separate it into two commits, one for Nile and another one for Yoshino

oshmoun:
maybe also completely switch to NO_HZ
seems possible judging by the defconfig of op5(t)
https://github.com/omnirom/android_kernel_oneplus_msm8998/blob/android-8.1/arch/arm64/configs/omni_oneplus5_defconfig#L82

kholk (approved):

kholk:
I approve, but separate the commits!
Make one for all Yoshino and one for all Nile!

nathanchance:
@oshmoun It's not full NO_HZ though: `# CONFIG_NO_HZ_FULL is not set`

 You still specify a HZ value: https://github.com/omnirom/android_kernel_oneplus_msm8998/blob/android-8.1/arch/arm64/configs/omni_oneplus5_defconfig#L422-L426

`CONFIG_NO_HZ` is already set in these configs.

oshmoun:
ah thanks, missed that.
Then not completely, but still partially might be worth investigating, i.e with CONFIG_NO_HZ_IDLE

alviteri:
https://android.googlesource.com/kernel/msm/+/android-msm-wahoo-4.4-oreo-m4/arch/arm64/configs/wahoo_defconfig#65

Shouldn't be better keep wahoo value ?

nathanchance:
@alviteri I've been using `CONFIG_HZ=1000` for a few months now with no regressions on the Pixel 2 XL. Google stated they have been experimenting with it and found positive results: https://source.android.com/devices/tech/debug/jank_jitter#long_threads

pablomh:
I'll separate them in a following PR.

pablomh:
I've created https://github.com/sonyxperiadev/kernel/pull/1665 and https://github.com/sonyxperiadev/kernel/pull/1666 to that effect.

PR #1665, opened 2018-06-12
defconfig: yoshino: enable CONFIG_HZ_1000 instead of CONFIG_HZ_100

pablomh:
This way yoshino platform will run with CONFIG_HZ_1000 as well.


kholk:
Merge!

PR #1667, opened 2018-06-18
tone: dts: convert i2c_3 gpios to pinctrl

erikcas:
see https://github.com/sonyxperiadev/bug_tracker/issues/137


erikcas:
tested on kagura. Not tested on keyaki, afaik CASH is not used on keyaki atm (?)

PR #1671, opened 2018-07-04
[6.4.r1] treewide: Linux 4.4.139

nathanchance:
https://cdn.kernel.org/pub/linux/kernel/v4.x/incr/patch-4.4.138-139.xz


tomgus1:
Not experienced any regressions on Suzu.

PR #1672, opened 2018-07-05
[6.4.r1] Backport clearpad changes from k4.9

artem:


kholk:
Thanks for cherry-picking.
Can anyone quickly test this before merging, please?

Thanks again!

oshmoun:
lost touch input on kugo after turning screen on! turning screen off and on again fixed it.
https://hastebin.com/amiromulas.sql
~touch completely breaks after a while, with the only way to get it back being a reboot~
it gets much harder to revive touch after a while

erikcas:

jerpelea:
I would prefer to freeze the stable kernel and add only kernel security patches

PR #1675, opened 2018-07-14
[6.4.r1] [URGENT] rqbalance: Schedule CFL setting on hotplug if cluster unavailable

kholk:
If a cluster is totally down, it is impossible to set a cpufreq
policy on it.

Change the logic so that, if the cluster is currently up we set
the limits right in the sysfs handler but, if the cluster is
unavailable, schedule the setting for the next upcore event and
artificially produce a non-blocking upcore event on the
navailable cluster: this will make us get into the newly
introduced hotplugging notifier that will take care of updating
the policy (thus setting the frequency limits) as soon as the
cluster will come up.

### DIFF
--- a/drivers/cpuquiet/governors/rqbalance.c
+++ b/drivers/cpuquiet/governors/rqbalance.c
@@ -120,7 +120,6 @@ static unsigned long recheck_delay;
 static unsigned long last_change_time;
 static unsigned int  load_sample_rate = 20; /* msec */
 static struct cpu_limits rqb_freq_limits[MAX_CLUSTERS];
-static enum cpuhp_state rqb_hp_online;
 static struct workqueue_struct *rqbalance_wq;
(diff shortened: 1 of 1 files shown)


kholk:
Build errors due to k4.4 having the old hotplug notifiers. Reopening when backported correctly.

kholk:
The PR has been fixed.

MERGEME!

alviteri:
the main problem was fixed but now i can observe 

07-16 00:14:07.953   698   698 E RQBalance-PowerHAL: Error writing to /sys/devices/system/cpu/cpuquiet/nr_min_cpus: Invalid argument

kholk:
This is not an issue and it is absolutely not related to the RQB-CFL patchset.
Whenever this happens, the PowerHAL will retry and succeed. Don't worry.

PR #1683, opened 2018-07-21
[2.3.2.r1.4] Linux 4.9.113

nathanchance:
Build tested on all Nile, Tama, and Tone configs, no warnings or errors.

Some good arm64, ext4, and mm fixes here.


kholk:
Thanks for the upgrade, but unfortunately this will wait some ... time.

The 4.9 branch is not fully stable yet and it's not a good idea to merge upstream patches until all devices are totally usable.
Don't close the PR though, even though it'll take quite some time, I'm sure that we will start upgrading, sooner or later!

kholk:
Ah, last thing.

I agree about merges being correct like this, but can you do it like I started doing it on k4.4 (treewide: Linux xxxx)?
It's only for porting purposes... otherwise we'll have to dig through hundreds of commits to find our board specific ones.

Thanks!

nathanchance:
I agree about not shipping this right away but since we're only ten revisions behind upstream, I figured it is important a

nathanchance:
I agree about not shipping this right away but since we're only ten revisions behind upstream, it's important to start testing because the delta is only going to grow.

Couldn't you just hide all of the stable commits by using `git log --no-merges LE.UM.2.3.2.r1.4..HEAD ^v4.9.113` when forward porting? That's a big advantage to merging (getting full history but being able to filter as you need).

kholk:
Yes, but this will get complicated over time.... the 4.9 kernel is not EOL and there will be more updates which will make the task difficult.
I've had experiences and they weren't the best.

nathanchance:
I guess I am not sure I understand what you mean by complicated or difficult, it's rather easy to filter stuff down with git. For me personally, applying the patch files has been more cumbersome than doing merges. But this is your project. I want to put this on a new branch anyways so I'll resubmit.

kholk:
GIT is not perfect, Nathan... btw, thanks.
I appreciate that.

PR #1686, opened 2018-07-21
clk: cpu-8976: Fix refcount for A72ss MUX clock

oshmoun:
The refcount for the A72 MUX can reach zero when the CPUs are
down: this way the API will also shut down the APC1 regulator
and the A72 cluster will go totally down and save power.


kholk:
Thanks for the backport.

MERGEME

PR #1687, opened 2018-07-21
arm64: DT: MSM8996: Use broadcast timer for aggregating sleep mode

Paulbouchara:
The CPU Power Collapse mode usually triggers L2/CCI (system)
Power Collapse (PC).
The system PC activates a low power module, which relies on
the external timer (also called "broadcast timer") to be
able to aggregate the next wakeup in a cluster: in this
case the CPU also switches to use the broadcast timer instead
of the architected timer hardware.

Without using the external timer, the system will deadlock,
as the architected timer hw will reset during power collapse.


kholk:
Thanks, MERGEME

PR #1689, opened 2018-07-21
[2.3.2.r1.4] Remove GCC wrapper and add CONFIG_CC_WERROR

nathanchance:
The GCC wrapper adds an unnecessary dependency on Python when regular `-Werror` will do. Add CONFIG_CC_WERROR to prevent warnings from creeping in.


kholk:
I totally agree with this one

PR #1698, opened 2018-07-26
[2.3.2.r1.4] enable WQ_POWER_EFFICIENT_DEFAULT

tomgus1:
Enabling this config has given the experience of a more responsive experience with a cooler device. 

Tested on LA.UM.6.4.r1 on suzu and kugo


kholk:
Power efficient workqueues are marked as unbound.
When a workqueue is unbound, it can run on any CPU in the system, leaving the target selection to the scheduler (so the task gets scheduled on the least loaded _already online_ CPU).

The change is valid and efficient on k4.4 and k4.9 implementations.
MERGEME!

tomgus1:
I need to add tama
edit: all good to go

PR #1707, opened 2018-07-31
Revert "usb: gadget: Update default class subclass and protocol for r…

aolexiy:
…ndis"

Swithching RNDIS interface descriptor parameters breaks
tethering for Linux hosts since rndis_host driver expects
USB_CLASS_WIRELESS_CONTROLLER/1/3. It's possible to add
a line into init script writing 1 into [1] at the boot time
but it seems a better idea to kick this out since the original
patch obviosly wasn't tested enough.

[1] /config/usb_gadget/g1/functions/rndis_bam.rndis/wceis

This reverts commit b90d1eb5b8e2f5d76e025bdc212fa1fd13212af7.

### DIFF
--- a/drivers/usb/gadget/function/f_qc_rndis.c
+++ b/drivers/usb/gadget/function/f_qc_rndis.c
@@ -106,7 +106,6 @@ struct f_rndis_qc {
 	u8				port_num;
 	u16				cdc_filter;
 	bool				net_ready_trigger;
-	bool				use_wceis;
 };
 
 static struct ipa_usb_init_params rndis_ipa_params;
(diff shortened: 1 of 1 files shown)


aolexiy:
@jerpelea - this should address tethering on 4.4 based builds, IMHO it's https://github.com/sonyxperiadev/bug_tracker/issues/24

kholk:
Please check that this revert is not breaking devices featuring the DWC3 controller and possibly different IPA versions before merging.

In case they break, do the revert with a build barrier to "solve" the issue. 

Good job!

jerpelea:
Please test tone and yoshino platforms !

aolexiy:
yoshino needs two more reverts so I'll add a two liner instead...

PR #1708, opened 2018-07-31
[2.3.2.r1.4] Switch zRAM backend on all devices to z3fold

artem:
Requires #1680


artem:
Do not merge until #1721 is tested

artem:
631ab7cf5051d2ab87b82ecb2792f96880b2d2d8

PR #1709, opened 2018-07-31
[LA.UM.6.4.r1] usb: gadget: enable wireless RNDIS by default

aolexiy:
Use default parameters suitable for both Windows and Linux hosts.

### DIFF
--- a/drivers/usb/gadget/function/f_gsi.c
+++ b/drivers/usb/gadget/function/f_gsi.c
@@ -2929,6 +2929,8 @@ static struct f_gsi *gsi_function_init(enum ipa_usb_teth_prot prot_id)
 		goto error;
 	}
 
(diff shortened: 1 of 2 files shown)


aolexiy:
@jerpelea, @kholk - instead of reverting 3 patches this sets defaults to use wireless class to allow to use tethering on both windows and linux hosts (current code doesn't work on linux). Checked on yoshino@poplar, tone@dora and nile@pioneer on Ubuntu 16.04 and Windows hosts.

Loire@suzu requires more fixes - at least dwc3 calls should be stubbed in u_data_ipa.c, maybe more.

pablomh:
Hi @aolexiy, would you mind ammending the title to include the branch?

aolexiy:
@pablomh ugh, right, thanks for the tip

PR #1714, opened 2018-08-02
Android.mk: add SOMC_KERNEL_VERSION

alviteri:
projects using both kernel versions as 4.4 and 4.9 fails to build since there is not a rule for it


alviteri:
needed by projects using inline kernel headers build system

PR #1734, opened 2018-08-11
[2.3.2.r1.4] defconfig: loire: disable CONFIG_MSM_ISPIF_V1

oshmoun:
Loire has v2 of ispif, not v1.
Disabling the config makes the driver select the correct register header.


kholk:
MERGEME!!!!!

PR #1739, opened 2018-08-14
[6.4.r1] Android.mk: Replace obsolete ANDROID_BUILD_TOP

bartcubbins:
Since Android 9 (Pie) ANDROID_BUILD_TOP variable marked as obsolete.
So, use $(shell pwd) to return the current working directory.

This change completely compatible with Android 8 (Oreo).

See:
https://android.googlesource.com/platform/build/+/07699636b0f65c0ca756f883ca6d8ccac88fbe11
https://android.googlesource.com/platform/build/+/HEAD/Changes.md


MarijnS95:
This change seems to break dependencies on kernel headers. It exports the [`KERNEL_HEADER_INSTALL` target](https://github.com/sonyxperiadev/kernel/blob/40063c548ce7767e208c8498afda287d64f94f9a/Android.mk#L211) as `$(shell pwd)$(TARGET_OUT_INTERMEDIATES)/KERNEL_OBJ/usr`, whereas other projects depend on `$(TARGET_OUT_INTERMEDIATES)/KERNEL_OBJ/usr`, causing the following error:
```
ninja: error: 'out/target/product/suzu/obj/KERNEL_OBJ/usr', needed by 'out/target/product/suzu/obj/EXECUTABLES/android.hardware.biometrics.fingerprint@2.1-service.sony_intermediates/BiometricsFingerprint.o', missing and no known rule to make it
```

I have only been able to compare this to an omnirom 8.1 tree, where it works because `OUT_DIR` is not equal to `out`, thus taking the [false case](https://github.com/sonyxperiadev/kernel/blob/40063c548ce7767e208c8498afda287d64f94f9a/Android.mk#L28) with a relative path. They have [a patch applied](https://gerrit.omnirom.org/c/android_build/+/1058) to force the path to be absolute, so I do not know how this compares to previous Sony AOSP builds.

I solved this locally using an [ugly change](https://github.com/MarijnS95/kernel/commit/fd0cd250ba134ce16ed4152bcb3adbbe7511376c), but have also considered migrating to [`AndroidKernel.mk`](https://android.googlesource.com/kernel/msm/+/android-9.0.0_r0.1/AndroidKernel.mk) from AOSP.

bartcubbins:
Thank you for your reply, I will recheck this

bartcubbins:
PR is on the way, need to do some tests first

bartcubbins:
updated, build ok

alviteri:
@MarijnS95 @bartcubbins what happened here ? 

https://github.com/omnirom/android_build/blob/android-8.1/core/envsetup.mk#L307

Can we revert this https://github.com/sonyxperiadev/kernel/pull/1739/commits/85eb7effab6ebfa676e87f3ed2875a84db673449 on both branches ?

alviteri:
Also 9.0 uses same https://github.com/omnirom/android_build/commit/1c0b52fdbce5a9e7025a2d42efb44bffd8676a6f can we revert ?

PR #1740, opened 2018-08-14
[2.3.2.r1] Android.mk: Fix kernel headers dependencies

bartcubbins:
Correct the kernel headers path for compatibility with other
projects which depend on them.


bartcubbins:
Suzu build ok on both O and P

PR #1748, opened 2018-08-20
arm64: DT: MSM8956: Declare wakegic as interrupt controller

oshmoun:
Add the missing property "interrupt-controller" to wakegic node.
Without the property, the kernel reports the following error:
[    0.000000] OF: of_irq_init: children remain, but no parents
It subsequently skips the wakegic node, which causes mpm to route irqs to QGIC2 without processing.
As a side effect, sys_pm_ops is not registered by mpm in the lpm driver.
That then leads the lpm driver to skip over any lpm level that notifies rpm.
Avoid this chain of events by correcting the wakegic node declaration.

### DIFF
--- a/arch/arm64/boot/dts/qcom/msm8956.dtsi
+++ b/arch/arm64/boot/dts/qcom/msm8956.dtsi
@@ -490,6 +490,7 @@
 		clock-names = "xo";
 		qcom,num-mpm-irqs = <96>;
 		qcom,ipc-bit-offset = <1>;
+		interrupt-controller;
 		interrupt-parent = <&intc>;
 		#interrupt-cells = <3>;
 	};


oshmoun:
ping @kholk

kholk:
No, no... The description is incorrect.
The mpm gets initialized but routes all the interrupts to the QGIC2 instead of processing them and sending if not relevant to the modem power manager interrupt controller. 

Fix the description, please! 

The commit is valid. I'm sorry for the bad bad bad bad miss (though I recall I did that already, I forgot yet another patch, as always....).

oshmoun:
But mpm is not initialised, at least not according to the kernel
https://github.com/sonyxperiadev/kernel/blob/aosp/LE.UM.2.3.2.r1.4/drivers/irqchip/qcom/mpm.c#L667
this init function is definitely not called, far as I checked

kholk:
Oh yes it is.
If the mpm GIC does not initialize, being the primary interrupt controller the kernel wouldn't be able to boot (it would refuse) because there is no way for the Linux kernel to work without any interrupt support, being it basic functionality for the scheduler! :D

kholk:
The description is now correct.
Merge! :D

PR #1750, opened 2018-08-24
Revert "Android.mk: Fix kernel headers dependencies"

alviteri:
This reverts commit 3eb5f107da0d29ba0c68ced7ec2ea2f7153750e3.


alviteri:
important custom roms as lineage and omni which build with inline kernel use same patch and not cause conflicts during build see
https://github.com/LineageOS/android_build/blob/lineage-16.0/core/envsetup.mk#L374
https://github.com/omnirom/android_build/blob/android-9.0/core/envsetup.mk#L374

so it needs respect KERNEL_OUT conditional here https://github.com/sonyxperiadev/kernel/blob/aosp/LE.UM.2.3.2.r1.4/Android.mk#L25

PR #1752, opened 2018-08-24
Add TTL support

russel5:
Share unlimited internet for free


jerpelea:
please follow the project layout
- submit the change on all platforms when you submit a config change 
- the commit message must explain why it is needed

russel5:
As i'm write this is needed because some internet providers check ttl value and if it's change we can't share our unlimited internet via wifi to other devices
This commit add support to "freeze" ttl value and share internet

jerpelea:
this comment has to be added to the commit message

PR #1755, opened 2018-08-24
Add TTL support

russel5:
Some internet providers check TTL value and if it's change we can't share our unlimited internet via wifi to other devices.
This commit add support to "freeze" TTL value and share internet


jerpelea:
Please follow the layout!
There must be 1 separate commit for each platform for easy porting and revert

russel5:
Sry but i dont understand what are you want

bartcubbins:
1 commit for all loire devices
1 commit for all yoshino devices
etc

bartcubbins:
also don't forget about commit description

jerpelea:
here is a nice example 
https://github.com/sonyxperiadev/kernel/commits/aosp/LA.UM.6.4.r1/arch/arm64/configs

bartcubbins:
something like this:
arm64: configs: Loire: Enable CONFIG_IP_NF_TARGET_TTL

Some internet providers check TTL value and if it's change we can't share our unlimited internet via wifi to other devices.
This commit add support to "freeze" TTL value and share internet

PR #1758, opened 2018-08-25
drivers: mmc: Move zero-length array to end of struct.

MarijnS95:
Zero-length arrays should always be placed at the end of the struct.
If not they use the same memory as the fields that come after it,
resulting in corrupted data.

See #1757


kholk:
MERGEME (4.4+4.9)

PR #1767, opened 2018-09-02
arm64: defconfig: update all defconfig

alviteri:
using update_defconfig.sh from kernel-defconfig repo
updated to commit b0f0c153a75ca5bfc4206a8b87a02df33dfe7cef


jerpelea:
to avoid any possible diverge we will update the aosp configs from time to time after all  comits are tested properly

PR #1770, opened 2018-09-08
techpack: msm8998: tasha: fix yoshino conditional

alviteri:
yoshino should pick right btn_high[1] for tasha not both


kholk:
A=B;
B=C;
C=D;
A=X;

A==B = FALSE
A==X = TRUE

WAKE UP! :P

PR #1773, opened 2018-09-09
msm-audio: dtsi: remove double pcm2: qcom,msm-ultra-low-latency

alviteri:
already defined after

pcm1: qcom,msm-pcm-low-latency

### DIFF
--- a/arch/arm64/boot/dts/qcom/msm-audio.dtsi
+++ b/arch/arm64/boot/dts/qcom/msm-audio.dtsi
@@ -29,13 +29,6 @@
 		compatible = "qcom,msm-compr-dsp";
 	};
 
-	pcm2: qcom,msm-ultra-low-latency {
-		compatible = "qcom,msm-pcm-dsp";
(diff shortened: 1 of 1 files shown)


kholk:
Utterly wrong.
See #1774.

jerpelea:
unfortunately this is wrong because some platforms need pcm2 config

alviteri:
alin your comment seems not ok since it is already defined and defined again ... check the DT ... anyway i still think this is double def

jerpelea:
where is the second definition?

alviteri:
https://github.com/SonyAosp/kernel/blob/aosp/LE.UM.2.3.2.r1.4/arch/arm64/boot/dts/qcom/msm-audio.dtsi#L32
https://github.com/SonyAosp/kernel/blob/aosp/LE.UM.2.3.2.r1.4/arch/arm64/boot/dts/qcom/msm-audio.dtsi#L46

jerpelea:
you are right
please open a PR and add those 2 lines in commit message

PR #1774, opened 2018-09-09
msm-audio: dtsi: remove double pcm2: qcom,msm-ultra-low-latency

alviteri:
already defined after

pcm1: qcom,msm-pcm-low-latency

### DIFF
--- a/arch/arm/boot/dts/qcom/msm-audio.dtsi
+++ b/arch/arm/boot/dts/qcom/msm-audio.dtsi
@@ -29,13 +29,6 @@
 		compatible = "qcom,msm-compr-dsp";
 	};
 
-	pcm2: qcom,msm-ultra-low-latency {
-		compatible = "qcom,msm-pcm-dsp";
-		qcom,msm-pcm-dsp-id = <2>;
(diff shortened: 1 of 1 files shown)


kholk:
Utterly wrong.
There are two ULL channels (not in all SoCs) and the device gets used only when assigned to a route.

alviteri:
Should be declared as well https://github.com/sonyxperiadev/kernel/blob/aosp/LE.UM.2.3.2.r1.4/arch/arm64/boot/dts/qcom/msm-audio.dtsi#L649 for me this is double declared only that ... Also not sure but since 670 uses same 660-common audio look ... It's not declared double https://github.com/sonyxperiadev/kernel/blob/aosp/LE.UM.2.3.2.r1.4/arch/arm64/boot/dts/qcom/msm-audio-lpass.dtsi

alviteri:
After this ... In 4.4 where audio works ... I can play mobile legends under ultra low latency as well (it uses noirq as default where supported)

jerpelea:
unfortunately this is wrong because some platforms need pcm2 config
on the affected platforms you can remove the node !

PR #1775, opened 2018-09-10
techpack: audio: remove SND_SOC_MSM8X16

alviteri:
The SND_SOC_MSM8X16 is not used because it was replaced to SND_SOC_MSM8956


kholk:
That should be used for some config conditionals and maybe some ifdefs in code!!!

jerpelea:
all those conditionals (if any) must be moved to the relevant platform!!!

jerpelea:
I checked and there is not trace to SND_SOC_8X16

PR #1776, opened 2018-09-12
Discovery: enable front flash

MarijnS95:
By pulling in the latest changes to `msm_flash.c` (from 50.1.A.3xxx), front flash now functions correctly.

The commit description might be off as I am not certain what this change exactly adds and how this prevents bootloops when the front flash is enabled (I have not been able to pull logs from the device). I assume something that requires the new functionality in order 

Note that the bodies of `msm_flash_high_brightness_set` and `msm_flash_low_brightness_set` are almost identical. This has intentionally not been refactored to stay in line with "our upstream" (Sony's copyleft kernel).

This change should also be tested and pull-requested for K4.9.

### DIFF
--- a/arch/arm/boot/dts/qcom/sdm630-nile-discovery_camera.dtsi
+++ b/arch/arm/boot/dts/qcom/sdm630-nile-discovery_camera.dtsi
@@ -41,7 +41,7 @@
 		qcom,flash-source = <&pm660l_flash1>;
 		qcom,torch-source = <&pm660l_torch1>;
 		qcom,switch-source = <&pm660l_switch1>;
-		status = "disabled";
+		status = "ok";
 	};
 
 	cam_vdig_imx300_219_vreg: cam_vdig_imx300_219_vreg {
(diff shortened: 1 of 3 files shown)


kholk:
The msm flash driver contains a custom Sony interface that is not used for qcom standard stuff. That's wrong.

jerpelea:
please cleanup the custom interfaces

jerpelea:
did you manage to cleanup the commits?

MarijnS95:
@jerpelea As far as I see it there's nothing to cleanup *yet*. The custom interface is needed, otherwise the device crashes on boot. Either `led_classdev_register` can be removed and flash still works, but remove both and get the aforementioned crash. Further testing reveals that the interfaces are not used (eg. `msm_flash_brightness_get` is called by the kernel to retrieve the initial value, and `msm_flash_*_brightness_set` is never called.
I have yet to find someone with a UART (or buy one myself) to identify what happens. This behavior ~needs to be checked~ happens on 4.9 as well.

jerpelea:
@MarijnS95 please port it to 4.9

PR #1780, opened 2018-09-16
uapi: Define __BITS_PER_LONG based on compiler target

alviteri:
* We may compile 32-bit ARM code against these kernel headers in many
  situations, so provide a compiler-defined method of obtaining the width
  of long.

Change-Id: Iac5e48200d70f1258ab3caca1a8f1eb6e8f7f2d3


alviteri:
needed when build audio.primary.xxxx (for inline kernel build) causing the following error in build

format specifies type 'ssize_t' (aka 'int') but the argument has type 'ssize_t' (aka 'long')

alviteri:
@jerpelea @kholk what about this .. I need this to build audioHAL with inline kernel headers

PR #1781, opened 2018-09-17
arm64: dts: add initial camera support for Akari

sraase:
Disable existing camera nodes and add new nodes for both
camera sensors and flash on Akari.

Since the tama platform boots using a dtb overlay, removing nodes
or properties in the dts no longer works. This makes it impossible
to unset boolean properties (such as rgltr-cntrl-support) at all.


jerpelea:
the same is needed for apollo

PR #1783, opened 2018-09-23
Brcmfmac: add support for bcm43455

oshmoun:
Add basic support for bcm43455 in brcmfmac by pulling in upstream patches.
Also add additional quirks specifically for bcm43455 to avoid crashing and disassociating from AP upon suspending.
Additionally, extend some ifdefs in mmc that only considered bcmdhd to also apply for brcmfmac.


kholk:
MERGEME!!!!!

drivers/mmc/core/sdio.c
```diff
@@ -1090,7 +1090,7 @@ static u32 mmc_select_low_voltage(struct mmc_host *host, u32 ocr)
 {
 	int bit;
 	u32 ocr_orig = ocr;
-#ifdef CONFIG_BCMDHD
+#if defined(CONFIG_BCMDHD) || defined(CONFIG_BRCMFMAC)
```
artem, line 1093: Shouldn't BCMDHD ifdef be removed?

oshmoun: that's intentional, so that a switch back to bcmdhd is still possible

PR #1784, opened 2018-09-23
Loire: switch to brcmfmac instead of bcmdhd

oshmoun:
Apply the necessary dt and defconfig changes to make brcmfmac work fine on loire devices.
This needs https://github.com/sonyxperiadev/kernel/pull/1783 which adds support for the wifi card used in loire devices to the driver.


jerpelea:
@oshmoun 
please add https://github.com/sonyxperiadev/kernel/pull/1784/commits/254f8d3f9e7c52544ea1865baf858b46af5cfe88 in a separate PR

kholk:
@jerpelea
Please note that the gpio switch to active high should be incompatible with bcmdhd, as it is designed to pull it up at probe stage from the Sony wificontrol subdriver. 
If you want the defconfig change as a separated pr for that purpose, then you may rethink about that.... :p

@oshmoun 
Why didn't you remove the unused bcmdhd node?
Switching back shouldn't be an option.

oshmoun:
simple, forgetfulness :)
will revisit this PR hopefully soon, once there's a decision concerning original 4.9 driver vs backports

PR #1790, opened 2018-09-25
[2.3.2.r1.4] [CRITICAL] URGENT: regulator: qpnp-labibb: Shut down machine immediately if VREG_NG

kholk:
On VREG_NG detection, we are sure that there is something wrong
going on.
To try to avoid machine hardware damage, immediately cut power
by shutting it down.


MartinX3:
This pull request will be the workaround against the tama hardware damage?
Good luck. :)

kholk:
That's .... something that can effectively prevent damage.
The issue that actually shorts the DriverIC is still not solved, but it's better to have a better way to prevent a permanent damage in the meanwhile.

PR #1791, opened 2018-09-25
[2.3.2.r1.4] [TAMA] Reconfigure LAB and IBB regulator and QPNP-LABIBB safety measures.

kholk:
The LAB and IBB configuration may differ between devices.
Split it to clean it up and to set it on a per-device basis
to prepare the DTs for OLED panel (requiring LAB/IBB OLED mode)
and for a probe-time preconfiguration of the regulators.

Instead of forcefully relying on somc_panel setting the regulator
parameters on the first display powerup, pre-set the params
so that they get set at labibb driver probe.

Set:
- Max voltage 5.5V (5.6V for Akari only, for LGD panel)
- Init voltage 5.5V (5.6V LGD set by somc_panel on poweron)
- Current limits - LAB: 200mA, IBB: 800mA
- Soft start: 800uS, 300K discharge resistor (inrush prevention)
- Maximum precharge time: 500uS

Also, take safety measures in qpnp-labibb code.

### DIFF
--- a/arch/arm64/boot/dts/qcom/sdm845-tama-akari_common.dtsi
+++ b/arch/arm64/boot/dts/qcom/sdm845-tama-akari_common.dtsi
@@ -225,3 +225,7 @@
 	qcom,fg-sys-term-current = <(-180)>;
 	qcom,fg-chg-term-current = <155>;
 };
+
+&labibb {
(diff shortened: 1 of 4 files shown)


kholk:
PR updated to include additional qpnp-labibb-regulator safety code.

PR #1799, opened 2018-10-12
[LE.UM.2.3.2.r1.4] Update loire and tone config

humberos:
1. Enable and fix MMC_SD_DEFERRED_RESUME.
2. Remove CONFIG_USB_CC_FUSB301 from suzu as it is micro usb compatible with.
3. Also, organize defconfig file based on generator output.


jerpelea:
please update the defconfigs !

humberos:
;)

PR #1819, opened 2018-11-10
[2.3.2.r1.4] Add rules to generate DTBO image

kholk:
This allows to generate a DTBO image from the kernel's Android.mk
for kernel inline building on AOSP: if called, it will by default
generate a dtbo image named "dtbo-arch.img" where "arch" is the
target architecture.
For example, on ARM64, the image will be named "dtbo-arm64.img".

It is possible to override the image name by setting the env
variable called "BOARD_DTBO_IMAGE_NAME" in your BoardConfig.mk:
this will generate a file named "dtbo-myname.img" in the out
folder.

Once image generation is done, it will be copied with the
standard name that the Android build system wants "dtbo.img"
directly into the out folder again.
This flexibility may be removed if in the future the build system
will change to add it by default.

This code is inspired to Christian (myself5) and Luca(luca020400)'s
kernel.mk changes for inline kernel building with dtbo.

----------------------------------------------------------------------------------------------
NOTE:
The DTBO build process is triggered ONLY when the environment variable "TARGET_NEEDS_DTBOIMAGE" is set to "true", to avoid breaking non-dtbo
devices.

### DIFF
--- a/Android.mk
+++ b/Android.mk
@@ -75,6 +75,14 @@ TARGET_PREBUILT_INT_KERNEL := $(KERNEL_OUT)/arch/$(KERNEL_ARCH)/boot/$(TARGET_PR
 KERNEL_DTB := $(KERNEL_OUT)/arch/$(KERNEL_ARCH)/boot/dts/qcom
 KERNEL_DTB_OUT := $(PRODUCT_OUT)/dtbs
 
+ifneq ($(BOARD_DTBO_IMAGE_NAME),)
+  TARGET_PREBUILT_INT_DTBO_NAME := $(BOARD_DTBO_IMAGE_NAME)
+else
+  TARGET_PREBUILT_INT_DTBO_NAME := dtbo-$(KERNEL_ARCH).img
+endif
+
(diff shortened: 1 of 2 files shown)


MartinX3:
Thank you very much, @kholk .

Sorry for my short question:
But is it possible to integrate the dtbo.img into the boot.img on a DTBO Device?
I heard that it is needed by "fastboot boot boot.img", which doesn't work on the tama device.
Or maybe @jerpelea knows it.

jerpelea:
this is not possible because you have 3 partitions that must match boot dtbo and vbmeta

MartinX3:
Even with
`fastboot --disable-verity --disable-verification flash vbmeta vbmeta.img`
?


Would it be possible in the future to use the AOSP kernel with the stock firmware?

kholk:
The Device Tree Overlay is being ... overlayed ... by the bootloader's ufdt, hence it's not possible to do what you want here.
I understand that the fastboot boot command is not working, but that's again bootloader-level stuff.

For more informations, you may definitely want to read the DTO implementation guides at Google:
https://source.android.com/devices/architecture/dto

oshmoun:
~not working here for some reason
`make[1]: *** No rule to make target 'dtbo.img'.  Stop.`~
nevermind, mea culpa

PR #1822, opened 2018-11-11
clk: qcom: hfpll: initialize spinlock before using

oshmoun:
Add a simple init function to initialize spinlock before being utilized in other functions.
How this ever worked is unclear...


kholk:
MERGEME

PR #1827, opened 2018-11-17
[2.3.2.r1.4] fs: sdfat: Register sdFAT filesystem as exFAT

danielgusvt:
Since Android 9 there is native vold support for the exfat
driver [1], which expects the "exfat" driver name instead
of sdfat, so in order to avoid patching the vold source code,
and to respect the AOSP decisions [2], let's register this
filesystem as "exfat".

Signed-off-by: Daniel Vasquez <danielgusvt@yahoo.com>

### DIFF
--- a/fs/sdfat/sdfat.c
+++ b/fs/sdfat/sdfat.c
@@ -4969,7 +4969,7 @@ static void sdfat_debug_kill_sb(struct super_block *sb)
 
 static struct file_system_type sdfat_fs_type = {
 	.owner       = THIS_MODULE,
-	.name        = "sdfat",
+	.name        = "exfat",
 	.mount       = sdfat_fs_mount,
 #ifdef CONFIG_SDFAT_DBG_IOCTL
 	.kill_sb    = sdfat_debug_kill_sb,


danielgusvt:
Ping @oshmoun

kholk:
This is totally wrong.

There is no filesystem called "sdfat" and at this point we can just call it "exfat" instead, without any additional BS configuration option.

jerpelea:
i agree 
we should just use extfat and fix 8.1 patches

kholk:
I'm fine with the new patch.

MERGEME

artem:
@kholk Should this line remain untouched? https://github.com/sonyxperiadev/kernel/pull/1827/commits/50d0e9c544d8344a97c8a9bdde152577973e2f57#diff-b50736185d4342221a46e926ced06401L4991

kholk:
That's fine, that name is used only for debugging paths.
Since the driver name effectively is sdfat, even if this driver named "sdfat" is registering a FS named "exfat" we can leave it as it is.

I personally don't care about that.

PR #1839, opened 2018-12-04
[2.3.2.r1.4] MSM8998 - SDE Enablement

kholk:
This enables MSM8998 to work on the SDE driver with the right QSEED3 configuration,
Display QoS, HDR CSC, Destination Scaler and SDE Rotator.

Tested on SoMC Lilac RoW.


oshmoun:
no issues noticed on tama

PR #1845, opened 2018-12-07
[2.3.2.r1.4] power: supply: qpnp-smbcharger: save irq numbers

aolexiy:
The smbchg_request_irq doesn't save the interrupt number breaking all
the IRQ control logic with in the rest of the module. Change irq_num
argument to a pointer type to fix the issue.


aolexiy:
This fixes at least wakeup issue with suzu@loire when device doesn't wake up from deep sleep when USB cable is inserted.

@jerpelea, @kholk - please have a look

kholk:
Let's say that it's not the best solution.... But it works, so.... Thanks!

Merge!

PR #1846, opened 2018-12-08
[1/2] input: misc: fpc1145: Add poll interface for IO multiplexing in fp HAL

MarijnS95:
This poll interface allows the HAL to multiplex on the fingerprint device and other event sources simultaneously, instead of calling the ioctl that wakes up every .5 seconds to check other events (such as cancellation). See commit descriptions for further details.

Tested on Suzu, android 9.0 and k4.9.
Partially tested on Tama, also android 9.0 and k4.9.
Build-tested in an Android 8.1 tree with k4.9 and k4.4 (the latter which requires commit 9b25a132cc021 except the addition of `#include <linux/pm_wakeup.h>`: https://github.com/sonyxperiadev/kernel/pull/1847).

This needs futher testing on all other platforms before merging!
- [x] Loire
- [x] Tone
- [x] Yoshino
- [x] Nile
- [ ] Tama

### DIFF
--- a/drivers/input/misc/fpc1145_platform.c
+++ b/drivers/input/misc/fpc1145_platform.c
@@ -50,6 +50,7 @@
 #include <linux/pm_wakeup.h>
 #include <linux/regulator/consumer.h>
 #include <linux/platform_device.h>
+#include <linux/poll.h>
 
 #define FPC1145_RESET_LOW_US 1000
 #define FPC1145_RESET_HIGH1_US 100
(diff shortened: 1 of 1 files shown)


jerpelea:
Tested devices:
suzu
dora
maple
pioner
tama FPC is WIP

kholk:
This is great. I love it.

Only one strong suggestion! 
For the sake of implementation completeness, can you also add an IOC R AWAKE, too?
That'd give to the userspace the current status of the wakeup schedule (yes/no) by returning the result of function device_may_wakeup(dev).

PR #1855, opened 2018-12-14
power: reset: msm-poweroff: set a restart reason when probing

oshmoun:
If a device using an S1 bootloader crashes due to a hardware failure, then the restart reason does not get set by the kernel.
For some devices, the default in this case is to throw a verification failure, which is not quite accurate as the hardware failure is not related to verity and other verification methods. After the message is shown, the device is shut down.
To achieve some robustness, set a restart reason at probe time that ensures the bootloader will not show a misleading error message and will instead reboot the device.

Change-Id: Id7507bb29fee1ff23584801d95f3e07402a5fe29


kholk:
I've requested this change a lot of time ago.
Finally.

MERGE!!!

PR #1857, opened 2018-12-17
[1/2] Rewrite et510 driver

MarijnS95:
This is a rewrite of the ET510 driver based on the FPC1145 driver. The original driver contains lots of commented out and unused code, and does not deal with power management. The FPC driver on the other hand is well-maintained and has gotten some neat features over time.

All the code is intentionally duplicated for future robustness. Some things (such as SODP's custom `/dev/fingerprint` interface) could have been abstracted away to separate files; let me know if this is desired.

Furthermore, there are some cleanups and formattings in the FPC driver (in separate commits of course).

TODO:
- [x] Test on Nile devices with FPC sensors
- [x] Quick test on other platforms, to make sure nothing breaks (tested on Loire/Suzu already)
- [ ] Figure out timings for (power-on) reset. Those from FPC seem to work fine, but are 2 (!!) orders of magnitude smaller than those from the original Egistec driver.
- [x] Egistec sets the regulator to 3.3v whereas FPC uses 1.8v. 1.8v works fine on Egistec, no stability issues or anything.
- [ ] The ioctl on the miscdev is far from ideal as it relies on both drivers to implement and expose it. Find a better solution: check which driver is loaded, or expose the hardware as a separate interface how Stock CEI does it?


jerpelea:
@kholk

jerpelea:
please remove formating changes

PR #1862, opened 2018-12-22
[2.3.2.r1] [SPECTRA-CAM] Bring in line with LA.UM.7.3 05900

kholk:
Update the LE.UM.2.3.2.r1.4 camera to be in line with LA.UM.7.3-05900-sdm845.0.

Tested on SoMC Tama Akatsuki.


kholk:
P.S.: Even if it is working (for some unknown reason) on my device, and actually even showing better performance, it's not working on @oshmoun 's Tama Apollo.
This probably requires blobs to be recompiled.

kholk:
Offending commit:
6c36a40ed550e7366acfa2cc93a56c3f35eb3ddf
"
    Register Actuator with CRM as real time device.
    Correct request ID for actuator manual lens move command
    for the lens to move in sync with request ID with frame
    delay of 1.
"

Does this need support in userspace?

EDIT:
The commit has been reverted to allow this update to work with the current blobs.

kholk:
@sraase This PR should be totally interesting to you.
Can you please validate?

Thanks!

oshmoun:
the commit can probably be added again, no changes noticed

kholk:
I agree, but that one makes a lot of spam in dmesg and maybe it requires some sort of userspace support.

For now let's keep it reverted, if/until Sebastian says otherwise.

sraase:
Sorry for not replying earlier, but I was on vacation.
I'll try to take a look this week.

kholk:
Sure, no problem :D

sraase:
As far as I can tell, the actuator works in both cases. Keeping it reverted seems the better option for now. Maybe this will be required in the future, but I don't know. Userspace shouldn't matter in this case, since it is mostly descriptive rather than code.

kholk:
Well, from our previous experience, we know that sometimes the smallest change in the camera drivers can break things unexpectedly... And that's why I wanted you to take a look :D

Anyway, if you approve on your side, I think that the pr can be merged at this point.

PR #1867, opened 2019-01-10
[2.3.2.r1.4] Sound techpack updates and stabilization

kholk:


kholk:
Added fix for SoMC Nile audio crash.

MERGEME.

kholk:
Also added SoMC Nile MBHC fix for headphones detection.

PR #1868, opened 2019-01-11
[2.3.2.r1.4] Encryption upgrades, F2FS hwcrypto support, hardening

kholk:
This PR adds some hardening, F2FS support for ICE and updates hardware
encryption to use the right key size on UFS and eMMC (which is different!).

Also adds some memory management improvement and some fixes for
some race conditions.

Tested on SoMC Discovery DSDS (SDE)

### DIFF
--- a/block/bio.c
+++ b/block/bio.c
@@ -568,8 +568,11 @@ EXPORT_SYMBOL(bio_phys_segments);
 static inline void bio_clone_crypt_key(struct bio *dst, const struct bio *src)
 {
 #ifdef CONFIG_PFK
-	dst->bi_crypt_key = src->bi_crypt_key;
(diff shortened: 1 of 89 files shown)


ix5:
Seems to work fine so far one kagura. Camera is ok.
Not tested: encryption.
[SSR bug](https://github.com/sonyxperiadev/bug_tracker/issues/268) is still present.

kholk:
Waiting for a Tama test before the magic word.

oshmoun:
works fine on tama after including a couple missed commits
breaks boot on tama in the current state, since one of the missed commits is critical to get it to boot

kholk:
Missing commits merged in the PR. Tama working totally fine now.

MERGE!

PR #1874, opened 2019-01-15
[2.3.2.r1.4] Remove WALT and SCHED_CORE_CTL/ROTATE

kholk:
These options are totally incompatible with rqbalance and even conflicting.


kholk:
GO GO GO :)

Akatsuki is flying! :D

PR #1881, opened 2019-01-21
[2.3.2.r1.4]tone: dts: convert i2c_3 gpios to pinctrl

erikcas:
See:
https://github.com/sonyxperiadev/kernel/pull/1667
and
https://github.com/sonyxperiadev/bug_tracker/issues/137

tested with aosp camera on kagura. Dmesg is clean now

### DIFF
--- a/arch/arm64/boot/dts/qcom/msm8996-tone-common.dtsi
+++ b/arch/arm64/boot/dts/qcom/msm8996-tone-common.dtsi
@@ -175,8 +175,8 @@
 		clocks = <&clock_gcc GCC_BLSP1_AHB_CLK>,
 			<&clock_gcc GCC_BLSP1_QUP3_I2C_APPS_CLK>;
 		pinctrl-names = "i2c_active", "i2c_sleep";
-		pinctrl-0 = <&msm_gpio_47 &msm_gpio_48>;
-		pinctrl-1 = <&msm_gpio_47 &msm_gpio_48>;
+		pinctrl-0 = <&i2c_3_active>;
+		pinctrl-1 = <&i2c_3_sleep>;
 		status = "disabled";
 	};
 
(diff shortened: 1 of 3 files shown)


erikcas:
Earlier had some issues with on build with SnapdragonCamera. Turned out to be an issue with SnapdragonCamera issue (not working properly on suzu either) so the SnapCamera idea is dropped

kholk:
I agree with you about the idea, that's proper configuration. Indeed.

But it wouldn't be the real world if things were THAT SIMPLE. :P
I have to request some changes.

First of all, by doing that you're generating a possible power consumption issue by leaving the not connected pins floating and open for input on Tone Dora: this phone doesn't have any of the two sensors.
Second thing, you've removed them from somc-pinctrl after changing their name/binding/function. That's also very very bad.
Put it like this: something goes bad while using camera, kernel panic happens, the device reboots and you get back into the system with the ToF/RGBC GPIOs in the previous state (HIGH) while they're expected to be LOW.
The best situation at this point would be a camera probe failure requiring a manual system reboot (or shutdown and boot).

So... Override these pins' config in dora-common to be in Hi-Z (or whatever they were, I couldn't read the entire thing since I'm on phone) and DEFINITELY readd them to somc-pinctrl!

Disclaimer: if I've missed anything it's because I'm on phone. In case, I'm preventively sorry.

kholk:
P.S.: I've missed the worst situation. That'd be that you burn the sensors while power supplies are initializing. Cool, eh? :P

erikcas:
oke @kholk understood your point. Should https://github.com/sonyxperiadev/kernel/commit/46688f79ff791adf9921e5c06f829e39ffe10272 be reverted then?

erikcas:
close till properly fixed, prevent leaving a pr open that can damage phones

PR #1882, opened 2019-01-21
drm: somc_panel: Apply color calibration when selecting pcc profile.

MarijnS95:
Calling setup only prepares the profiles, but does not apply them to the
display, meaning that writing to the pcc_profile sysfs attribute has no
effect.
Change the call to a function applies the pcc instead (which calls setup
as well to make sure the profiles are available).


kholk:
On the fbdev version of the somc_panel driver, this was an intentional hack.
The display would get blanked and unblanked to avoid a crash.

I didn't verify whether this was fixed or not in SDE, but hell if it is let's merge this.
P.S.: Test on Tama and Nile!

MarijnS95:
Closing. This has already been addressed in the refactor in https://github.com/sonyxperiadev/kernel/pull/1892.
More specific, here: https://github.com/sonyxperiadev/kernel/pull/1892/files#diff-47ff5f87ed80b90d13c6d8d2e899cce7L550

PR #1883, opened 2019-01-21
[DONOTMERGE] drm: somc_panel: Override set_property to intercept and merge display color calibration

MarijnS95:
Both the somc_panel display manager and the Android system apply color calibrations to the panel through the PCC dspp, the latter of which is usually used to apply night-light.

This change intercepts the call from the Android system, and blends it together with the selected PCC calibration profile, before forwarding the entire calibration to the PCC module.

Currently only enabled for Akatsuki displays.


kholk:
I don't like having that property.
If a device supports PCC, I really don't see why it shouldn't support different parameters. Told you Marijn, I appreciate the effort in making it foolproof, but having a DT property to enable it should be overkill..... 

Well, to whoever else reads this comment: the code is basically perfect and fixes Android's night light via color transform. 

I'll do some testing on Nile and Yoshino before telling you to SEEK&DESTROY that DT commit.... :P

MarijnS95:
> telling you to SEEK&DESTROY that DT commit...

Destroyed. :stuck_out_tongue_winking_eye:

MarijnS95:
Also, don't forget the comment I made about a possible optimization. It's a waste of space and cycles to store a copy of the PCC profile as `drm_msm_pcc` (especially since there are a lot of zeroes). On the other hand, it allows a complete matrix multiplication, may we ever need that in the future (though the merge is not nearly complete anyway as all other values are omitted).
What do you prefer @kholk ?

kholk:
Allowing a complete matrix multiplication equals to code flexibility. When the code is flexible, it's open for improvements, enhancements and reuse. Keep that in mind.

Then... You have a point about the memory consumption but IMHO storing a struct (that's not even that big actually...) is way better than reading it back and re-multiplying the whole thing at every display power on. 
I think that the best option would be to keep the parameters in memory, and to even keep the processed (multiplied) ones until we either receive a zero (or is it one?) param from userspace that means "no more overrides from here". 

Fast recap:
1. Allow flexibility as much as possible
2. Keep things in memory, prefer using RAM over CPU cycles
3. Dispose of the things when userspace disables the transformation.

What do you think?

MarijnS95:
Memory consumption on it's own isn't worrying me; it's the detour of reading the profile and storing it back in memory, leading to unnecessary duplication and confusion.

Currently the final matrix isn't stored, but recalculated when the display turns on. While caching would be easy to do when storing just the profile (meaning that it's only invalidated when the profile changes or the system requests a different calibration), comparing the source calibrations to *another* copy will cost more work, memory and code complexity than just multiplying it "every time".

Disabling this codepath is possible when the system requests an identity transform (again, assuming it doesn't use any of the other 2nd-order parameters).

Following your points, the most performant thing I can think of:
1. Upon receiving a calibration from the system, check if it's not an identity matrix (approx 9 ~ 12 comparisons, depending on seemingly unused constant offsets). Disable the codepath if it's an identity matrix (but apply the pcc profile if it wasn't an identity previously).
   Otherwise, compare to a stored version, and proceed merging+applying if different.
2. Upon changing pcc profile, only proceed if it's different from a previously stored profile (comparison on the profile index is easiest). Again, we can disable the merging codepath if the choosen profile results in an identity transform, much like the sytem calibration (except that it takes 3 comparisons).
3. Make either:
   - A simplified matrix multiplication with all the zeroes worked out, leaving just `r_data`, `g_data` and `b_data` for the profile.
   - Local conversion from `dsi_pcc_color_tbl` to `drm_msm_pcc`, followed by a full matrix multiplication.
4. When turning the display on, apply the cached transformation.

I personally think the above flow isn't worth all the extra code and maintenance. I'd personally:
1. Disable the merge if either is an identity;
2. Keep a cached version of the result;
3. Invalidate and update the cache when the pcc profile index *changes* or when the system requests a different calibration.
   Do this without comparing the system calibration. Night light scales gradually, sending a lot of updates and requiring lots of memory loads and comparisons that result in a merge anyway. Chances that the same transform is sent are low;
4. Do not keep a global copy of the matrix representation of the current pcc profile, but either:
  - Calculate a local copy on the fly;
  - Optimize the matrix multiplication to take only rgb components into account;
5. Apply the cached matrix when the screen turns on (or recompute if it's invalid for some reason).

In the end, this requires the system pcc, final pcc and profile index to be stored.

kholk:
So....

1. Calibrations from the kernel are gold: they must **always** be a base for _anything_ coming from the userspace: you have to think about them like correcting the display from being a total red (put_different_color_here) garbage to showing something that has some sense.

2. 
if (transformed_pcc_index == 2 && current_pcc_index == 2) {
    apply_current_index(); // is that needed?
    use_cached_transform();
} else {
    apply_current_index(); // is that needed, again?
    multiply_the_stuff_and_cache_it(); // it's heavy to do it every time, of course
}
P.S.: I don't have to tell you, but the apply_current_index would be out of the if branch :P 

3. Don't simplify the matrix calculation: we may reuse this code for a future feature enhancement and having it there, ready and used (even if not at its full potential), makes it easier to just "combine" things together when time will come.

4. Remultiplying wastes cycles. Checking if var_a == var_b is way better..... of course, unless you need to check 40 variables vs 5 multiplications..... but I don't think that this is really the case, since - as you wrote - we can just check for the kernel calibration index and if it matches... well.

5. Of course you need to reapply the stuff whenever you turn the screen on.. and I really have no comment on that.

NOTE: Points (numbers) != your points :P

MarijnS95:
@kholk The optimizations have been pushed. A full matrix multiplication is still done, but only when the choosen pcc profile or system pcc calibration changes. With that, a few fixes to enable PCC profiles on Nile SDE have been pushed as well.

Again, I do think that `memcmp`ing the system calibration is unnecessary. The system only sends transforms when it changes anyway (not even when the screen turns on to re-apply), so this is just a waste of time to capture maybe one or two duplicate transforms (over the few hundred that are sent while smoothly fading night-light).

The new changes make https://github.com/sonyxperiadev/kernel/pull/1882 obsolete. Since this works just fine on Discovery and Tama (except that the former still errors out when the system applies the color transform), we can conclude that blanking and unblanking is not necessary anymore :smiley:

jerpelea:
please remove the first merge.

MarijnS95:
@jerpelea The commits have been removed (sorry for that, blind mid-night rebase on top of the wrong branch), but github *will not* update the commits shown until it is reopened.

jerpelea:
please create  a new PR

PR #1884, opened 2019-01-23
[2.3.2.r1.4] No more zombies! Plus a load of other fixes and enhancements

kholk:
Who likes zombies?
We don't like zombies! At least not zombie phones, they're bad.

This PR contains loads of fixes, including:
- USB Tethering should now finally work on SoMC Loire
- NULL checking here and there
- Some code hardening
- Solution for an IPA related power consumption issue
- Switch to the kernel generic QRWLOCKs, ARM64 wide
- Generic hotplug code fixes
- MPM fixes

.... and last but not least, I've just DESTROYED the sleep bug issue.
Finally.

Tested on:
- SoMC Yoshino Lilac RoW
- SoMC Nile Pioneer+Discovery DSDS

Needs testing on SoMC Loire BEFORE MERGE.

_Definitely_ enjoy!


oshmoun:
randomly breaks boot on loire
also causes slow sim card detection
pstore: https://del.dog/mibatuqisa.cs
also just noticed there are irq issues that weren't there before: https://del.dog/amomufuyuf.cs
the mpm changes seem to break compatibility with gic v1 upon first glance

oshmoun:
on an unrelated note: usb tethering does indeed (finally) work on loire

drivers/irqchip/qcom/mpm.c
```diff
@@ -261,16 +261,23 @@ static void msm_mpm_gic_chip_disable(struct irq_data *d)
...
+}
+
 static struct irq_chip msm_mpm_gic_chip = {
 	.name		= "mpm-gic",
 	.irq_eoi	= irq_chip_eoi_parent,
 	.irq_mask	= msm_mpm_gic_chip_mask,
 	.irq_disable	= msm_mpm_gic_chip_disable,
 	.irq_unmask	= msm_mpm_gic_chip_unmask,
-	.irq_enable	= msm_mpm_gic_chip_disable,
+	.irq_enable	= msm_mpm_gic_chip_enable,
```
oshmoun, line 290: please fix this in the original commit introducing the change :)

PR #1888, opened 2019-01-24
arm64: DT: Nile: Reconfigure LEDs PWM parameters

MarijnS95:
From https://github.com/sonyxperiadev/kernel/pull/1876
Nile needs the same lut flags to show a breathing effect. The old flags result in a ramp-down only, just like on Tama.

Change the rest of the parameters to be consistent with the configuration for Tama and Yoshino.


kholk:
I agree

PR #1891, opened 2019-01-24
fs: Improve eventpoll logging to stop indicting timerfd

ix5:
cherry-picked from https://review.lineageos.org/c/LineageOS/android_kernel_xiaomi_jason/+/236442

timerfd doesn't create any wakelocks; eventpoll can, and is creating the wakelocks we see called "[timerfd]".  eventpoll creates two kinds of wakelocks: a single top-level lock associated with the eventpoll fd itself, and one additional lock for each fd it is polling that needs such a lock (e.g. those using EPOLLWAKEUP).
Current code names the per-fd locks using the undecorated names of the fds' associated files (hence "[timerfd]"), and is naming the top-level lock after the PID of the caller and the name of the file behind the first fd for which a per-fd lock is created.

### DIFF
--- a/fs/timerfd.c
+++ b/fs/timerfd.c
@@ -44,6 +44,8 @@ struct timerfd_ctx {
 	bool might_cancel;
 };
 
+static atomic_t instance_count = ATOMIC_INIT(0);
+
 static LIST_HEAD(cancel_list);
 static DEFINE_SPINLOCK(cancel_lock);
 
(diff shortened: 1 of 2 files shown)


kholk:
Harmless and makes debugging life easier.

MERGE

ix5:
NOOOO
I haven't even compiled this yet!

fs/timerfd.c
```diff
@@ -387,6 +389,10 @@ SYSCALL_DEFINE2(timerfd_create, int, clockid, int, flags)
 {
 	int ufd;
 	struct timerfd_ctx *ctx;
+	enum alarmtimer_type type;
```
kholk: type: unused variable

kholk:
There's only an unused variable issue to solve, then it's fine.

ix5:
Fixed unused var, compiled, booted, works.

kholk:
Now there's no excuse for not merging it :)

PR #1892, opened 2019-01-25
drm: somc_panel: Override set_property to intercept and merge display color calibration

MarijnS95:
Recreated from https://github.com/sonyxperiadev/kernel/pull/1883 because _*I*_ made a mess there.
Please read it if you are interested in the ideas behind the code.

Both the somc_panel display manager and the Android system apply color calibrations to the panel through the PCC dspp, the latter of which is usually used to apply night-light.

This change intercepts the call from the Android system, and blends it together with the selected PCC calibration profile, before forwarding the entire calibration to the PCC module.

The first two commits contain the PCC profiles and a force case to enable this functionality on Nile SDE. The (nightlight) color calibrations are still rejected elsewhere though.

Copied [comment](https://github.com/sonyxperiadev/kernel/pull/1883#issuecomment-457405778) regarding the current state, which should still be considered before merging:
> A full matrix multiplication is still done, but only when the choosen pcc profile or system pcc calibration changes.
> 
> Again, I do think that `memcmp`ing the system calibration is unnecessary. The system only sends transforms when it changes anyway (not even when the screen turns on to re-apply), so this is just a waste of time to capture maybe one or two duplicate transforms (over the few hundred that are sent while smoothly fading night-light).
>
> The new changes make #1882 obsolete. Since this works just fine on Discovery and Tama (except that the former still errors out when the system applies the color transform), we can conclude that blanking and unblanking is not necessary anymore :smiley:


kholk:
What :D

I can calibrate two displays, but the others that I don't have... You can't copy the Akatsuki calibrations to Nile panels!! :P

Replace the three rgb components with a "100%" (0x8000 0x8000 0x8000), so that at least we don't get strange things going on with uncalibrated panels... And regarding the adacolor, I really wouldn't dare using the Akatsuki parameters, because that panel has a bad quirk on low brightness that I've needed to address with that, and I imagine that you'll get a very oversaturated image on Nile panels (or any LCD based panel...) on low brightness with these settings....

Regarding all the rest of your work, as you know, I consider it great.
I can help you to try to fix the Nile userspace issue, as long as it's getting something wrong from the kernel, though. 

Anyway, the whole point is: if you leave these crapped calibrations on the Nile DTs we won't be able to merge this PR but, if you make them neutral as I said, we can start merging this and keep them like this until I properly calibrate Nile. 
I mean, at least we get the functionality up on Tama.

MarijnS95:
@kholk The calibrations were merely to test the functionality on Nile, as asked in the other PR. I don't like having dead and incomplete calibrations lying around, so opted to remove the commit altogether (unless you insist on having those as a starting point).

The Nile userspace issue is a weird one - I see the same ioctls going from a surfaceflinger thread to the kernel, but even with full drm debugging enabled no errors are shown. Unfortunately, the userspace part lives in a proprietary blob (`libsdm-color.so`) making it hard to insert debugging probes, unless we put those in the kernel. This is something I can do later, with a slowly growing understanding of the `set_property` codepath.

kholk:
I insist, keep the 0x8000 non-calibrations as a starting point.
Makes life easier to people that wouldn't know if there are any differences between Nile and the rest ;)

MarijnS95:
@kholk An initial no-op configuration has been added. The DT table "parsing" code has been simplified to use `prop->length` instead of the explicit `-size` property.

Three final things:
- Should the `-size` removals be separated across multiple commits (one per file, or one per device)?
- Should we deprecate `-enable` as well, in favor of simply checking if the table property is there?
- When one of the color tables is not available, `dsi_pcc_applied` is never set to `true`, meaning that it's ran every time a different profile is selected. Is this intended?

kholk:
The DT changes should be done as one commit per PLATFORM, so in one commit you do it for the entire Tama, one for Nile, one for.. Blahblah.

The enable property shall not be deprecated: it's great for debugging, for early porting, and in case some device experiences a problem that is not immediately fixable. You just remove the enable property and you don't have to comment out or remove an entire load of data. It's just commodity!

The applied variable... I honestly don't remember... Sorry!

MarijnS95:
@kholk @jerpelea This PR is done from my side. Tested on SoMC Suzu, SoMC Discovery SDE and SoMC Akatsuki.

@kholk As to the variable, see the latest commit. It's mainly about removing the `goto exit;`s to prevent it from redoing the setup time and time again, when one of the color tables is missing (which wasn't fatal btw, just a `return 0` so that the rest of the code could still run).

The DT commits have been separated to the best of my abilities. Some have been split per DEVICE, because no other device of that platform was using them, and others are a merge of two DEVICES from two **different** platforms (which is why I made that first decision) because they use the same panel.

kholk:
Good for me.

MERGE

PR #1893, opened 2019-01-25
fs: Improve eventpoll logging to stop indicting timerfd [LA.UM.6.4.r1 backport]

ix5:
This is a backport of #1891 to the 4.4 LA.UM.6.4.r1 branch.

timerfd doesn't create any wakelocks; eventpoll can, and is creating the wakelocks we see called "[timerfd]".
eventpoll creates two kinds of wakelocks: a single top-level lock associated with the eventpoll fd itself, and one additional lock for each fd it is polling that needs such a lock (e.g. those using EPOLLWAKEUP).  Current code names the per-fd locks using the undecorated names of the fds' associated files (hence "[timerfd]"), and is naming the top-level lock after the PID of the caller and the name of the file behind the first fd for which a per-fd lock is created.

### DIFF
--- a/fs/eventpoll.c
+++ b/fs/eventpoll.c
@@ -1237,15 +1237,23 @@ static int ep_create_wakeup_source(struct epitem *epi)
 {
 	const char *name;
 	struct wakeup_source *ws;
+	char task_comm_buf[TASK_COMM_LEN];
+	char buf[64];
+
+	get_task_comm(task_comm_buf, current);
 
 	if (!epi->ep->ws) {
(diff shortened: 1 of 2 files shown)


ix5:
Compiled, seems fine.

jerpelea:
@ix5 The 4.4 kernel is stable and we may risk breaking it with back-ports 
Please test the commit on several devices and report back

ix5:
Then let's not waste effort on this. The commit in 4.9 was mainly to aid in debugging sleep issues, from which 4.4 doesn't suffer as much.

PR #1904, opened 2019-02-02
[2.3.2.r1.4] arm64: setup: Sony: Add boot parameter to restore TLMM UART pins

kholk:
Some Sony MSM bootloaders are closing the UART port by setting
the GPIO4, GPIO5 in a "closed" state.

To get back access to the early console, add an early function
that will restore the correct pin configuration in the TLMM
block on the GPIO4 and GPIO5 (UART_TX, UART_RX).

This is activated with an early parameter, used as following:
restore_msm_uart=<TLMM_BASE address>

For example, on MSM8996:
restore_msm_uart=0x1010000

P.S.: TLMM_BASE can be very easily retrieved from the pinctrl.

### DIFF
--- a/arch/arm64/kernel/setup.c
+++ b/arch/arm64/kernel/setup.c
@@ -279,6 +279,36 @@ static int __init sony_param_warmboot(char *p)
 }
 early_param("warmboot", sony_param_warmboot);
 
+#define MSM_PINCONF_2MA_NO_PULL		0x8
+#define MSM_PINCONF_2MA_PULL_DN		0x9
+#define MSM_TLMM_PIN_UART_RX		0x5000
+#define MSM_TLMM_PIN_UART_TX		0x4000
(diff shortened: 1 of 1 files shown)


bartcubbins:
this is not common. for example:
on loire we have GPIO20 (UART_TX) and GPIO21 (UART_RX)

kholk:
Oh craB :)

In the meanwhile, you can use a shifted TLMM_BASE.
Each GPIO has length 0x1000, so you can shift up your TLMM_BASE of ((0x1000*20)-0x4000).
That's of course no good.

I will refactor the function so that it will work by specifying the TX GPIO address: the RX one
is always TX_GPIO+1, so that would work.

kholk:
Sorry for the mess on PRs, this got on a bad branch on GitHub.
The commit has been fixed in the following PR: https://github.com/sonyxperiadev/kernel/pull/1907

PR #1907, opened 2019-02-05
[2.3.2.r1.4] [RESEND] arm64: setup: Sony: Add boot parameter to restore TLMM UART pins

kholk:
Some Sony MSM/APQ/SDM bootloaders are closing the UART port by
seting its relative GPIOs in a "closed" state.

To get back access to the early console, add an early function
that will restore the correct pin configuration in the TLMM
block on the UART_TX, UART_RX GPIOs.

This is activated with an early parameter, used as following:
restore_msm_uart=0x<TLMM+PIN_UART_TX_OFFSET>

NOTE: On MSM, GPIOs are at a 0x1000 "distance" from each other
Example: TLMM start     = 0x1010000
         UART TX        = GPIO4
         UART TX OFFSET = 0x4000
param:   restore_msm_uart=0x1014000

P.S.: TLMM_BASE can be very easily retrieved from the pinctrl.

### DIFF
--- a/arch/arm64/kernel/setup.c
+++ b/arch/arm64/kernel/setup.c
@@ -279,6 +279,52 @@ static int __init sony_param_warmboot(char *p)
 }
 early_param("warmboot", sony_param_warmboot);
 
+/*
+ * sony_param_restore_uart: Restore pin configuration for UART pins
+ *                          UART_TX: Drive Strength 2mA, Pull-Down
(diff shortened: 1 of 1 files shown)


bartcubbins:
tested on my suzu. can confirm

kholk:
Tone also works fine.

MERGE.

PR #1909, opened 2019-02-06
[2.3.2.r1.4] GCC 8 support, SMP2P RDBG disablement, clock fixes, audio fix

kholk:
In this patchset:
- Support to build the kernel on GCC 8.x (getting it to work fine when compiled with that one is another story...)
- DT cleanup for SoMC Yoshino
- Some fixes for MSM8998 clocks
- Better -costless- debugging on Q6ASM
- Fixed sudden audio loss issue

Tested on SoMC Yoshino Maple RoW


kholk:
Added commits to fix the sudden loss of speaker usecase audio on all legacy platforms.

This PR is critical.
MERGE

PR #1906, opened 2019-02-05
[2.3.2.r1.4] Couple of fixes for all platforms

kholk:
This PR includes a couple of fixes and enhancements for all platforms,
but mainly MSM8998 Yoshino.

Tested on SoMC Yoshino Maple RoW.

Please compile test on all platforms before the green light.

PR #1912, opened 2019-02-10
input: misc: fpc1145: Remove unused variable and log hardware string.

MarijnS95:
Remainder of the fixes in https://github.com/sonyxperiadev/kernel/pull/1906/commits/8cd62e7f37a4311e6b70ca46659038c4ed23d96a and https://github.com/sonyxperiadev/kernel/pull/1906/commits/6a08052ea80a71fb9908f1cfee3c4c4274e3a07b. These were supposed to go with more useful commits, but those never happened.
The warning has already been addressed, this fixes an earlier request for _useful_ logging.

### DIFF
--- a/drivers/input/misc/fpc1145_platform.c
+++ b/drivers/input/misc/fpc1145_platform.c
@@ -560,10 +560,10 @@ static int fpc1145_probe(struct platform_device *pdev)
 	int rc = 0;
 	size_t i;
 	int irqf;
+	struct device_node *np = dev->of_node;
 #ifdef CONFIG_ARCH_SONY_NILE
 	int hw_type;
 #endif
-	struct device_node *np = dev->of_node;
 
 	struct fpc1145_data *fpc1145 =
 		devm_kzalloc(dev, sizeof(*fpc1145), GFP_KERNEL);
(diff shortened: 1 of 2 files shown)


drivers/input/misc/fpc1145_platform.c
```diff
@@ -598,13 +598,15 @@ static int fpc1145_probe(struct platform_device *pdev)
...
-	dev_info(dev, "Detected hw type %d\n", hw_type);
 
 	(void)vreg_setup(fpc1145, VDD_ANA, false);
 
 	if (hw_type != FP_HW_TYPE_FPC) {
+		dev_info(dev, "FPC sensor not found, bailing out\n");
 		rc = -ENODEV;
 		goto exit;
+	} else {
+		dev_info(dev, "Detected FPC sensor\n");
```
oshmoun: nitpick, else is not needed because of that `goto exit;`

MarijnS95: Thanks, being a vast proponent of removing unnecessary cases to improve readability, I don't know how that slipped in :sweat_smile:

PR #1920, opened 2019-02-20
[2.3.2.r1.4] Support displayless devices on Android

kholk:
This introduces CONFIG_DRM_MSM_EMU_HEADLESS:

Enable simulating a display connected on the DSI while using a
display-less system.

This configuration option is useful when using Android userspace
on a system that has no physically connected display and no
possible or not connected external display.

This way, the Android system will initialize the display system
just fine and it will be possible to interact with the UI by
using screen mirroring via the Android Debug Bridge (ADB).

Tested on SoMC Tama Akari DSDS (display physically not present).


kholk:
Added a PERFORMANCE CRITICAL commit.

PR #1927, opened 2019-03-04
[2.3.2.r1.4] msm-jpegdma: Add V4L2_CAP_DEVICE_CAPS to caps

ix5:
See `v4l_querycap()` in `v4l2-ioctl.c`
Otherwise the query fails with a crash:
```
[...]
Bad caps for driver msm_jpegdma, 4200003 0
------------[ cut here ]------------
WARNING: CPU: 1 PID: 23782 at v4l_querycap+0x90/0xd4
Modules linked in:
CPU: 1 PID: 23782 Comm: ExtCamHotPlug Tainted: G W 4.9.154-gb61c5e1840de-dirty #23
Hardware name: SoMC Kagura-ROW (DT)
task  ffffffc037b85e80 task.stack: ffffffc0b7188000
PC is at v4l_querycap+0x90/0xd4
LR is at v4l_querycap+0x90/0xd4
[...]
```

### DIFF
--- a/drivers/media/platform/msm/camera_v2/jpeg_dma/msm_jpeg_dma_dev.c
+++ b/drivers/media/platform/msm/camera_v2/jpeg_dma/msm_jpeg_dma_dev.c
@@ -637,8 +637,9 @@ static int msm_jpegdma_querycap(struct file *file,
 	cap->bus_info[0] = 0;
 	strlcpy(cap->driver, MSM_JPEGDMA_DRV_NAME, sizeof(cap->driver));
 	strlcpy(cap->card, MSM_JPEGDMA_DRV_NAME, sizeof(cap->card));
-	cap->capabilities = V4L2_CAP_STREAMING |
-		V4L2_CAP_VIDEO_OUTPUT | V4L2_CAP_VIDEO_CAPTURE;
+	cap->device_caps = V4L2_CAP_STREAMING |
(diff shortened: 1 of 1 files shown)


jerpelea:
@kholk

kholk:
So, this thing shouldn't even happen.
Uhm, the v4l_querycap shouldn't get called, especially from an external camera trigger, let alone internal (ISP) ones.

In any case, if we want to solve it like this... let's say that I don't want to tell you that at this point the entire camera_v2 needs basically the same fix....
...but anyway, your fix is wrong.

You want to do something like this:

```c
cap->device_caps = V4L2_CAP_STREAMING |
		V4L2_CAP_VIDEO_OUTPUT | V4L2_CAP_VIDEO_CAPTURE;
cap->capabilities = cap->device_caps | V4L2_CAP_DEVICE_CAPS;
```

But then obviously, this is something to keep in mind:
If anything is calling querycap to query cap->capabilities or cap->device_caps, then
that thing is expecting a reply containing either of the two bitmasks.
If the userspace code happens to be buggy (which is not that much unlikely), then
major crashes will happen.

Anyway, fix your patch like I said and let's see: if it works, then it .. works :)

ix5:
Compiled with your suggestion, seems to work ok so far(well it should ofc).
Force-pushed as well.

kholk:
Can anyone else please test this on any other legacy platform?

It looks fine to me now, anyway.
Thanks!

PR #1929, opened 2019-03-05
msm: mdss: share MDP smmu device mappings with other mdss clients

oshmoun:
Rotator and MDP share same stream ID on sdm600 target,
hence share the smmu device with rotator device to map/unmap
its buffers.

The change will also handle different secure usecase concurrencies
like, mdp running in secure and rotator in non-secure and vice versa.

Change-Id: I3ff118baed3984d63e9a9fe94289d99523c7b3e9
Signed-off-by: Kalyan Thota <kalyant@codeaurora.org>


oshmoun:
This does NOT fix any issues, since we are not using the fbdev driver.

oshmoun:
Not needed since the rotator is now unused

PR #1938, opened 2019-03-18
[2.3.2.r1.4] dts: tone: Set sound gpio to low by default

ix5:
The `output-high` setting in [pm8994_gpio_13: hph0_enable in msm8996-mtp.dtsi](https://github.com/sonyxperiadev/kernel/blob/aosp/LE.UM.2.3.2.r1.4/arch/arm64/boot/dts/qcom/msm8996-mtp.dtsi#L877) needs to be overwritten since the GPIO should be disabled by default and only switched on for call sounds.

This commit fixes the bug that the top speaker was only initialized after switching to call audio had force-toggled the GPIO.

### DIFF
--- a/arch/arm64/boot/dts/qcom/msm8996-tone-common.dtsi
+++ b/arch/arm64/boot/dts/qcom/msm8996-tone-common.dtsi
@@ -761,14 +761,8 @@
 
 /* GPIO_13: EAR_EN */
 &pm8994_gpio_13 {
-	/* Inherited from MTP */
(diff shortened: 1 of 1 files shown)


oshmoun:
@kholk pinctrl

kholk:
Well it should've been inherited from MTP with these parameters already, but apparently, at this point, I've missed some spaghetti style override that crapped it. 

Yeah that change looks valid. 
MERGE!

PR #1943, opened 2019-03-19
[2.3.2.r1.4] arm64: DT: Ganges: Move common nodes from kirin and mermaid to ganges

konradybcio:
These nodes are same for both devices, and therefore
they should be in the ganges dt instead.

It compiles, and shouldn't break anything, yet I wouldn't mind if someone tested it on a physical device.


konradybcio:
To clarify: the int_codec node comes from mermaid and kirin dts and is exactly same for these devices yet for some reason the ganges DT has a different one, so replacing it with mermaid/kirin one seems to be the a good idea.

konradybcio:
Also, don't mind the forcepush mess, I'm still learning how to use git

PR #1946, opened 2019-03-27
[2.3.2.r1.4] Linux 4.9.166

nathanchance:
@kholk Please see `lib/int_sqrt.c`, I replaced https://github.com/sonyxperiadev/kernel/commit/f65994212d7351d4fecfd8e29c819487b8ecd48f with upstream's seemingly equivalent fix: https://git.kernel.org/stable/linux/c/e6008a0525e4f9d6298762530e51dbf114455a47. If you object, I'll reset the file.


kholk:
@nathanchance Thanks for notifying me!

It's fine like this, we're losing very very small performance cause we're not inlining the now-very-little-higher performance function, but again it's totally fine... in the end, it's better to be in line with upstream.

PR #1949, opened 2019-03-29
[2.3.2.r1.4] [IN PROGRESS] Remove arm-smmu-legacy driver

kholk:
This patchset migrates all the supported platforms to the new arm-smmu driver.
This was possible after recent kernel updates.

Tested on:
- SoMC Ganges Mermaid DSDS
- SoMC Tone Keyaki RoW

** THIS NEEDS TESTING ON YOSHINO, NILE **


kholk:
Pushed more commits.
- Solved an issue with SDM636 devices, where booting would be impossible without disabling either the secure context on the SDE or the DisplayPort driver
- Solved a global issue with some IOMMUs not being able to probe due to the context fault interrupt requests failure (mostly on secure contexts): it is unclear whether this is a firmware bug or a GIC driver bug, but this * at least temporarily * makes it to work fine
- Introduced a IOVA range variation in the SDE for Qualcomm platforms with legacy IOVA ranges to solve an issue with OOB allocations: this is valid for legacy platforms, like SoMC Tone, Yoshino, Nile, Ganges

Enjoy

kholk:
New commits to solve an issue with the IOMMU probe when msm-bus is not yet ready.
Now it probes fine.

Also, adding SoMC Tone Keyaki RoW to the list of tested devices.

PR #1950, opened 2019-04-03
[2.3.2.r1.4] [Ganges] Add QSEECom TA heap

MarijnS95:
Sony Open Devices blobs for Ganges (which are different/newer than
stock) use a separate heap for TAs. For example, loading the Egistec
fingerprint tzapp results in the following error:
E QSEECOMAPI: Error::Error while trying to allocate data from heap 19
E QSEECOMAPI: Error::ION_memalloc failed
E QSEECOMAPI: Error::Loading image failed with ret = -12
because the heap was missing.
This is a copy of the sdm845 QSEECOM TA heap definition.

### DIFF
--- a/arch/arm64/boot/dts/qcom/sdm630-ion.dtsi
+++ b/arch/arm64/boot/dts/qcom/sdm630-ion.dtsi
@@ -27,6 +27,12 @@
 			qcom,ion-heap-type = "DMA";
 		};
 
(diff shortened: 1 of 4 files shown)


kholk:
The problem with this is that the firmware does NOT have any definition of the TA heap 19....

Run the board with this commit for a while before merging it, this stinks a lot like firmware mess around the corner.

PR #1967, opened 2019-04-27
clk: gcc-msm8976: vote for higher vdd on bt clock

oshmoun:


kholk:
Did you actually try NOM on the lower clock instead of locking the main vdd vote to TURBO whenever you have serial on (always when BT is on)?

oshmoun:
nope, I'm that desperate :)

kholk:
I know.
Globally voting for TURBO on that regulator.... If that works... Then this means that we have undervoltage on literally *any* of the gcc clocks (well, actually most of them, as far as I remember there is a different regulator for the GPU clocks)......... 

.... So if that works we can safely say that we need to research on what clock causes that, as this can be impacting something else that we still don't see, creating some other issue that we didn't experience.

oshmoun:
so to cover all bases, went ahead and diffed regulator consumers between downstream and mainline
there are some interesting differences (downstream is on the left)
https://www.diffchecker.com/eLkg3xpI
to sum it up:
`mdss_dsi_ctrl0` voters are noops currently (min/max are 0)
`gcc` is stuck at highest level 384
`pm8004_s2` is undervolted

EDIT: so just to make it clear, after seeing this, I highly doubt this PR will produce any positive results (gcc is already voting for max vdd level)

PR #1968, opened 2019-04-30
drivers: gpu: msm: remove features and enable soft fault detection

oshmoun:
remove preemption and 64bit addressing features
enable soft fault detection (was on by default on 4.4)
disable gpmu (probably unnecessary as a508/a509/a510 do not support it)

### DIFF
--- a/drivers/gpu/msm/adreno-gpulist.h
+++ b/drivers/gpu/msm/adreno-gpulist.h
@@ -255,7 +255,6 @@ static const struct adreno_gpu_core adreno_gpulist[] = {
 		.major = 1,
 		.minor = 0,
 		.patchid = ANY_ID,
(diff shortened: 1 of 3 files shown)


oshmoun:
@jerpelea

kholk:
Soft fault and gpmu parameters can and should be set on the kernel command line instead of this dirty thing.
About the 64 bits addressing features, they are usable since the iommu got switched to AArch64 addressing instead of AArch32 LPAE, which was done by me on Loire, but by Qualcomm themselves starting from MSM8998.

oshmoun:
the idea here is to disable specific features and hope that the issue can be narrowed down to a single one.
but, these seem innocent because both nile and loire crashed (nile with a gpu hang in pstore and loire with no error in pstore)

PR #1971, opened 2019-05-04
[2.3.2.r1.4] FAAAAAASSST! Performance improvements for ARMv8

kholk:
This patchset includes some fixes (with two of them being for all supported platforms)
and some performance improvements for ARMv8.1 onwards.

The main improvement is the introduction of the usage of the CRC32 instruction that
is available on selected AArch64 processors, which speeds up crc32 calculation by
a.. _whole lot_.
This is done through cpufeatures, the kernel will always have the CRC32 accelerated
routines compiled-in, but they will be enabled only if the processsor that we're currently
running does actually support the CRC32 extensions.

Regarding us - specifically - ALL of our supported devices' SoCs DO actually support
the CRC32 instruction, so all of them will get accelerated.

Tested on SoMC Tama Akatsuki RoW

### DIFF
--- a/arch/arm64/Kconfig
+++ b/arch/arm64/Kconfig
@@ -51,6 +51,7 @@ config ARM64
 	select CLONE_BACKWARDS
 	select COMMON_CLK if !ARCH_QCOM
 	select CPU_PM if (SUSPEND || CPU_IDLE)
+	select CRC32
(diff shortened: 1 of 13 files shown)


kholk:
This is working totally fine on Akatsuki and Mermaid.
What are we waiting for?

MERGEME

PR #1974, opened 2019-05-05
[2.3.2.r1.4] allow rapid switching of gfx3d clock rate

oshmoun:
https://github.com/sonyxperiadev/kernel/commit/8ac3f526b6587ce3427941326d39297199783abf introduced a regression where `gfx3d_clk_src` can not switch rates fast enough, and would cause the device to crash. revert gpucc to a previous state to avoid that issue.

### DIFF
--- a/drivers/clk/qcom/gpucc-sdm660.c
+++ b/drivers/clk/qcom/gpucc-sdm660.c
@@ -37,7 +37,7 @@
 
 static DEFINE_VDD_REGULATORS(vdd_dig, VDD_DIG_NUM, 1, vdd_corner);
 static DEFINE_VDD_REGULATORS(vdd_mx, VDD_DIG_NUM, 1, vdd_corner);
-static DEFINE_VDD_REGULATORS(vdd_gfx, VDD_DIG_NUM, 1, vdd_corner);
+static DEFINE_VDD_REGS_INIT(vdd_gfx, 1);
 
 enum {
 	P_CORE_BI_PLL_TEST_SE,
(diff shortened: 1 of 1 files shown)


oshmoun:
had a crash on kirin while testing stability:
https://del.dog/abocixuqop.cs
the reverted commit might have had the purpose of fixing this crash, though it doesn't say explicitly. or this crash might be unrelated and it's occurring just now because the device became stable enough to get into such a situation.

oshmoun:
ok that issue was indeed resolved by the reverted commit, but almost looks like that was an unintentional change. the commit inits GPU XO early, and sets it as a parent for other clocks, but also removes the entries in the freq tables that would reference it, rendering the early init and parent switch useless.
And indeed, the following diff, applied on top of this PR, seems to resolve the crashing:
```
diff --git a/drivers/clk/qcom/gpucc-sdm660.c b/drivers/clk/qcom/gpucc-sdm660.c
index b0e93c998b6c..0e50ec08653a 100644
--- a/drivers/clk/qcom/gpucc-sdm660.c
+++ b/drivers/clk/qcom/gpucc-sdm660.c
@@ -177,7 +177,6 @@ static struct clk_init_data gpu_clks_init[] = {
 */
 
 static const struct freq_tbl ftbl_gfx3d_clk_src[] = {
-       F_GFX( 19200000, 0,  1, 0, 0,         0),
        F_GFX(160000000, 0,  2, 0, 0,  640000000),
        F_GFX(266000000, 0,  2, 0, 0,  532000000),
        F_GFX(370000000, 0,  2, 0, 0,  740000000),
@@ -191,7 +190,6 @@ static const struct freq_tbl ftbl_gfx3d_clk_src[] = {
 };
 
 static const struct freq_tbl ftbl_gfx3d_clk_src_630[] = {
-       F_GFX( 19200000, 0,  1, 0, 0,         0),
        F_GFX(160000000, 0,  2, 0, 0,  640000000),
        F_GFX(240000000, 0,  2, 0, 0,  480000000),
        F_GFX(370000000, 0,  2, 0, 0,  740000000),
```

oshmoun:
Ignore the previous comment, the crash still happens, albeit a bit less frequently.
seems like an issue in `clk-rcg2.c`

kholk:
I've had a fast look at the driver.. I think I know how to solve the issue.
We need a custom clk_set_rate function that will switch the clock parent dynamically.

I think that the GPU on this SoC has some hardware quirk that makes it very sensitive to clock changes, or the PLL is acting weird with dynamic clock changes. 
So either we set the GPU_PLL0 to a high rate and postdivide it to get each rate, which would use some more power, or we switch between PLL0 and PLL1:

if (current_parent == PLL_0 && new_freq_is_in_PLL_ranges) {
clk_set_rate(PLL_1, new_freq);
/* gives time for the PLL to stabilize */
usleep_range(5, 10); /* most probably not necessary, handled in pll's set_rate */
clk_set_parent(GFX3D, PLL_1);
}

And vice versa, if current pll1, set pll0.

oshmoun:
@kholk tried to implement that in a notifier callback that dynamically switches cfg value of P_GPU_PLL0_PLL_OUT_MAIN between 1 and 3
https://github.com/sonyxperiadev/kernel/blob/aosp/LE.UM.2.3.2.r1.4/drivers/clk/qcom/gpucc-sdm660.c#L67
device was still able to boot and function, but did not survive the test of rapidly switching rates

kholk:
Show the codez!

oshmoun:
~the last commit added should help avoid going into the codeblock that throws the error. `halt_check` flags are also not declared in `gpucc-msm8998.c`~

kholk:
Noupe, absolutely. The default halt_check is BRANCH_HALT, so you just did nothing...

PR #1977, opened 2019-05-07
[2.3.2.r1.4] Stabilize SDM630/636/660 GPU

kholk:
This patchset stabilizes the SDM630/660 GPU by correctly setting up the
clocks for it.

Tested on SoMC Ganges Mermaid

P.S.: Do not merge until further notice. This PR is here for other people to test.


oshmoun:
both test devices crashed
tested with and without 
https://github.com/sonyxperiadev/kernel/pull/1977/commits/b0c2f5fb074b21df9c026375eb208f783df64ecc

now trying the clk-alpha-pll change

kholk:
Ye, try it also on top of this PR, as it's doing the pll ping-pong

oshmoun:
done, crashed
EDIT: added some debugging to check if the added code was called. It is not because the pll is always disabled when alpha_pll_set_rate is called, probably because of the ping-pong

kholk:
Force pushed.

Now the proper solution to the GPU crashes should be up.
Needs testing.

oshmoun:
does not work
https://del.dog/xejoqoceqi.htm

HALT_CHECK_SKIP is there for a reason

kholk:
Yes and it's not the same reason as ours, but hell we may just go for it.
One more idea left, then we do it.

oshmoun:
indeed not for the same reason, as the check was done on 4.4 without issue.
it remains unclear what changed between 4.4 and 4.9 to cause a change in hardware behavior like this

kholk:
Oh no what changed is clear to me.
The RPM framework is fucked, that's what changed.

kholk:
@oshmoun has won this round.

The hopefully temporary (working) solution has been pushed and now
the adreno driver will not BUG() the c**p out of it.

This PR can be merged.

P.S.: I hate it.

PR #1981, opened 2019-05-09
[2.3.2.r1.4] Sanitization and small Android Q preparation

kholk:
This (rather huge) patchset introduces some sanitization and improvements
for both qcom drivers and various kernel subsystems.

There are also patches to start supporting basic features of Android Q, which
are left disabled in all devices configurations, so that keeps full compatibility
(like not even having them) with the current releases of Android Pie.

The new (disabled) features for Android Q in this patchset are:
- PSI (Pressure Stall Information Tracking), useful to debug sched pressure statistics
- Android Verity via dm-android-verity (we are currently using dm-verity instead)
- Android AVB Verity
- DM-BOW backport for "Backup On Write" functionality

As already explained, everything else is sanitization and performance improvements.

This patchset shall be _carefully_ tested before merging.
Enjoy!

PR #1982, opened 2019-05-09
[2.3.2.r1.4] [MAY BREAK P] Android Q enablement

kholk:
This is the second part of PR #1981 containing the Binder part of
the Android Q enablement.

This patchset depends on #1981.

This patchset **may break Android P**.

Be careful.


jerpelea:
this is not needed

PR #1983, opened 2019-05-10
[2.3.2.r1.4] MSM8956 CPU PLL stabilization

kholk:
This patchset aims to stabilize the MSM8956(/76) clusters and the HFPLL
by adding some safety mechanisms and tidying up some parts of it.

There are also some cosmetic changes to increase human readability and
an important bugfix for the lock wait loop in the HFPLL driver.

Tested on SoMC Loire Suzu RoW


kholk:
This PR is tested and validated and can be merged.

PR #1984, opened 2019-05-11
clk: gcc-msm8976: declare correct enable mask for clocks

oshmoun:
The clocks gcc_apss_ahb_clk and gcc_apss_axi_clk had wrong enable_mask values
That caused them to wrongfully write to the bit designated for gcc_crypto_ahb_clk


kholk:
Verified, INSTAMERGE

PR #1993, opened 2019-05-24
[MERGEME] fix race condition in may_use_simd

oshmoun:


kholk:
Thanks for pushing, @MarijnS95 is verifying functionality, even though this
patchset should most probably be totally fine.

kholk:
MERGEME

PR #2000, opened 2019-06-04
[LE.UM.2.3.2.r1.4] [SECURITY] perf: core: Avoid race condition when releasing perf-events

ix5:
This is the only part of the 2019-06-05 kernel security patches that we need to apply.
https://source.android.com/security/bulletin/2019-06-01.html#kernel-components


ix5:
Compiles for kagura

PR #2001, opened 2019-06-04
[VERIFYME][2.3.2.r1.4] remove unneeded branch halt skipping

oshmoun:
These patches ended up only replacing the original issue (crashing) with a different one (frozen device with a white screen).
Revert them now that the root cause was handled.


jerpelea:
we may need some testing :)

Thaodan:
Would be great if the commits indicate the reason for this change.

PR #2003, opened 2019-06-07
[4/4] [2.3.2.r1.4] [Tone/Yoshino] Enable clearpad glove_supported.

MarijnS95:
For https://github.com/sonyxperiadev/packages_apps_ExtendedSettings/pull/49.

These features should work just fine now that these devices are brought up on 4.9 (for a while). Given that we have a toggle for it, allow the user to enable glove mode again on Tone/Yoshino.

### DIFF
--- a/arch/arm64/boot/dts/qcom/msm8998-yoshino-common.dtsi
+++ b/arch/arm64/boot/dts/qcom/msm8998-yoshino-common.dtsi
@@ -205,7 +205,7 @@
 			watchdog_delay_ms = <3000>;
 			charger_supported = <0>;
 			pen_supported = <0>;
-			glove_supported = <0>;
+			glove_supported = <1>;
 			cover_supported = <1>;
 			touch_pressure_enabled = <1>;
 			touch_size_enabled = <0>;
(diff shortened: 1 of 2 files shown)


kholk:
Thumbs up!

arch/arm64/boot/dts/qcom/msm8998-yoshino-common.dtsi
```diff
@@ -205,7 +205,7 @@
 			watchdog_delay_ms = <3000>;
 			charger_supported = <0>;
```
lzlrd, line 206: Sorry to bother you, but is this flag for Qnovo?

MarijnS95: Unfortunately I don't know; looking at the driver all this does is allowing to set a bit in the touchscreen. Whether it is to mitigate frequency issues during charging, display a low-power charge value/icon/animation on the screen, or make the display more aware of the battery and adapt power usage to that...

There also seems to be no documentation about clearpad in our kernel.

PR #2004, opened 2019-06-08
block: Do not do I/O stats accounting

tytydraco:
Stats accounting here requires RCU read locks to be held, as well as
disabling CPU preemption. Accounting for I/O statistics increases
overall CPU overhead, making processing block requests costly. Disable
I/O stats accounting by default in the blkdev macro configuration.

### DIFF
--- a/include/linux/blkdev.h
+++ b/include/linux/blkdev.h
@@ -513,13 +513,11 @@ struct request_queue {
 #define QUEUE_FLAG_FAST        27	/* fast block device (e.g. ram based) */
 #define QUEUE_FLAG_INLINECRYPT 28	/* inline encryption support */
 
-#define QUEUE_FLAG_DEFAULT	((1 << QUEUE_FLAG_IO_STAT) |		\
-				 (1 << QUEUE_FLAG_STACKABLE)	|	\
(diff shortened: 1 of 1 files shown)


kholk:
This is absolutely not a good idea. At all.
We wouldn't be able to do debugging in case we need to and we won't be able to analyze the IO stats for optimization phases.
No no no.

tytydraco:
> This is absolutely not a good idea. At all.
> We wouldn't be able to do debugging in case we need to and we won't be able to analyze the IO stats for optimization phases.
> No no no.

Perhaps for releases IO stats becomes disabled, leaving the statistics available to optimization analysis during testing, and reducing CPU overhead during block requests in final releases. Releases should have minimal debugging enabled to minimize wasted CPU cycles for debugging that the end user likely is oblivious of.

kholk:
I agree that reducing overhead during block requests is good, but the kernel 4.9 is
not in a feature freeze state yet and it's still in (well, sort-of) active development:
disabling this will make our life harder if anything happens.

Perhaps this can be considered after we migrate to the upcoming kernel 4.14,
which is currently in the works, so, after k4.9 goes in feature-freeze state.

Scheduled for the feature-freeze.

jerpelea:
this patch can be merged after we move to 4.14 
leave it open and it will be included in the test cases!

PR #2005, opened 2019-06-08
input: Do not use HID events as entropy

tytydraco:
The Linux Kernel uses HID inputs as randomness to add to the entropy
pool. However, the add_input_randomness function disables preemption,
taking processing time away from other tasks when processing multiple
events in quick succession. Remove the function call that attempts to
use input data as randomness.

### DIFF
--- a/drivers/input/input.c
+++ b/drivers/input/input.c
@@ -371,9 +371,6 @@ static void input_handle_event(struct input_dev *dev,
 {
 	int disposition = input_get_disposition(dev, type, code, &value);
 
-	if (disposition != INPUT_IGNORE_EVENT && type != EV_SYN)
-		add_input_randomness(type, code, value);
-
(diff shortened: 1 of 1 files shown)


kholk:
This will degrade performance a lot. The major randomness source is the input.

Otherwise, the kernel would have to use computational power to run an algorithm that actually spits a random number.

Using the touchscreen event gives less overhead, overall.

tytydraco:
> This will degrade performance a lot. The major randomness source is the input.
> 
> Otherwise, the kernel would have to use computational power to run an algorithm that actually spits a random number.
> 
> Using the touchscreen event gives less overhead, overall.

You're right, I didn't consider that. I've been using urandom, the nonblocking random pool, so perhaps that's why I haven't noticed much degraded performance. I'll revert this on my end and close the request. Thank you.

PR #2006, opened 2019-06-08
block: Force requests onto their origin CPU

tytydraco:
There is a high probability that the specific CPU that makes an I/O
request may be repeated. For example, the same CPU may read the same
data multiple times when reading a database. This flag forces requests
to be completed on the same CPU that requested them.

### DIFF
--- a/include/linux/blkdev.h
+++ b/include/linux/blkdev.h
@@ -516,11 +516,13 @@ struct request_queue {
 #define QUEUE_FLAG_DEFAULT	((1 << QUEUE_FLAG_IO_STAT) |		\
 				 (1 << QUEUE_FLAG_STACKABLE)	|	\
 				 (1 << QUEUE_FLAG_SAME_COMP)	|	\
+				 (1 << QUEUE_FLAG_SAME_FORCE)	|	\
 				 (1 << QUEUE_FLAG_ADD_RANDOM))
 
 #define QUEUE_FLAG_MQ_DEFAULT	((1 << QUEUE_FLAG_IO_STAT) |		\
 				 (1 << QUEUE_FLAG_STACKABLE)	|	\
 				 (1 << QUEUE_FLAG_SAME_COMP)	|	\
+				 (1 << QUEUE_FLAG_SAME_FORCE)	|	\
 				 (1 << QUEUE_FLAG_POLL))
 
 static inline void queue_lockdep_assert_held(struct request_queue *q)


kholk:
.... So when we push a big faster core online, we'll be relegated to the slow little core to finish that heavy i/o request?

This degrades performance and doesn't give any power consumption benefit either, because we'd need to keep the little cores at max freq (hence max voltage) and we'd waste way more energy... To complete the request in a slower fashion.....

tytydraco:
> .... So when we push a big faster core online, we'll be relegated to the slow little core to finish that heavy i/o request?
> 
> This degrades performance and doesn't give any power consumption benefit either, because we'd need to keep the little cores at max freq (hence max voltage) and we'd waste way more energy... To complete the request in a slower fashion.....

Keep in mind that the default setting for request affinity is to complete the request on the same cluster of CPUs.

tytydraco:
> .... So when we push a big faster core online, we'll be relegated to the slow little core to finish that heavy i/o request?

Not quite. I did some deeper research to drive my point further. So with the currently used 4.9 Linux Kernel in conjunction with MMC / UFS / (most flash), we use the SQ block. Therefore when inspecting the possibility for performing an IPI, we should be taking a look in the `blk-softirq.c` file.

Let's break down the `__blk_complete_request` function a bit, specifically the part where we set and utilize `ccpu`.

1) Check if the previously called CPU for the request exists
1a) If yes, set `ccpu` to that.
1ai) If we haven't specified `QUEUE_FLAG_SAME_FORCE`, then decide to just use the currently executing CPU instead *as long as it shares the same cache as `req->cpu`*.
1b) If no, execute the block request on the currently executing CPU (`smp_processor_id`)

2) Execute the block request
2a) If we chose `ccpu` to be the same as our currently executing CPU, or we find out that `cpu` shares the same cache (L2) as `req->cpu`, then just run the task locally.
2b) If we picked something besides the currently executing CPU, raise an IPI to complete the request on `ccpu`.
2c) FAILSAFE: If we fail to raise the IPI, fallback to `do_local`.

> This degrades performance and doesn't give any power consumption benefit either, because we'd need to keep the little cores at max freq (hence max voltage) and we'd waste way more energy... To complete the request in a slower fashion.....

We can conclude that even without `QUEUE_FLAG_SAME_FORCE` specified in the block flags, we're defaulting to a CPU in the same cluster as the previously executed request CPU. We wouldn't ever be forcing the block request to be executed on a little CPU if the origin request was on a big CPU.

With `QUEUE_FLAG_SAME_FORCE` specified, we ignore the possibility of targeting an L2 cache over an L1 cache. L1 cache can be hit with 2-3 CPU cycles, whereas L2 cache is far slower at 8-10 CPU cycles. It is likely the CPU will still be cache hot from the previous request as block requests occur in rapid succession.

In conclusion,
1) We won't be wasting any additional power.
2) We won't be locked to a slower CPU if a faster CPU is available.

I hope this clears things up for you a bit. I await your response.

tytydraco:
@kholk I did some benchmarks because I really want to stress why this change can be beneficial.

https://docs.google.com/spreadsheets/d/1UMJQ5-KlsZcgLjNeTqMzyZGXTTpLZVZbU4IQr8M3aEw/edit?usp=sharing

Here's an IOZone benchmark with a 1024 KB file size maximum. It is clear that an `rq_affinity` of 2 rather than 1 offers significant re-reader improvement (aka cache hits). The benchmarks are in microseconds.

kholk:
Hey. I'm sorry I've been a lil busy.

This is becoming reasonable enough, it seems, I LOVE when people are providing numbers. 

Though, two or three things:
- First of all, what are the column numbers referring to? It's not totally understandable.
- A 1MB read is the least likely thing to happen on any device: for like 75% of the times we hit very small random reads and writes (as small as the block size): this happens not only when loading the OS, but also when using basically any app, due to ldopen here and there, so you should really provide numbers for this very important case, and evaluating all the most common other cases would be really appreciated.
- Hitting the L1 means... Exactly what it means. Usually, we want the L1 to be mostly full of instructions instead of being full of random data that's read from any storage. Did you actually consider that? Running a mixed workload benchmark would be a good idea here (of course, not the apps on android, choose something decent on the command line, since you're trying to get very sensible numbers). 

And another last thing.
Yes you've provided some numbers, but... Where do they come from? Where was the read done? What SoC was that? What kind of storage was that? What device is it?!
Also, where is the writes performance?

tytydraco:
> Hey. I'm sorry I've been a lil busy.

No problem at all. I'm happy to hear back from you.

> First of all, what are the column numbers referring to? It's not totally understandable.

I went through and cleared everything up on the original sheet. Link [here](https://docs.google.com/spreadsheets/d/1UMJQ5-KlsZcgLjNeTqMzyZGXTTpLZVZbU4IQr8M3aEw/edit?usp=sharing) if you need it again.

* The content is measured in microseconds, therefore, less is better.
* The test type is displayed on the left column. It specifies a prefix with the `rq_affinity`. I.E. `[1] Writer` is the IOZone Writer test on an `rq_affinity` of 1.
* The other columns specify the record size. It includes 4 KiB, 8 KiB, 16 KiB, 32 KiB, and 64 KiB.
* The last column is the percent change of the test averaged across all record sizes. A negative change implies a latency reduction, while a positive change implies a latency increase.

> A 1MB read is the least likely thing to happen on any device.

I 100% agree with you here. I went ahead and did more reasonable record sizes (as seen above).

> Usually, we want the L1 to be mostly full of instructions instead of being full of random data that's read from any storage.

Modern devices separate the L1 CPU cache into L1i and L1d, where the former holds instructions and the latter holds data. This doesn't seem to be an issue for us. See [this](https://en.wikipedia.org/wiki/CPU_cache) for a brief rundown.

> Yes you've provided some numbers, but... Where do they come from? Where was the read done? What SoC was that? What kind of storage was that? What device is it?!

I'm so sorry for not specifying. I made a part in the sheet for this. Here it is in plain-text.

```Manufacturer | Xiaomi
Device | Mi A2
Device ID | jasmine
Android Version | 9.0-GSI-20190604 June 5, 2019
Android Status | STOPPED
Kernel | 4.4.182-Hex #794 Jun 21 2019
Storage Type | eMMC
Storage Capacity | 64 GB
```

> Also, where is the writes performance?

I didn't think that writes would be affected much since we're assuming that forcing the block request to it's origin CPU is meant to gain more cache hits, improving read performance. Writes aren't affected much from cache locality in the same context that reads are. Though, on a second look, IOZone offers a Re-writer test, which benchmarks the processing time of writing the same data twice, which MAY gain benefits from a cache hit. I included more thorough results in the new sheet.

I hope this clears things up. Please get back to me as soon as you can, and once again, thank you for the reply.

kholk:
There was no need to link me to that very very incomplete wannabe explaination of CPU caches,
I definitely know this stuff to perfection, don't worry. The problem here is that you've probably
misunderstood what I was meaning with "we want it fullfilled with instructions, not with data"
which, again briefly, *optimally* the L1D has to be filled with data that is useful to the pipeline,
not with "random data", but that's not really an issue that happens _because of_ your changes.

"words apart" .........

- It's interesting to see how the writer test behaves. I wonder if there's any underlying issue in
the EXT4 driver or in the block layer.

- Your tests are clearly showing that there is a _huge_ loss (despite the averages showing a -1%
gain) on random reads and random writes, which are basically again 90% of the "disk" activity
in an entire boot life for an OS like Android. Of course, 16 and 32KiB records are generally
seeing big improvements, but these cases aren't hit a lot of times in the random R/W case.

- Your tests are also showing that the re-writer (not too much important, realistically) and the
reader cases are (excuse me for being crude) _royally fucked_ at 4, 8 and 64 KiB record sizes,
which is a kind of "ARE YOU SERIOUS?!?!?!" situation, as 4, 8 and 16 are the ones that we
do hit the most in the boot live.

- I don't know how "true" are two of your results there (sure, that's not intentional, I know),
especially the [2]Reader 32KiB and the [2]Re-reader 16KiB .... you will agree with me that
these values do really appear A LOT strange compared to the rest and for this reason I have
ignored them (but even if I didn't it wouldn't change what I said) for my analysis of your results.


Finally, I think that your conclusions, quoted from your (really nice, love it) spreadsheet
> Placing block requests on their origin CPU significantly reduces task processing latency on average with requests that are <= 64 KiB

are *totally wrong* and I totally disagree with you.

My conclusions on this:
Forcing to place block requests on their origin CPU, whether it's too busy (hence having to migrate other processes to other CPUs, which costs a lot), or "not too busy" (but not idling, as this is _basically_ never the case) is very detrimental on I/O performance for all cases on 4KiB record
size, which is something that gets hit a lot and also almost generally bad for 8KiB record size.

9/10 of the most common cases where we need the most performance are seeing performance
regressions, while 9/10 of the most uncommon cases where we can safely ignore a [ sometimes
small, other times medium, but never high ] performance loss have got good performance
improvements, including the safely ignorable write performance, since we write to the OS disk
(on Android, the data partition) for around 5% of the entire OS lifetime.

In conclusion, this change -  as shown by your test - will be generally detrimental for the disk
performance and this will lead to a general responsivity drop of the entire system.

tytydraco:
> There was no need to link me to that very very incomplete wannabe explanation of CPU caches,
I definitely know this stuff to perfection, don't worry.

Your last post was as follows:
> Usually, we want the L1 to be mostly full of instructions instead of being full of random data that's read from any storage. Did you actually consider that? 

It doesn't seem like you "know this stuff to perfection," as you clearly implied that L1 shared it's cache with instructions and data. Obviously the Wiki link I sent you is incomplete, as is most Wiki links. I shared it with you to inform you, rather than to insult you. We won't get anywhere if you cannot to look into related subsystems. Recall previously that you didn't understand how `blk-softirq.c` was only able to select CPUs in the same L2 cache pool. This is just an example, but I strongly, **strongly** urge you to dig deeper and truly investigate related parts of the hardware and software if we want to stay on the same page.

> ... optimally the L1D has to be filled with data that is useful to the pipeline,
not with "random data"

The L1 cache expires and is invalidated extremely quickly with a size in the Kilobytes range. We're not filling it with "random data." In fact, we're not filling it up at all with the addition of this pull request. I'm not quite sure where you got this... The L1 gets filled as data is processed by the CPU, so it's not in our control what enters and exits the L1 cache. We're simply targeting data that _already_ exists in the L1 cache, betting on a possible cache hit to make processing latency exponentially lower. Additionally, are you implying that block-related data isn't "useful to the pipeline?" I hope you realize that hardware I/O is one of the slowest subsystems apart from SPI and a few others, placing it as one of our largest bottlenecks. It is **crucial** that we expedite I/O _as much as possible_ if we want to see any noticeable improvements in userspace. Once again, I strongly urge you to take a peek at that Wikipedia page. It contains a very brief explanation on most of the stuff I'm talking about and I think you will find it beneficial.

> Your tests are clearly showing that there is a huge loss (despite the averages showing a -1%
gain) on random reads and random writes, which are basically again 90% of the "disk" activity
in an entire boot life for an OS like Android.

Please note that write requests are almost _always_ non-blocking. Nothing is pending on the completion of the write request. However, what _is_ blocking are most read requests. That's the real concern here, as that's where we're going to visually perceive our changes in userspace.

> Your tests are also showing that the re-writer (not too much important, realistically) and the
reader cases are (excuse me for being crude) royally fucked at 4, 8 and 64 KiB record sizes ...

Allow me to explain. My device uses Micron eMMC. It is a widely known issue in our device that random writes are extremely unstable, causing I/O freezing for upwards of seconds. See [this thread](https://forum.xda-developers.com/mi-a2/how-to/slow-emmc-isnt-software-bug-t3871668) for additional information about that. With that said, I ask you to entirely disregard write performance in this test. The testing device is unstable for this test. I will re-run the same exact test on my Mi A1 (tissot), which does not exhibit unstable eMMC performance, and get back to you here.

> I don't know how "true" are two of your results there (sure, that's not intentional, I know),
especially the [2]Reader 32KiB and the [2]Re-reader 16KiB ...

Please note that userspace I/O benchmarks are highly variable. Sure, the 4 KiB Reader test shows over a 400 Ms completion latency addition, but subsequent tests marginally change. However, what we should be focusing on are examples like the 32 KiB Reader test. These are examples of times where the L1 cache hit succeeded, leading to almost a 2 second (yes, _second_) latency reduction. Do not expect these tests to be accurate to such a degree, as they will never be. I/O is constantly occurring, even with Android stopped. We can't get stuck on marginal changes.

> Forcing to place block requests on their origin CPU, whether it's too busy (hence having to migrate other processes to other CPUs, which costs a lot), or "not too busy" (but not idling, as this is basically never the case) ...

An IPI does not migrate. An IPI is appended to a CPU's IPI FIFO list, to be executed on the CPU as soon as possible. As for the CPU being idle, this is not at all rare. By definition, a CPU is idle when `nr_running` on said CPU is zero. Please investigate `/proc/loadavg/` (more info [here](https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/4/html/Reference_Guide/s2-proc-loadavg.html)). The fourth column displays the currently running tasks for the entire system (`nr_running` for all CPUs). I did a test by accumulating the currently running tasks on a device during a standard workload of opening the Play Store and closing it again. See the results [here](https://del.dog/izecoromur.sh).

Notice how `nr_running` is almost _always_ <= `NR_CPUS` (8 in my case). On average, the device is running almost 3 tasks at any given moment (if we round up). The chances that our `req->prev_cpu` will be available and idle is more than likely. In fact, I have a pull request that ensures this case is always true (I've accounted for it): [here](https://github.com/sonyxperiadev/kernel/pull/2013). 

> 9/10 of the most common cases where we need the most performance are seeing performance
regressions...

I'm not quite sure where you got this. Latency sensitive requests are almost _always_ synchronous read requests. You will almost __never__ see a synchronous write request. Like I stated previously, we see examples where Re-reader and Reader strike cache hits and execute exponentially sooner than they would have without `rq_affinity` set to 2. You're hung up on marginal (and yes, I do say marginal) changes that are subject to variability depending on the benchmark environment.

In conclusion, the benchmark results are inaccurate due to the testing device and the test must be redone on a trusted device. Feel free to run this test yourself. The spreadsheet contains the command used to execute the test.

EDIT: I had Kuran Kuname (@celtare21) run the same benchmark under the same conditions on his device. I added his results to the Google Sheets under a new page titled "mata." This device is reported to have no I/O quirks like `jasmine` did. Take a look when you can.

```
Manufacturer | Essential
Device | Essential Phone
Device ID | mata
Android Version | 9.0-AICP-20190620 June 5, 2019
Android Status | STOPPED
Kernel | 4.4.183-Artemis #98 Jun 22 2019
Storage Type | UFS 2.1
Storage Capacity | 128 GB

kholk:
As I wrote already, you posting that link wasn't taken as an offense, but rather as a
misunderstandment of my too fast and poorly written concept in my first explaination.

Your last answer though, is touching the limit of the offense and I will not argue with
you about the CPU caches, how they're hit, why and what do I precisely mean, as
now it looks like you're misunderstanding my words _on purpose_, which is a
whole different thing, but hey! that's still okay, it's fun to discuss sometimes, but I'm
living pretty busy weeks so I can't go on with useless topics. Perhaps another time!

About the write speeds thing, please note that I've previously said that this is not
something to really care about in this topic, as all the other - more important - tests
are showing an important performance drop in important areas.
Your Micron eMMC is fine, you just have sort of unreliable write results with that
part, but again, my issue is on _reads_, not _writes_.

I have already politely told you that your changes are detrimental to the performance
and I was already expecting that before you pushed the tests (which, again, are
clearly showing what I've already said now more than once).

Just accept your own numbers: everyone of us makes mistakes every now and then,
it's not a tragedy, you're not gonna be buried in hell for this. Take it easy man!


That said, I'm sorry, but **this change is detrimental for performance and it is definitely
not mergeable in this project**.

Have a nice evening,
Angelo

tytydraco:
@kholk I'm so sorry to reopen this but you've gotta see this.

A few things:
1) My first test included a problematic commit in the kernel: https://github.com/sonyxperiadev/kernel/pull/2013/commits/3673809d800a1ec0aaa325a052dcb983af09b332

This totally invalidated all our tests since basically it didn't even test our change at all here most of the time (see previous post for why).

2) The IOZone commands I used proved high deviation and highly inaccurate results. I went and took a deep dive into the man page and chose a more accurate benchmark command.

`./iozone-static -a -s 16 -S 64 -b [file].xls`
-a: Run all tests
-s: Use a 16 KiB file size to test
-S: Align cache to 64 B (my cache line size)
-b: Output results to file

So why is this new configuration more valid? We need to tell IOZone our cache line size for buffer alignment and proper purging:
`Set processor cache size to value (in bytes). This tells Iozone the size of the processor cache. It is used internally for buffer alignment and for the purge functionality.`

Also, test only up to 64 KiB since that's more realistic of a test in actual usage.

The new results were FAR more stable. Deviation was far more consistent compared to the previous test, in which running the same test twice could result in massive outliers. I redid the entire spreadsheet and the results are absolutely staggering. I had @celtare21 test once again with the new command under the same circumstances and he get similar results.

Please take another look. I think these new results warrant another look:
https://docs.google.com/spreadsheets/d/1UMJQ5-KlsZcgLjNeTqMzyZGXTTpLZVZbU4IQr8M3aEw/edit?usp=sharing

tytydraco:
If you want to actually chat, PM my Telegram (@tytydraco) and we can talk. We can discuss potential concerns through there instead of relying on information from other people. Thanks.

jagogardiner:
I also decided to test this commit, with the command
`echo 3 > /proc/sys/vm/drop_caches`
ran beforehand.

These are my results, please take a look:
https://docs.google.com/spreadsheets/d/1xIB8qx7M20rZ2rzJmSb_lL28rqKQQz6xp4ftdPzSSo0/edit?usp=sharing

jerpelea:
@tytydraco the tests show that there is a small performance increase but the test pool is too limited and we should also consider the power measurements since we are on a mobile device with a limited power source.

PR #2007, opened 2019-06-12
LA.UM.6.4.r1: input: ts: synaptics_dsx_td4322: fix display panel detection.

krnlyng:
When flashing Android 9 to XA2 pioneer devices and afterwards flashing AOSP the
kernel will flash the wrong firmware due to incorrect panel detection resulting
in garbage to be displayed. This uses the detected panel information from
mdss_dsi to flash the correct firmware.

Together with the downgrade patch this will hopefully fix the remaining panel issues.

### DIFF
--- a/drivers/input/touchscreen/synaptics_dsx_td4322/synaptics_dsx_core.c
+++ b/drivers/input/touchscreen/synaptics_dsx_td4322/synaptics_dsx_core.c
@@ -4122,26 +4122,20 @@ static int synaptics_rmi4_probe(struct platform_device *pdev)
 
 	/* tp_source default set to 0xFF (unknow) */
 	rmi4_data->tp_source = TP_SOURCE_UNKNOW;
-	rmi4_data->project_id = 0x00;
 
 	/* For TouchView TD4322 firmware upgrade -and repair- */
-	if (of_machine_is_compatible("somc,nile")) {
-		if (of_machine_is_compatible("somc,pioneer")) {
-			rmi4_data->project_id = 0x01;
-			if (rmi4_data->lcd_id == 1)
-				rmi4_data->tp_source = TP_SOURCE_TRULY;
-			else if (rmi4_data->lcd_id == 0)
-				rmi4_data->tp_source = TP_SOURCE_CSOT;
-		} else if (of_machine_is_compatible("somc,discovery")) {
-			rmi4_data->project_id = 0x02;
-			rmi4_data->tp_source = TP_SOURCE_INX;
-		}
+	if (strstr(saved_command_line, "qcom,mdss_dsi_td4322_csot_fhd_cmd") != NULL) {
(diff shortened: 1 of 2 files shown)


drivers/input/touchscreen/synaptics_dsx_td4322/synaptics_dsx_core.c
```diff
@@ -4125,17 +4128,12 @@ static int synaptics_rmi4_probe(struct platform_device *pdev)
...
-		}
+	if (!strcmp(mdss_panel_name, "qcom,mdss_dsi_td4322_csot_fhd_cmd")) {
+		rmi4_data->tp_source = TP_SOURCE_CSOT;
+	} else if (!strcmp(mdss_panel_name, "qcom,mdss_dsi_td4322_innolux_fhd_cmd")) {
+		rmi4_data->tp_source = TP_SOURCE_INX;
+	} else if (!strcmp(mdss_panel_name, "qcom,mdss_dsi_td4322_truly_fhd_cmd")) {
+		rmi4_data->tp_source = TP_SOURCE_TRULY;
 	}
 
 	TP_LOGI("rmi4_data->project_id = 0x%02X, "
```
sledges: rmi4_data->project_id will always be zero as its setting above has been removed

kholk:
I'm sorry, but even if it works, this is totally incorrect.

First of all, exporting a variable from the fbdev to the input is... C'mon.
Second thing, this commit, even if I would be okay with such a hack, is already obsolete and the problem that you describe will show up on kernel 4.9 and 4.14 because your hack takes care of the obsolete fbdev driver only.

The best solution would be to read some bits (if possible, I didn't research on this topic) from the controller in the dsx driver itself without interactions with other subsystems.

Please note that we have switched from the obsolete fbdev driver to the new DRM SDE on kernel 4.9 and, even if it's still on early development, on kernel 4.14.

kholk:
Sorry for the double comment, but something just came to my mind and it's most probably the easiest approach.

1. Nile (and Ganges?) should have one possible panel per machine as far as I remember (!)
2. Nile and Ganges bootloader are injecting the panel name in the command line params before booting the kernel: parsing it would give you exactly what you want. This doesn't change between drivers because it's always the same bootloader!

krnlyng:
I've tried using both 1) module_param_string with #define MODULE_PARAM_PREFIX "mdss_mdp." and 2) __setup for reading the parameter from the command line but this doesn't work because apparently kernel command line parameters can only be read once (and the mdss_mdp code already reads it).

I am not parsing the entire command line again, but i only use strstr to check whether the panel type is specified, i hope this is acceptable.

kholk:
That's not optimal but - of course - it's the only solution to a big issue.
Thanks for the proper code!

This PR can be merged.
P.S.: Remember to push this commit also on k4.9.

PR #2008, opened 2019-06-12
[2.3.2.1.r1.4] arm64: DT: Tone: Fix custom mapping for PM(I)8994 MPPs

ix5:
Backport from Angelo's 7.1 branch

During porting of the MPPs to the new pinctrl scheme a long time ago a bad typo happened. All the pins were ... bad.


kholk:
This is a critical fix. MERGEME

PR #2009, opened 2019-06-14
[LE.UM.2.3.2.r1.4]: input: ts: synaptics_dsx_td4322: fix display panel detection and allow firmware downgrading.

krnlyng:
As per request also making a PR to 4.9. I haven't tested this yet beyond cherry-picking (only tested on 4.4).


kholk:
I've requested the PR. Merge!

PR #2010, opened 2019-06-14
[LE.UM.2.3.2.r1.4] [3/3] Change et51x voltage to 3.3v.

MarijnS95:
For https://github.com/sonyxperiadev/vendor-sony-oss-fingerprint/pull/49.

Ganges requires 3.3v on the sensor or it'll misbehave most of the time. Nile sensors operate on this voltage as well, but seemed to behave perfectly fine on the lower 1.8v used for FPC.
In the end, it turns out this lower voltage is what required hacks/quirks like `enable-on-boot`, which have been removed now.
Still, to err on the safe side, initialize to 1.8v by default on platforms that can have such lower-voltage sensors. This can be disabled in DT (as is the case for Ganges), or the voltage is upped after detecting the right sensor.

As with Nile, we've seen variations in HW response (https://github.com/sonyxperiadev/vendor-sony-oss-fingerprint/commit/375934d2edb2f70f57c9ec438b8e2f2ecbff2094), which is why this needs to be tested on all relevant platforms by someone other than me:
- [ ] Nile with FPC
- [ ] Nile with Egistec
- [ ] Ganges

Keep in mind that this includes commits from a [**closed** pull request](https://github.com/sonyxperiadev/kernel/pull/1950): I don't know if this is the right way to add the heap, apart from it seemingly working fine.

### DIFF
--- a/arch/arm64/boot/dts/qcom/sdm630-ganges-common.dtsi
+++ b/arch/arm64/boot/dts/qcom/sdm630-ganges-common.dtsi
@@ -80,6 +80,7 @@
 		et51x,gpio_rst = <&tlmm 20 0>;
 		et51x,gpio_irq = <&tlmm 72 0>;
 		vdd_ana-supply = <&pm660l_l6>;
+		et51x,no-low-voltage-probe;
 
 		spi-max-frequency = <1000000>;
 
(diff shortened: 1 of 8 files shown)


kholk:
I have reviewed the driver voltage commits in private with him while in the making and they're fine.

About the TA Heap thing that I had refused the last time because of my doubt about the firmware doing something, I've had to examine that situation for k4.14, since there it's strictly needed, and there's no problem since that thing is being used only on the HLOS.

The devices will not stop working with older blobs because that won't conflict in any way, the old ones will just ignore the existance of the new ION heap for the QSEE TA.


From my side, the PR looks totally safe.

Good job :)

jerpelea:
Nile looks good!

PR #2013, opened 2019-06-17
block: Do not wake the request CPU if idle

tytydraco:
Refrain from waking up an idle CPU if possible since the exit latency
of taking req->cpu out of an idle cstate will likely exceed the
rq->deadline constraint compared to executing the request locally
instead.


tytydraco:
In hindsight, this is a terrible idea. After doing research for https://github.com/sonyxperiadev/kernel/pull/2006, I've determined that the request CPU is likely idle at any given moment. This commit would damage block performance. Closing the request.

PR #2014, opened 2019-06-17
[2.3.2.r1.4] Linux 4.9.182

nathanchance:


nathanchance:
4.9.182 was just pushed to address this advisory: https://seclists.org/oss-sec/2019/q2/182

kholk:
I don't know how the 4.9.178 patch was tested, because it did introduce a regression on
the debug-monitors that f#*$ed up the processors at the SMP boot stage, producing
likely (and really happening) FAULTS and REBOOTS while heavily using the device
due to a race condition that is happening in the firmware (and should be in the TEE
CPU setup stage).

This is mostly seen on SDM845 Tama, but I'm sure that using other SoCs will trigger
the same issue: you just have to use them so that one of the little cores gets plugged
in while the last core of the big cluster goes down. This creates HAVOC.

So PLEASE guys test the mainline patches more carefully before saying
that you don't notice regressions.
Of course, this is not @nathanchance 's fault. By the way, thank you for the continuative
mainline updates pushing, Nathan.

The message was for the team.

By the way, please MERGE THIS INSTANTLY, because it solves the pretty bad
regression that has been introduced in Linux 4.9.178!!!!

PR #2015, opened 2019-06-17
[6.4.r1] Linux 4.4.182

nathanchance:


nathanchance:
4.4.182 was just pushed to address this advisory: https://seclists.org/oss-sec/2019/q2/182

PR #2016, opened 2019-06-18
[2.3.2.r1.4] Backport new vmalloc for "large performance benefits"

kholk:
This is a backport from Linux 5.2-rc1 of a patch series to greatly enhance vmalloc's performance
especially on embedded systems, plus all of its dependencies that were missing in kernel 4.9.

For all the informations, refer to LKML: https://lkml.org/lkml/2018/10/19/786

Brief informations:
Currently an allocation of the new VA area is done over busy list iteration until a suitable hole is found between two busy areas. Therefore each new allocation causes the list being grown. Due to long list and different permissive parameters an allocation can take a long time on embedded devices(milliseconds). 

This patch organizes the vmalloc memory layout into free areas of the VMALLOC_START-VMALLOC_END range. It uses a red-black tree that keeps blocks sorted by their offsets in pair with linked list keeping the free space in order of increasing addresses.

Quote Phoronix:
With this patch from Uladzislau Rezki, calling vmalloc() can take up to 67% less time compared to the behavior on Linux 5.1 and prior, at least with tests done by the developer under QEMU. 

--------------------------------------------------------------------------------------

Personal tests are showing that the device is more responsive when memory pressure is
high and when huge allocations are to be done, it's also noticeably faster in this case, like
when starting Chrome with more than 100 opened tabs after a system reboot (so, an
uncached complete load of it).

Please, test this patchset on more platforms.
This patchset should get merged in _7 days_ starting from the current date (18/06/19).

Tested on SoMC Tama Akatsuki RoW.


Paulbouchara:
I had a great improvement on SoMC Yoshino Maple RoW and Tone Kagura RoW (which is the slowest one of my devices) it got a very nice boost.
10/10 would merge again

MartinX3:
Testing on akari now together with the .182 pull request against my softreboots.
I feel the same UI speed got by forcing all cores always to be online in rqbalance (of course deactivated for this test).
10/10 would merge again (until now)

dev-harsh1998:
@kenny3fcb, Try it out!

lzlrd:
Working nicely on maple, used alongside my Adoptable patch.

kholk:
The testing period for this PR has ended.
The code is working as expected. 

MERGEME!

PR #2017, opened 2019-06-18
[LE.UM.2.3.2.r1.4] [2/2] et51x/fpc1145: Fix unbalanced poll_wait and race conditions

MarijnS95:
For https://github.com/sonyxperiadev/vendor-sony-oss-fingerprint/pull/51

Bailing out early was a wrongly assumed optimization during very early
development. Every poll attempt needs to register itself using
poll_wait, to keep internal accounting in check.

This problem manifests itself specifically on et51x devices (though the
fpc1145 driver is identical) because it starts with the IRQ line in
triggered (low) state. As soon as the misc device is opened, poll is
invoked and returned early without calling poll_wait(), leaving the
mechanism in an unbalanced state. Consequent polling attempts from
userpsace will - despite calling poll_wait() properly - not wake up due
to an unbalanced number of waiters.

Tested on:
SoMC Loire Suzu RoW
SoMC Nile Discovery RoW with Egistec sensor
SoMC Tama Akatsuki RoW
SoMC Ganges Mermaid RoW

### DIFF
--- a/drivers/input/misc/et51x.c
+++ b/drivers/input/misc/et51x.c
@@ -365,28 +365,20 @@ static unsigned int et51x_poll_interrupt(struct file *fp,
 	struct et51x_data *et51x = to_et51x_data(fp);
 	struct device *dev = et51x->dev;
 
+	/* Add current file to the waiting list */
+	poll_wait(fp, &et51x->irq_evt, wait);
+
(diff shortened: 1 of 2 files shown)


kholk:
Already reviewed in private while in the making.

MERGE!

PR #2018, opened 2019-06-22
[2.3.2.r1.4] arm64: DT: sdm660-vidc: reduce number of iommu groups

oshmoun:
The number of iommu register groups assigned to vidc is too high.
The maximum number of register groups available at IOMMU init time is
reached.
No more groups can be registered, and that causes other devices to fail.
One such device is jpeg_dma, which renders the camera nonfunctional.
Specifically, the snapshot capability in opencamera does not work.
Removing two iommus from vidc dt makes jpeg_dma able to register.


kholk:
I've requested this change. Solves unability to perform snapshot with camera.

MERGEME.

PR #2019, opened 2019-06-26
[2.3.2.r1.4] Improve ION driver for low-memory device configurations

artem:
This PR makes ION cached pools movable and removes usage of kmalloc for big allocations. These changes aim to reduce memory fragmentation which experience devices with insufficient (nowadays) amount of RAM (let's say, <6GB).

Quoting Vitaly Vul from his LKML message:

> We see that memory fragmentation increases sharply over time when the camera is used a lot. Neither background compaction nor user-triggered one help it that much:
>
```# cat /proc/buddyinfo
Node 0, zone DMA 4366 2404 6 1 0 1 1 1 0 0 0
Node 0, zone Normal 2736 2094 91 21 8 0 1 1 0 0 0
# echo 1 > /proc/sys/vm/compact_memory
# cat /proc/buddyinfo
Node 0, zone DMA 2591 1955 391 113 13 2 1 1 0 0 0
Node 0, zone Normal 2159 1155 487 158 9 0 1 1 0 0 0
```
>With the patch, triggering compaction gives far better results:
>
```# cat /proc/buddyinfo
Node 0, zone DMA 3101 4845 605 0 0 0 0 0 0 0 0
Node 0, zone Normal 5271 4459 573 20 3 2 3 3 1 0 0
# echo 1 > /proc/sys/vm/compact_memory
# cat /proc/buddyinfo
Node 0, zone DMA 579 303 151 114 91 67 30 24 7 2 1
Node 0, zone Normal 762 508 400 271 195 100 45 10 5 0 0
```

Also, PR includes fix for a possible memory leak

### DIFF
--- a/drivers/staging/android/ion/ion_cma_heap.c
+++ b/drivers/staging/android/ion/ion_cma_heap.c
@@ -107,7 +107,7 @@ static int ion_cma_allocate(struct ion_heap *heap, struct ion_buffer *buffer,
 
 	info->table = kmalloc(sizeof(struct sg_table), GFP_KERNEL);
 	if (!info->table)
-		goto err;
+		goto free_mem;
 
 	info->is_cached = ION_IS_CACHED(flags);
 
@@ -123,6 +123,12 @@ static int ion_cma_allocate(struct ion_heap *heap, struct ion_buffer *buffer,
 	buffer->priv_virt = info;
 	return 0;
 
+free_mem:
+	if (!ION_IS_CACHED(flags))
+		dma_free_writecombine(dev, len, info->cpu_addr, info->handle);
+	else
+		dma_free_attrs(dev, len, info->cpu_addr, info->handle,
+			       DMA_ATTR_FORCE_COHERENT);
 err:
 	kfree(info);
 	return ION_CMA_ALLOCATE_FAILED;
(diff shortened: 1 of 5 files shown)


artem:
Has been tested on blueline for several months, no regressions spotted

lzlrd:
Has been tested on maple, no regressions spotted.

kholk:
I agree about the first patch and the last one, but I've got a lot of doubts generally about the entire PR (well, not the memory leak fix).

The problems are, in short:
- The first patch was a RFC in mainline and it got abandoned, apparently. If it was a good idea, probably the upstream guys would merge it. It's not clear how it went though, because there wasn't a clear "no no no" reply nor a merge.
- The second patch is veeery dubious: using vmalloc is going to create a performance hit. As you should know, kmalloc has much more performance than vmalloc for a number of reasons. 

So, in case, I totally disagree about the second patch, as we have devices with old SoCs here (and slower RAM)... And anyway on machines using 8998/845 we have plenty of RAM, for which this trick wouldn't be worth it as the vm api will anyway find something free for us.

And again, about the first patch.... I'm making the same request that was done on the mailing lists: I'd like to see some numbers. Allocation performance numbers, not compaction results.

jerpelea:
lets stay on the safe side since the patches can't be properly tested

PR #2021, opened 2019-06-29
[LE.UM.2.3.2.r1.4] Backport 4.14 SDE improvements

MarijnS95:
Backport minor improvements in the SDE driver from @kholk


kholk:
The backports were _successfully_ tested by Marijn on Nile, so merging them there should be totally safe.

PR #2026, opened 2019-07-10
[7.1.r1] Enable SMMU deep prefetch errata for SDM845 on msm-4.14 source

artem:
969cdfb: Compilation fix
08bc31c: SoCs like sm6150 are not affected by that errata, so QCOM made w/a optional. SDM845 IS affected, so enable it explictly.
c7dc624: "efficiency" property is obsolete in k4.14, rename it to "capacity-dmips-mhz"
edfabae: Just a cleanup, no functional change

### DIFF
--- a/arch/arm64/boot/dts/qcom/msm-arm-smmu-sdm845.dtsi
+++ b/arch/arm64/boot/dts/qcom/msm-arm-smmu-sdm845.dtsi
@@ -62,6 +62,7 @@
 		qcom,use-3-lvl-tables;
 		qcom,no-asid-retention;
 		qcom,disable-atos;
+		qcom,min-iova-align;
 		#global-interrupts = <1>;
 		#size-cells = <1>;
 		#address-cells = <1>;


artem:
k4.14 also has 8ec5666fb11a4836b5d180bb460cc7c37ea39767 and 7d4ac74fdea7ea850fad47695e437417d12ad81b compared to k4.9 driver, but I'm not sure if they are critical and what to do in this case

kholk:
2daf3c62e1be00635e8090b0af4dced92b344180 is wrong.
Do not ifdef. Use a DT property or check if the "reg" property in DT is != 0xC to get done what you wanted to.

08bc31cb0f82a4bc864d6bee2abffa6579d2958b lots of doubts.
On kernel 4.9 it's not really like that.

c7dc624f01096791bdb3048bfa8163c3563a3395 fine, should be done on all others as well.

edfabaecece49c75930cb32f32a6b2d9ffa14c5c leave the properties.
If anything happens and the machine crashes for some R/W around that iospace, at least
I will notice that there's something filthy about it.

969cdfb69fe60b15f89546bd0bca9f84c6c36e0e bad title and no description, it's a typo fix.

artem:
> 08bc31c lots of doubts.
> On kernel 4.9 it's not really like that.

If I'm not mistaken, qcom enables that errata for MMU-500 regardless on k4.9. See d49c4e9f74a6a0b1bf9de148ed0d5f76aa866682

artem:
force-pushed

kholk:
Erm.. You didn't compile test it, did you....

DT ; ; ; ; :P 

BTW, on an afterthought, I have to check why did they apply this limitation on SM8150. It may be just a smart way of doing something right for both... I'll check as soon as I reach Tama. I'm basically done with Yoshino, so maybe tomorrow, but that's not a promise...

In the meanwhile, if you separate the typo fix for the operating points and PR it separately we can start merging that other one, which is basically urgent!

kholk:
bad commits and others forcepushed out

the SDM845 MMU500 errata enablement is the only one left and it's CORRECT.

MERGE

PR #2027, opened 2019-07-10
[7.1.r1] Rename efficiency property to capacity-dmips-mhz

artem:
"efficiency" property is obsolete in k4.14, rename it to "capacity-dmips-mhz"


kholk:
Yes. This is fine. Thanks!

PR #2033, opened 2019-07-15
[2.3.2.r1.4] f2fs: Fix compilation of f2fs after vmalloc changes

mlehtima:
Partially reverts previous changes to match upstream kernel.


kholk:
Was reverted upstream as well.
Go for it.

PR #2035, opened 2019-07-20
[7.1.r1] SoMC Kumano implementation

kholk:
This is an early implementation of the SoMC Kumano platform,
including the SoMC Kumano Griffin device.

Compile tested. Not boot tested.


Paulbouchara:
"Kumano Ganges" typo ?

kholk:
LOOOOOOOL! Typo corrected :D

Thaodan:
Is there any info on anything that is missing to support Kumano?
Should I fill a bug report on this at bug_tracker?

kholk:
Yes, the somc_panel driver still misses to handle a gpio for that new panel (because I want to do that differently.....) and the audio enhancements for this platform are not ported yet. Some of them are not even necessary, but anyway, noone has tested booting it yet because userspace support is still missing.

Kernel side, however, should be fine, apart the things I just said. In any case I couldn't go on with these things without an userspace to reliably boot so... Don't worry, the platform will be up in a while.
In this moment, since it's still in a sort of early porting stage, it's useless to use the bug tracker as you would likely flood it :D

Oh, thanks for the interest, anyway!

Thaodan:
So it will take some weeks? I already asked at twitter and wanted are more detailed answer because it sounded more like a maybe.

Would like to port Sailfish OS later and I need a least some of the android userspace running.

One sidenote: why don't use open devices in the first place instead of having two codebases doing the same thing.

kholk:
Rule #1: there is no ETA!
Rule #2: if there is one, THERE IS NO ETA!!! 

... But I am going on vacation today and will be back in one month. Alin will most probably go on with this even if I'm out.
But.
I'd say that you won't see it reliably booting for at least another 3-4 weeks. Maybe less. Maybe more. :)) 

For now, the only thing that you can do is to patiently wait a lil :)))

PR #2037, opened 2019-07-24
[2.3.2.r1.4] kbuild: refactor and cleanup

lzlrd:
This may also speed up the build process.

### DIFF
--- a/Documentation/ia64/xen.txt
+++ b/Documentation/ia64/xen.txt
@@ -41,7 +41,7 @@ Getting and Building Xen and Dom0
 
  5. make initrd for Dom0/DomU
     # make -C linux-2.6.18-xen.hg ARCH=ia64 modules_install \
-      O=$(/bin/pwd)/build-linux-2.6.18-xen_ia64
+      O=$(pwd)/build-linux-2.6.18-xen_ia64
     # mkinitrd -f /boot/efi/efi/redhat/initrd-2.6.18.8-xen.img \
       2.6.18.8-xen --builtin mptspi --builtin mptbase \
       --builtin mptscsih --builtin uhci-hcd --builtin ohci-hcd \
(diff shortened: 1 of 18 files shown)


kholk:
Needs to be build tested a lot.
Do a mrproper, then add a file to a random makefile, then compile dirty. Is the new file picked?

Also, do a clean, then remove that file from the random makefile. Is the new file picked?!

Then, without mrproper or clean, do add and remove a file from a random makefile and dirty buils. Are the changes picked?

lzlrd:
@kholk I moved the worrysome commits to https://github.com/sonyxperiadev/kernel/pull/2038.

lzlrd:
> Needs to be build tested a lot.
> Do a mrproper, then add a file to a random makefile, then compile dirty. Is the new file picked?
> 
> Also, do a clean, then remove that file from the random makefile. Is the new file picked?!
> 
> Then, without mrproper or clean, do add and remove a file from a random makefile and dirty buils. Are the changes picked?

I tested what you said on the SLMK driver, and yes the expected occurs (removed a core file and expected it to break by not picking it). I and kdrag0n both used the commits for a while now, they seem to work fine as expected.

Only difference is instead of clean I killed the out/ dir.

kholk:
No, it has to be tested like I said and out of the Android build system!

kholk:
P.S.: Same testing goes for #2038!!

lzlrd:
@kholk your tests completed with the following result:
```
Case: Adding a header to the USB HID driver and building it - AKA <commit hash>.

Pre-actions: export ARCH=arm64 KBUILD_OUT=out CROSS_COMPILE=fakepath CROSS_COMPILE_ARM32=fakepath

The following tests were run one after another.

Test 1: make mrproper; make -j10
Result 1: New HID files picked and built.

Test 2: make clean; git revert <commit hash>; make -j10
Result 2: Files weren't picked.

Test 3: [REPEAT Test 1 then] git revert <commit hash>; make -j10
Result 3: Files weren't picked.
```

PS - I wasn't using the Android build system before or now :P.

lzlrd:
I would like if others could test too and verify.

PR #2038, opened 2019-07-24
[2.3.2.r1.4] kbuild: cache expensive vars

lzlrd:
This will speed up the build process, but may be dangerous and requires testing.


lzlrd:
This needs testing if anyone doesn't mind it.

PR #2039, opened 2019-07-27
[2.3.2.r1.4] Performance improvements

kholk:
This patchset brings some performance improvements and the addition of the LZO-RLE
algorithm to the kernel, also usable in zram (yup, tested, works but LZ4 is still ok for us).

The main performance improvement is for SWAP space: the locking has changed and
the swap cache is now split in 64MB trunks.
This gives us a **reduction of the median page fault latency of 375%**,  from 15uS to 4uS,
and an **improvement of 192% on the swap throughput** (this includes "virtual" swap
devices, like zRAM!). The real world user experience improvement of this on a mobile
device is seen after a day or two of usage, where it usually starts losing just a little
performance due to the large amount of apps kept open in background: now I cannot
notice any more performance loss and the user experience is **now basically the same as
if the phone was in its first 2 hours of boot life.**

Other performance improvements include, in short:
- UDP v4/v6: 10% more performance on single RX queue
- Userspace applications will be faster when checking running time of threads
- 2-5% improvements on heavy multipliers (yeah, not a lot, but was totally free...)
- Improvements on rare conditions during sparsetruncate of about 0.3% to a
  way more rare around 20% improvement (that's never gonna happen, but there
  is no performance drop anywhere).

Tested on SoMC Tama Akatsuki RoW


lzlrd:
Tested for two days on maple, performance is going great!

MartinX3:
what's still needed here to get merged? The other platforms? :)

PR #2042, opened 2019-08-07
[2.3.2.r1.4] Disable CONFIG_SCHED_DEBUG

mlehtima:
Not needed option which causes quite many debug messages in kernel log.


MarijnS95:
This should've been done in https://github.com/sonyxperiadev/kernel-defconfig/tree/aosp/LE.UM.2.3.2.r1.4 instead.

artem:
@jerpelea please, revert this. Android depends on SCHED_DEBUG and has it as part of VTS: https://android.googlesource.com/kernel/configs/+/0c277ea5a76c1a538a5147850fcc2b366108c34b%5E%21/

mlehtima:
Just curious, if that always needed why was it disabled in 4.4 kernel? I asked @jerpelea before doing the PR if that is needed or whether I should disable it only in Sailfish OS fork of the kernel.

tytydraco:
Android's init scripts setup the latency and granularity sysctl entries. These should either be set in the kernel manually, or this commit should be reverted.

jerpelea:
unfortunately I forgot that android needs them

PR #2044, opened 2019-08-07
[6.4.r1] (QualPwn) icnss: Add check on msa region

pablomh:
When icnss receive server arrive it send wlfw_msa_mem_info_send_sync_msg
QMI request to firmware and in response expect range of addresses and size
to be mapped. Add condition to check whether addresses in response falls
under valid range otherwise it asserts.

Change-Id: I9a8542cb6c3b3cefe112d1f08a76dd2eadf68d2f
Signed-off-by: Naman Padhiar <npadhiar@codeaurora.org>


pablomh:
For background:

[QualPwn - Exploiting Qualcomm WLAN and Modem Over The Air](https://blade.tencent.com/en/advisories/qualpwn/)

TLDR: "The full exploit chain allows attackers to compromise the Android Kernel over-the-air in some circumstance"

PR #2050, opened 2019-08-18
[LA.UM.6.4.R1] fpc1145: fix build on none-NILE targets

erikcas:
fix unused variable 'hw_type'

### DIFF
--- a/drivers/input/misc/fpc1145_platform.c
+++ b/drivers/input/misc/fpc1145_platform.c
@@ -559,7 +559,9 @@ static int fpc1145_probe(struct platform_device *pdev)
 	int rc = 0;
 	size_t i;
 	int irqf;
+#ifdef CONFIG_ARCH_SONY_NILE
 	int hw_type;
+#endif
 	struct device_node *np = dev->of_node;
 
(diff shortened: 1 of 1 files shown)


erikcas:
Oh, I see @kholk already did this on the LE.UM.2.3.2.r1.4 branch so that prevents a question if it needs to be done on other branches as well.
Forcepushed corrected author and commit msg

MarijnS95:
This was initially fixed in a different commit together with some other cleanups, but Angelo made this change separately so the cleanup has been rebased. Feel free to pull that one in as well for consistency: d66a700fadd34372a46c794a00f7eadf6aedce99

And the same for the Egistec driver: 6a08052ea80a71fb9908f1cfee3c4c4274e3a07b

erikcas:
@MarijnS95 I will push those on first bad weather occasion during vakantie in eigen land 🤣

MarijnS95:
@erikcas No worries I can push those as well (while checking if there aren't any other forgotten commits in there :see_no_evil:).

Hier schijnt de zon ook weer :sun_with_face:

PR #2057, opened 2019-09-13
iommu: qcom: iommu-v1: Fix typo in OR statement

oshmoun:


MarijnS95:
Functionality wise, this means the same. `!=` binds over `|` so there are no precedence issues, and there is no real reason to lazy evaluate. Fun fact: gcc4.9 for aarch64 and a random gcc on some Ubuntu system optimize this to a lazy evaluation on `-O3` :stuck_out_tongue:

PR #2059, opened 2019-09-14
[2.3.2.r1] Revert "usb: gadget: f_mtp: Increase default TX buffer size"

bartcubbins:
Should fix this issue: https://github.com/sonyxperiadev/bug_tracker/issues/432#issuecomment-517964636

### DIFF
--- a/drivers/usb/gadget/function/f_mtp.c
+++ b/drivers/usb/gadget/function/f_mtp.c
@@ -43,7 +43,6 @@
 #include "../configfs.h"
 
 #define MTP_RX_BUFFER_INIT_SIZE    1048576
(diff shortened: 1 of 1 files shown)


kholk:
I would agree with you if future devices needed this commit, and I definitely agree that fixing that in the kernel is generally a better and cleaner way, but it's legacy devices the ones needing it.

Let it get fixed the userspace way... Or this one will backfire on us on future devices because, if we merge this, then we will have to apply a fix for Tama and Kumano (the latter only on 4.14), plus anytime a new platform gets out.

You understand the point, right?

jerpelea:
please fix it in userspace for legacy devices

bartcubbins:
Let's fix it in userspace 😉

PR #2062, opened 2019-09-17
[7.1.r1] RPMSG hacks and other fixes for a successful boot on new bootloaders

kholk:
New bootloaders are sometimes shutting down the RPM related communication
channels and the kernel boots only when they don't get closed.

Workaround this issue by hacking the rpm-smd downstream frankenstein to always
allow to create RPM requests at the end of its probe, removing the requirement to
fail the subdevices probe before actually succeeding.

To workaround a new bug that gets up after this behavior modification, protect the
probe of the devices on the rpm bus with a spinlock.

I'm sorry for this, but it's the only way to get this done fast.

### DIFF
--- a/arch/arm64/boot/dts/qcom/msm8996.dtsi
+++ b/arch/arm64/boot/dts/qcom/msm8996.dtsi
@@ -458,6 +458,12 @@
 		rpm-channel-name = "rpm_requests";
 		interrupts = <GIC_SPI 168 IRQ_TYPE_EDGE_RISING>;
 		rpm-channel-type = <15>; /* SMD_APPS_RPM */
+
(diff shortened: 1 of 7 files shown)


oshmoun:
it would be ideal if only a subset of this PR is sufficient to resolve the issue.
the following commit out of this PR should be tested alone on a nile device running the latest bootloader:
https://github.com/sonyxperiadev/kernel/pull/2062/commits/b7306b7b9ce881776b1d28fb799a5b4d5ca3498c

jerpelea:
it will boot happy :)

kholk:
Then just cherry-pick the DT commits, it's fine :)))

kholk:
The PR has been updated with fixes.

Can be merged now.

PR #2063, opened 2019-09-18
cpuquiet: fix cross referencing, potentially dead loop bug

cocafe:
When _same_ notifier block is registered to different notifier chains,
this actually makes these notifier chains join together, which is
absolutely unwanted. for example:

    chain_add(min, foo);
    chain_add(max, foo);

        chain: min +---> ... +---> foo +---> null
                                    ^
                                    |
        chain: max +---> ... +------+

Then if there are one or more this kind of register pattern which also
register to same chains, this will cause dead loop while walking through
either chains:

    chain_add(min, new);

        chain: min +---> ... +---> foo +---> new +---> null
                                    ^
                                    |
        chain: max +---> ... +------+

    chain_add(max, new);
    notifier_chain_register() will result following:

        chain: min +---> ... +---> foo +---> new +---+
                                    ^         ^      |
                                    |         |      |
        chain: max +---> ... +------+         +------+

Signed-off-by: Ji Huang <cocafehj@gmail.com>

### DIFF
--- a/drivers/cpuquiet/cpuquiet.c
+++ b/drivers/cpuquiet/cpuquiet.c
@@ -266,7 +266,11 @@ static int minmax_cpus_notify(struct notifier_block *nb, unsigned long n,
 	return NOTIFY_OK;
 }
 
-static struct notifier_block minmax_cpus_notifier = {
+static struct notifier_block min_cpus_notifier = {
+	.notifier_call = minmax_cpus_notify,
+};
+
+static struct notifier_block max_cpus_notifier = {
 	.notifier_call = minmax_cpus_notify,
 };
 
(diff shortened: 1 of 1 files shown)


cocafe:
I'm trying to port cpu quiet framework from sony opensource kernel to my custom kernel, and these register patterns actually exist in code which cause dead loop and system hang:

 - https://github.com/cocafe/star2lte_kernel_cocore-s/commit/e0fe0462ec2ee87f60c09c07459bb6271f156a7e
 - https://github.com/cocafe/star2lte_kernel_cocore-s/commit/e34139d9b29e58f4290280af96eeb8c6ff43a9d2

Compilation tested and passed on `drivers/cpuquiet/cpuquiet.c` with current branch ` aosp/LA.UM.7.1.r1 `.

kholk (approved):

kholk:
Thanks!

PR #2065, opened 2019-09-21
[2.3.2.r1.4] Linux 4.9.194

nathanchance:


stefanhh0:
I've tested this PR with lilac/yoshino in combination with android 10 and found that it builds and runs just fine.

PR #2053, opened 2019-09-07
[DONOTMERGE] [2.3.2.r1.4] Support TCM incell touch displays on Apollo and Akari

MarijnS95:
Fixes https://github.com/sonyxperiadev/bug_tracker/issues/351
Fixes https://github.com/sonyxperiadev/bug_tracker/issues/262

Please do not merge this. The text below is merely an overview of what _has been addressed_. The commits might very well do this _the wrong way_. Let's discuss this together.

Unfortunately these devices have a hardware lottery with either Clearpad or TCM hardware.
The panel driver is rather messy and reports a powered panel+touch when it's not fully powered and/or still missing the mandatory reset. Clearpad _**hacks**_ around this by restarting a worker multiple times that probes i2c until it succeeds, which happens finally when the panel prepares.

This patch addresses that issue on Clearpad by forcing the incell touch on and the mandatory reset before probing. The 5-or-so failed attempts including stacktraces are now gone (and hopefully users report less instabilities).
Since TCM doesn't have such a thing, it calls for the incell to power up directly from `probe()`.

Finally, a little mess has been added that allows the TCM and Clearpad drivers to co-exist in the same kernel at runtime. Either one will be switched off when the other touch type is detected.
Stock does it the same way, from a userspace shell-script with kernel modules.

<details>
<summary>I have not included `CONFIG_TOUCHSCREEN_SYNAPTICS_TCM_FWUPG` which enabled firmware upgrading, and it seems to work fine without. But in order to make sure this works on every device, we _should_ add the images listed below to the repositories, and enable `FWUPG` in the defconf.</summary>

```
project device/sony/akari/
diff --git a/device.mk b/device.mk
diff --git a/device.mk b/device.mk
index a99479a..77cdac7 100644
--- a/device.mk
+++ b/device.mk
@@ -82,6 +82,13 @@ PRODUCT_COPY_FILES += \
 # BT/Wifi patch file
 PRODUCT_COPY_FILES += \
     $(DEVICE_PATH)/vendor/firmware/wlan/qca_cld/bdwlan.bin:$(TARGET_COPY_OUT_VENDOR)/firmware/wlan/qca_cld/bdwlan.bin

+# Synaptics TCM touch firmware files
+PRODUCT_COPY_FILES += \
+    $(DEVICE_PATH)/vendor/firmware/touch_module_id_0x20.img:$(TARGET_COPY_OUT_VENDOR)/firmware/touch_module_id_0x20.img \
[...]

PR #2076, opened 2019-10-09
[2.3.2.r1.4] [PLEASE REVIEW] TCM/Clearpad: Fix probe, replace broken ifdefs with DT property

MarijnS95:
For https://github.com/sonyxperiadev/kernel/pull/2053#issuecomment-539958913

input: ts: clearpad/tcm: Return ENODEV when a different device is found. 

> The probe must be failed since this device doesn't exist. A successful
> probe leaves an uninitialized device in the tree, where pm functions are
> called in certain circumstances.
> This crashes the device as drvdata that is used pretty much everywhere
> remains unset.

panel/touchscreen: Replace broken macro with DT property.

> This macro throws warnings about undefined behaviour when expanding
> defined() preprocessor functions. Instead of coming up with a different
> solution, store the behaviour in a DT property since it is panel-bound
> anyway. This saves the unnecessary panel name comparison.
> 
> This is safe as long as all devices using TCM define somc,has-tcm-touch.
> If we ever get a device with purely TCM and no CLEARPAD, just stating so
> in the defconfig doesn't work. Should this patch include ifdefs to fall
> back to the right type?

@kholk Let me know if this is close to what you wanted to see following https://github.com/sonyxperiadev/kernel/pull/2053#issuecomment-539978713.

### DIFF
--- a/drivers/gpu/drm/msm/dsi-staging/somc_panel/dsi_panel_driver.c
+++ b/drivers/gpu/drm/msm/dsi-staging/somc_panel/dsi_panel_driver.c
@@ -928,6 +928,11 @@ int dsi_panel_driver_parse_dt(struct dsi_panel *panel,
 
 	spec_pdata = panel->spec_pdata;
 
+	if (of_property_read_u32(np, "somc,incell-touch-type",
+			&panel->touch_type))
+		panel->touch_type = INCELL_TOUCH_TYPE_DEFAULT;
+	pr_debug("%s: In-Cell touch IC is %d\n", __func__, panel->touch_type);
+
 	rc = somc_panel_parse_dt_colormgr_config(panel, np);
 
 	if (of_find_property(np, "somc,pw-off-rst-b-seq", NULL)) {
--- a/drivers/gpu/drm/msm/dsi-staging/somc_panel/incell.c
+++ b/drivers/gpu/drm/msm/dsi-staging/somc_panel/incell.c
@@ -289,24 +289,13 @@ void incell_driver_init(struct msm_drm_private *priv)
 		incell->state = INCELL_S000;
 }
 
-#if INCELL_TAMA_MULTIPLE_TOUCH_DRIVERS
-incell_touch_type incell_get_touch_type() {
+bool incell_touch_is_compatible(incell_touch_type type) {
 	struct dsi_display *display = dsi_display_get_main_display();
 	struct dsi_panel *panel = display->panel;
-	incell_touch_type type = INCELL_TOUCH_TYPE_CLEARPAD;
 
-#if defined(CONFIG_MACH_SONY_AKARI)
-	if (!strcmp(panel->name, "7"))
-#elif defined(CONFIG_MACH_SONY_APOLLO)
-	if (!strcmp(panel->name, "8"))
-#else
-#error "INCELL_TAMA_MULTIPLE_TOUCH_DRIVERS without AKARI or APOLLO!"
-#endif
-		type = INCELL_TOUCH_TYPE_TCM;
+	if (panel->touch_type == INCELL_TOUCH_TYPE_DEFAULT)
+		/* Default/unset. Any driver is allowed to probe: */
+		return true;
 
-	pr_info("%s: Panel %s is %s\n", __func__, panel->name,
-		type == INCELL_TOUCH_TYPE_TCM ? "TCM" : "CLEARPAD");
-
-	return type;
+	return type == panel->touch_type;
 }
-#endif
--- a/arch/arm64/boot/dts/qcom/dsi-panel-somc-synaptics-auo-1080p2160-cmd.dtsi
+++ b/arch/arm64/boot/dts/qcom/dsi-panel-somc-synaptics-auo-1080p2160-cmd.dtsi
@@ -99,6 +99,8 @@
 		somc,fps-mode-enable;
 		somc,fps-mode-panel-mode = "dynamic_mode";
 
+		somc,incell-touch-type = <2>;
+
 		qcom,mdss-dsi-display-timings {
 			timing@0 {
 				qcom,mdss-dsi-panel-width = <1080>;
--- a/arch/arm64/boot/dts/qcom/dsi-panel-somc-synaptics-jdi-1080p2160-cmd.dtsi
+++ b/arch/arm64/boot/dts/qcom/dsi-panel-somc-synaptics-jdi-1080p2160-cmd.dtsi
@@ -99,6 +99,8 @@
 		somc,fps-mode-enable;
 		somc,fps-mode-panel-mode = "dynamic_mode";
 
+		somc,incell-touch-type = <1>;
+
 		qcom,mdss-dsi-display-timings {
 			timing@0 {
 				qcom,mdss-dsi-panel-width = <1080>;
@@ -239,6 +241,8 @@
 
 		somc,fps-mode-enable;
 		somc,fps-mode-panel-mode = "dynamic_mode";
+
+		somc,incell-touch-type = <1>;
 
 		qcom,mdss-dsi-display-timings {
 			timing@0 {
--- a/arch/arm64/boot/dts/qcom/dsi-panel-somc-synaptics-lgd-1080p2160-cmd.dtsi
+++ b/arch/arm64/boot/dts/qcom/dsi-panel-somc-synaptics-lgd-1080p2160-cmd.dtsi
@@ -99,6 +99,8 @@
 		somc,fps-mode-enable;
 		somc,fps-mode-panel-mode = "dynamic_mode";
 
+		somc,incell-touch-type = <2>;
+
 		qcom,mdss-dsi-display-timings {
 			timing@0 {
 				qcom,mdss-dsi-panel-width = <1080>;
--- a/drivers/gpu/drm/msm/dsi-staging/dsi_panel.h
+++ b/drivers/gpu/drm/msm/dsi-staging/dsi_panel.h
@@ -205,6 +205,7 @@ struct dsi_panel {
 
 #ifdef CONFIG_DRM_SDE_SPECIFIC_PANEL
 	struct panel_specific_pdata *spec_pdata;
+	int touch_type;
 #endif /* CONFIG_DRM_SDE_SPECIFIC_PANEL */
 };
 
--- a/drivers/gpu/drm/msm/dsi-staging/somc_panel/incell.h
+++ b/drivers/gpu/drm/msm/dsi-staging/somc_panel/incell.h
@@ -51,15 +51,11 @@ typedef enum {
 	INCELL_DISPLAY_POWER_LOCK,
 } incell_pw_lock;
 
-#define INCELL_TAMA_MULTIPLE_TOUCH_DRIVERS \
-	(defined(CONFIG_MACH_SONY_AKARI) || defined(CONFIG_MACH_SONY_APOLLO))
-
-#if INCELL_TAMA_MULTIPLE_TOUCH_DRIVERS
 typedef enum {
-	INCELL_TOUCH_TYPE_CLEARPAD,
-	INCELL_TOUCH_TYPE_TCM,
+	INCELL_TOUCH_TYPE_DEFAULT = 0,
+	INCELL_TOUCH_TYPE_CLEARPAD = 1,
+	INCELL_TOUCH_TYPE_TCM = 2,
 } incell_touch_type;
-#endif
 
 /* Compatibility with older incell */
 #define INCELL_DISPLAY_HW_RESET		INCELL_TOUCH_RESET
@@ -146,15 +142,13 @@ extern int incell_get_display_pre_sod(void);
 static inline int incell_get_display_pre_sod(void) {return 0; }
 #endif /* CONFIG_DRM_SDE_SPECIFIC_PANEL */
 
-#if INCELL_TAMA_MULTIPLE_TOUCH_DRIVERS
-
 #ifdef CONFIG_DRM_SDE_SPECIFIC_PANEL
-extern incell_touch_type incell_get_touch_type(void);
+bool incell_touch_is_compatible(incell_touch_type type);
 #else
-static inline incell_touch_type incell_get_touch_type(void) {return 0; }
+static inline bool incell_touch_is_compatible(incell_touch_type) {
+	return true;
+}
 #endif /* CONFIG_DRM_SDE_SPECIFIC_PANEL */
-
-#endif /* INCELL_TAMA_MULTIPLE_TOUCH_DRIVERS */
 
 #endif /* __INCELL_H__ */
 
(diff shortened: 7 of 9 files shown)


kholk:
I agree yours is better,  though, you have underestimated this.
Please, use a different property, like "somc,incell-touch-type" and read it as an integer, so that we get something like
`somc,incell-touch-type = <N> `where N is:
0 - Default (Clearpad)
1 - TCM
2 - Samsung Electronics (in case needed in the near future for Kumano)

P.S.: better safe than sorry... Obviously, if the property cannot be read (not specified in DT) the driver has to assign 0 to the touch_type variable

MarijnS95:
> though, you have underestimated this.

Well, you see, this is only a step-up to the final implementation. Hence the request for a review, as I already mentioned this will not scale nicely to TCM-only platforms, which suddenly need an obnoxious DT property.

Anyway, I was thinking slightly differently:

0. Default. This value is ignored and drivers continue the probe anyway, as if they have purely been enabled by defconfig/DT;
1. Explicit Clearpad. Anything else returns `-ENODEV`;
2. Explicit TCM. Anything else returns `-ENODEV`;
3. ... andsoforth for other drivers.

However, on the other hand this requires the DT property specified on the Clearpad Apollo/Akari panel. In my opinion, making that explicit there is much better than making it explicit on any TCM panel.

Also, I want to move the enum to `dt-bindings`. Seeing as there are no other bindings yet... Good idea? Right now I just put `int touch_type;` in `dsi_panel.h` instead of polluting that with `incell.h`.
Btw, I have implemented the (n)ok in the function instead of returning just `panel->touch_type`, to not duplicate the if structure (just like `of_machine_is_compatible`.

kholk:
No it's not good like this. You'd have to change all the panels that we have to declare clearpad when that's exactly the current default behavior.
Then you'd have to change the default to a hypothetical behavior that we may (most probably not) have in a future that can be far and pretend that the hypothetical default does the right things. Or leave "NOT IMPLEMENTED" comments all over. Oh God. 

Nnnnnoupe.

Just go for logical clearpad default (zero). If anything changes in the future, then it'll be a problem for the future me or the future you......

MarijnS95:
@kholk I think you misunderstand. A default of `0` makes this piece of code a no-op. It doesn't prevent any driver from probing and instead relies on defconfig+DT, **just like it was before the patches**.

Now, **only** if we have a device with multiple digitizers (which implies multiple drivers compiled in through defconfig as well as the relevant DT bindings) do we need to specify `somc,incell-touch-type` on all relevant panels, or in an override node in a device dtsi.

kholk:
Okay, I have misread and I'm sorry for that.

Still, I don't get why we shouldn't just "include" the clearpad behavior as zero (default).. For example, Griffin's SEC TS (if my memory is still working right) uses the default code flow, which is the same as the devices using the clearpad driver.... And there is no clearpad device using a different flow from the others.

Basically, my idea is basically exactly as your previous bool property to specify TCM, but just expressed as a u32 number instead for future possible incell ts differences.

I still think that your 0-default 1-clearpad idea is duplicating a behaviour hence overcomplicating the human understanding of the code......

Or perhaps my brain just can't get your concept. That's also possible, of course.

MarijnS95:
> Still, I don't get why we shouldn't just "include" the clearpad behavior as zero (default)

Because that will break on devices with just TCM. We don't have any at the moment, but I can definitely imagine the bringup annoyances when someone finds it hasn't been working because somehow TCM is explicit and Clearpad isn't.

> Griffin's SEC TS (if my memory is still working right) uses the default code flow, which is the same as the devices using the clearpad driver.... And there is no clearpad device using a different flow from the others.

I take it Griffin has just one panel and digitizer, right? There my "default value does nothing" approach strikes again, requiring no additional configuration to get the digitizer to probe.

> Basically, my idea is basically exactly as your previous bool property to specify TCM, but just expressed as a u32 number instead for future possible incell ts differences.

And that is exactly what I pushed before responding :grin:. The only difference being that your version **enforces** Clearpad by default (and requires alteration if we ever get devices with purely TCM - and other touch drivers if this edgecase is added there as well), while my version (the one that I've pushed) behaves as if this piece of code wasn't even there, by default.
That way, one can safely use TCM or Clearpad without ever worrying about this explicit value. Only if we get another device with panel+digitizer roulette do we need to set the property and force a specific digitizer driver on a specific panel.

Perhaps this is just a design problem in the kernel/DT itself. IMO there should have a been a more clear and explicit DT relation between panels and digitizers. Though that sure makes things complicated and may not work as well in certain scenarios. This integer is like a retrofit hacky version of that.

drivers/gpu/drm/msm/dsi-staging/somc_panel/incell.c
```diff
@@ -289,24 +289,13 @@ void incell_driver_init(struct msm_drm_private *priv)
...
-#elif defined(CONFIG_MACH_SONY_APOLLO)
-	if (!strcmp(panel->name, "8"))
-#else
-#error "INCELL_TAMA_MULTIPLE_TOUCH_DRIVERS without AKARI or APOLLO!"
-#endif
-		type = INCELL_TOUCH_TYPE_TCM;
-
-	pr_info("%s: Panel %s is %s\n", __func__, panel->name,
-		type == INCELL_TOUCH_TYPE_TCM ? "TCM" : "CLEARPAD");
+	if (type == INCELL_TOUCH_TYPE_DEFAULT)
```
MarijnS95: Funny how one can watch a random movie, and suddenly - hours after typing this - realize it is wrong. This should be `panel->touch_type == INCELL_TOUCH_TYPE_DEFAULT`, which occurs when the DT property is not set.

In fact, `type` (which I should rename to be more descriptive) can never be `DEFAULT` because it is the type of the touch driver calling this function.

kholk:
Again, this way you'd have to add this property to also Tone, Nile, Yoshino panels, because they are still using the incell clearpad.

Besides that, the only incompatibility that you get is between TCM and clearpad... On any other touch chip, when you read the I2C on the clearpad driver, it won't recognize the touch model and bail out instead of probing fine.

I really don't get why you want to forcefully probe a driver based on a DT property instead of using its own i2c detection mechanism even on devices that use other touch chips, because doing it this way, that's what you're doing.

I mean, checking for INCELL_TYPE_BLAH_CLEARPAD in the clearpad driver means that if you set DEFAULT, that driver will just discard whatever detection it properly did on I2C and say goodbye.
This is a bad design and it's counter intuitive: you are being extremely paranoid about this thing and you should remember that we still already have two detection mechanisms (DT compatible and i2c bus read). You are adding a third one that is enforcing not to probe clearpad on clearpad devices if the clearpad property is not found in the PANEL DT........

EDIT
If you really disagree with me on this, then okay it's fine, but the PR is still wrong as it will break other devices because you didn't declare clearpad on all of them.
I would be fine merging this so that, again, if you don't understand now what I mean, you'll see in the future why I'm saying this and you'll learn... :P

MarijnS95:
> Again, this way you'd have to add this property to also Tone, Nile, Yoshino panels, because they are still using the incell clearpad.

I must have explained it incorrectly. The default option means... just probe, always and whatever driver.
https://github.com/sonyxperiadev/kernel/blob/a69691addd453b833452640751c59128a67544e7/drivers/gpu/drm/msm/dsi-staging/somc_panel/incell.c#L296-L298

There, if the `touch_type` is default (unset) the actual digitizer type is completely disregarded, making this work as if the new function and `-ENODEV` is not even there.
On other words... You could say that, just like in your example, default == clearpad. But default also == TCM. 

> Besides that, the only incompatibility that you get is between TCM and clearpad... On any other touch chip, when you read the I2C on the clearpad driver, it won't recognize the touch model and bail out instead of probing fine.

That is... Something I totally forgot about. During initial bringup - because I do not have a TCM device close -  I have intentionally disabled Clearpad to prevent any conflicts.
Initially Incell didn't report poweron status correctly, making Clearpad fail and retry a delayed probe many times before finally giving up, since the I2C device was unreachable.
That has been "addressed" in the previous PR, so it is worth checking just as you say.

I will check with @MartinX3 to publish a test build without any of this code. If it works I totally agree that this edgecase should be gone forever - I just totally forgot about it and don't really trust these drivers anyway.

EDIT:
> you should remember that we still already have two detection mechanisms (DT compatible and i2c bus read)

DT compatible means nothing on a device that has panel/digitizer roulette. We have two DT nodes, so two digitizer drivers probe. If it is however as you say, where the driver figures out over i2c whether this device is the right one, and bail out accordingly **without messing up other state**, I totally agree and this PR should instead remove the useless piece of code altogether.

kholk:
No I don't know whether your explaination was incorrect or if it's me just being elsewhere with my brain... And reading this stuff on phone doesn't help either. 

In any case, I've just understood what you did and yup that's basically what I asked you to do, so the only complain on my side is about that thingy that you've also understood about the I2C detection...

But anyway, from what I remember, the clearpad driver wants to probe (and does recognize) in place of TCM. In this case I think that we can exclude the TCM touch IC's identifier in the I2C model recognition part of clearpad, unless we have a clearpad with the same model, which I think is unlikely.

MarijnS95:
> No I don't know whether your explaination was incorrect or if it's me just being elsewhere with my brain... And reading this stuff on phone doesn't help either.

I see, the diff is very messy after all. And I agree it is not really logical code, that doesn't help either.

>  In this case I think that we can exclude the TCM touch IC's identifier in the I2C model recognition part of clearpad, unless we have a clearpad with the same model, which I think is unlikely.

I cannot really comment on this since Clearpad and TCM use very different messaging semantics. It seems very unlikely that one such query message is understood and responded with magically the same version code, which are strings in Clearpad:
https://github.com/sonyxperiadev/kernel/blob/a69691addd453b833452640751c59128a67544e7/drivers/input/touchscreen/clearpad_incell_core.c#L425-L436
However I cannot find the code responsible for bailing if it doesn't find a matching chip. From what I can see right now after a very brief look, it'll most likely try quite a few times just like when the device is powered off.
To be more specific, `clearpad_initialize` that reads the `chip_id` is  called from `clearpad_process_irq` within `clearpad_handle_if_first_event`. This function ignores the result, so it'll most likely go on forever, perhaps erroring out a few times when sending other commands. Unless there is something that stops if no dt configuration is bound (which happens only on a known chip_id).

Again, there is just too much to read and the only way to be sure is to try.

I just want to remind you that stock also binds the drivers to the panel, by reading the name in userspace and `insmod`ing the right driver. And I really don't want to move towards praying TCM/Clearpad detect their own devices properly **without messing up the others' state by sending invalid messages**.

kholk:
Uhm.... Well... You can always unregister the driver at that stage, even after hypothetically succeeding the probe function! 
Of course I agree this wouldn't be clean, but I think it'd be better than the stock way, which is something that I REEEEEEAAALLLY don't like.

About sending wrong messages to the IC, well, I don't think that you should worry about that really. In the event of some very bad message that locks up the IC (which should be unlikely at that stage) you can just pull the RST pin for the touch ic and start from scratch without any problem.

MarijnS95:
Of course the stock way with modules and userspace andsoforth is far from optimal (after all, we want the kernel to "just work" regardless of what is (or is not) in userspace).

But my opinion is that - just like stock - since we know beforehand which driver to use we make sure only that particular driver gets to probe. Yes, theoretically we can figure out in the driver itself whether it's talking to a compatible device, mess with the reset pin and whatnot, but that seems like much more of a hassle, resulting in only more complicated code.
Sure, if it works "just fine" out of the box we can go ahead, but remember we do not have a large enough test surface (in fact, I'm relying on @MartinX3 to relay builds and logs+feedback) and I do not want to break functionality for existing Clearpad users.
Say for example when there is whatever timing/racecondition/noise, causing one of the drivers to say "hey this is not my incell touch" when it in fact is. Solve that with retrying/testing a couple times... you see where this is going.

kholk:
After all, this is a downstream kernel....

Yeah let's see if it works just fine OOTB, otherwise we merge this one and call it a day.

MarijnS95:
Aaand the results are in! While it works without this code on both Clearpad and TCM, I feel less than confident in using it.
TCM on  Clearpad device fails in the following way:
```
[    1.996877] i2c_geni 894000.i2c: i2c error :-107
[    1.996986] synaptics_tcm_i2c 2-003c: syna_tcm_i2c_read (line 379): Transfer attempt 1 failed
```
Most likely the Clearpad device didn't have any data to send, as the protocol is different.

Then, the Clearpad driver on TCM. This one worries me a bit more because it is purely based on probing order. We will have to check if it also works the other way around.
```
[    3.914453] genirq: Flags mismatch irq 381. 00002002 (clearpad) vs. 00002008 (synaptics_tcm)
[    3.914599] clearpad clearpad: (clearpad_probe:8852) failed to request threaded irq 381 (rc=-16)
[    3.914785] clearpad: probe of clearpad failed with error -16
```

In any case, I will defend my statement once more despite already having "sort of" an agreement. We _know_ perfectly fine what driver to use based on the panel that's detected. Now to throw that data away and instead rely on a bunch of **error-spitting** randomness seems nonsense to me.

drivers/gpu/drm/msm/dsi-staging/somc_panel/dsi_panel_driver.c
```diff
@@ -928,6 +928,11 @@ int dsi_panel_driver_parse_dt(struct dsi_panel *panel,
 
 	spec_pdata = panel->spec_pdata;
 
+	if (of_property_read_u32(np, "somc,incell-touch-type",
+			&panel->touch_type))
+		panel->touch_type = INCELL_TOUCH_TYPE_DEFAULT;
+	pr_debug("%s: Panel digitizer is %d\n", __func__, panel->touch_type);
```
kholk: %s: In-Cell touch IC is %d

kholk:
I disagree about the probe failure being bad but that needs some love to work flawlessly and we'd need time that noone currently has.

Fix the print that I said and then let's merge this.

MarijnS95:
> I disagree about the probe failure being bad but that needs some love to work flawlessly and we'd need time that noone currently has.

Sure, we might be able to make something that is both stable _and_ easy to identify in future bringup (of course, there's nothing worse than such a stupid piece of code breaking driver probe due to a missing regulator/gpio somewhere).
But as you note correctly, this is going to take time to validate _on all devices_, better spent elsewhere. I have a feeling there's always going to be a source of instability, ruling out a driver sometimes when it shouldn't (and instead needs a bit more time to intialize/power-up), but to be honest that's just a gut feeling and not based on critical facts.

> Fix the print that I said and then let's merge this.

I was thinking about stringifying the otherwise "meaningless" number. Good idea?
And, how about adding the enum/defines to a dt-bindings header? And documentation for the dt property?

MartinX3:
Dear @kholk is someone else needed here?
My tests on 14.10.2019 were successful on apollo with TCM and akari with clearpad.

kholk:
No, I was waiting for that kernel message fix, but it's fine without as well.

Go on and merge this

MarijnS95:
> No, I was waiting for that kernel message fix, but it's fine without as well.

@kholk Huh?

https://github.com/sonyxperiadev/kernel/compare/a69691addd453b833452640751c59128a67544e7..18650dfbf4dc8db0605dfd1617e0db78df378e3e

kholk:
Ah you've even fixed it. Oh well. :D

PR #2077, opened 2019-10-15
[LA.UM.7.1.r1] Revert "iommu: qcom: iommu-v1: Enable DMA allocations"

pablomh:
This reverts commit 313f4a2fc9eb4af5f540c9e58181c7301c1556bc.


pablomh:
According to @oshmoun:

"It introduces some nice cyclical dependencies by assigning `.iommu_dev` to `&ctx_drvdata->pdev->dev` which sets up `ctx_drvdata` in the first place."

kholk:
Merge it, this commit would need some more work and it's not worth it to save this driver from destruction.

One day (soon, but not too much) I will anyway remove this driver and replace it.

PR #2081, opened 2019-10-24
[LA.UM.7.1.r1] kbuild: Disable 'address-of-packed-member' warning not only for clang

Thaodan:
GCC also emits those warnings which breaks builds that enable modules
like the audio techpack from qualcomm.

Fix errors like this which are already ignored by clang:
```
kernel/sony/msm-4.14/kernel/techpack/audio/asoc/msm-dai-q6-v2.c:6413:10: error: taking address of packed member of 'str\
uct afe_param_id_group_device_tdm_cfg' may result in an unaligned pointer value [-Werror=address-of-packed-member]                                                                            
 6413 |   (u32 *)&tdm_group_cfg.group_id);     
```


nathanchance:
I’d recommend taking the upstream version just so that there isn’t a conflict when merging in stable: https://git.kernel.org/stable/c/8e69458509d2acebb1297753d0320ff2e8ce5f64

Thaodan:
Changed as suggested.

PR #2088, opened 2019-11-06
[7.1.r1] arm64: DT: HACK: sdm660-audio: Remove cdc-vdd-mic-bias supply

kholk:
This supply makes Linux to -ENOMEM fail when devres_alloc'ing
the bulk supplies in the sdm660 audio driver.

The audio works without this supply, though... this is a pure
hack and it's not expected to work without issues in the future.

-----------------------------------------------------

I don't know if it is a good idea to merge this commit. With it in,
audio will work on at least SoMC Nile Discovery.... but I cannot give
any guarantees about future issues, as stated in the commit
description...................


oshmoun:
please add a TODO or some marker above the commented line

PR #2092, opened 2019-11-09
[7.1.r1]arm64: DT: msm8996: Remove remnants of ea references

tomgus1:
These are also required to be removed to allow tone to compile

@kholk @jerpelea


kholk:
Remove only the EA subnodes, not the entire thing please

kholk:
Noupe, sorry for the noise. It's not needed to keep the main node.

Merge it as it is!

PR #2109, opened 2019-11-27
[7.1.r1] uapi: media: msm_vidc_utils: Compatibility for old Venus MISR struct

kholk:
Some targets need to use a different structure for the MISR
informations, since that goes back and forth to the Venus
firmware, which is old on these targets and obviously cannot
understand a different struct when sending/receiving packed
data.


MarijnS95:
Pasting https://github.com/sonyxperiadev/hardware-qcom-media/pull/1 To be able to cross-reference this in the future :grin:

PR #2111, opened 2019-11-28
[7.1.r1] Linux 5.4 ExFAT driver 

kholk:
This patchset integrates the ExFAT driver in our 4.14 kernel.

The ExFAT driver itself is not backported, but instead I've backported the
new i_version API from 5.4 to avoid touching the ExFAT driver itself.
FS drivers such as EXT4, FAT, NFS, EXT2, ecryptfs and Android sdcardfs
were forwardported to this new API instead.


artem:
Why couldn't we continue using SdFAT? Driver from mainline is based on an ancient source drop from Samsung and is admittedly more buggy and slower. I would even suggest to use https://github.com/arter97/exfat-linux because @arter97 has done a nice work to improve overall shape of the driver (removing dead code and backporting patches intended for currently staging driver)

kholk:
That was my initial intention. Then, people asked for it.

I've decided to use this and make them happy because, at the end of the day, this driver works, gets free maintenance from upstream and... Well... I tried it before calling it a day... It's not bad. Perhaps slower, for now, but things are going to change in one month, exactly, and then again after another month and another, and another, and...... :)))

arter97:
Maybe you should consider the fact that this thing is still in staging. There are real critical fixes made to sdFAT over exFAT 1.x, few of them include panic and corruption fixes.

Samsung will upstream sdFAT sooner than later, it's just unfortunate to see mainline merging exFAT 1.x without any prior notice to Samsung for any help.

arter97:
A totally new exFAT base will soon go upstream and the existing exFAT will be removed:
https://lkml.org/lkml/2019/11/13/116
(Already got fs maintainer's nod of approval: https://lkml.org/lkml/2019/11/13/738 )

I can personally guarantee you that this will bring you more maintenance headache.

arter97:
@kholk

kholk:
Whenever that implementation lands on an upstream release the driver will also be updated here.

Stuff being sent to the mailing list means nothing. It's just stuff floating around.

PR #2115, opened 2019-11-30
[7.1.r1] [ URGENT] Android.mk: TARGET_KERNEL_MODULES is no longer supported

kholk:
This is no longer supported in AOSP and the build will error out if
we define it.

The kernel modules are compiled and installed in a different way now.

### DIFF
--- a/Android.mk
+++ b/Android.mk
@@ -152,10 +152,6 @@ else
 endif
 endif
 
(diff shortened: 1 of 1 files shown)


kholk:
Then it fails elsewhere... so don't bother.... let's wait for Felix...

PR #2117, opened 2019-12-03
[LA.UM.7.1.r1] gpu: drm: panel_color_manager: SDE_CP_CRTC_DSPP_PCC is atomically set.

MarijnS95:
Starting with kernel 4.14, the color transform property that we were
overriding is now set atomically. Override atomic_set_property to be
able to merge requested color transformations with the color
calibrations coming from the panel driver again.

... And finally I have night light back... For what's still left of those burnt-away eyes :grin:


kholk:
I kept postponing this fix to look at other stuff....

Thanks man :D

PR #2118, opened 2019-12-03
[7.1.r1] msm: camera: actuator: Fix range for BU64747 actuator

kholk:
The actuator has a sensible range of 200-800, where 800 is
for near subjects and 200 is "infinity".

Address the userspace framework issues with the actuator range
by normalizing the incoming position request data in the
supported range.

NOTE: THIS IS A WORKAROUND - MOSTLY THE SAME THAT I
PUSHED FOR KERNEL 4.9 BASED BSPs.

THIS IS NOT **THE** SOLUTION (but myopia stops, so feel free
to use it while the camera userspace gets properly fixed/calibrated).

### DIFF
--- a/drivers/media/platform/msm/camera/cam_sensor_module/cam_actuator/cam_actuator_core.c
+++ b/drivers/media/platform/msm/camera/cam_sensor_module/cam_actuator/cam_actuator_core.c
@@ -166,6 +166,72 @@ static int32_t cam_actuator_power_down(struct cam_actuator_ctrl_t *a_ctrl)
 	return rc;
 }
 
+static int32_t bu64747_set_dac(struct camera_io_master *io_master_info,
+				struct i2c_settings_list *i2c_list)
+{
+	struct cam_sensor_i2c_reg_array *reg_setting = NULL;
+	uint16_t bu64747_dac = 0;
+	uint16_t new_max, new_min, new_range;
+	uint16_t old_max, old_min, old_range, req_val;
+	bool bad_range = false;
+	int32_t rc = 0;
+
+	reg_setting = i2c_list->i2c_settings.reg_setting;
+	req_val = reg_setting[0].reg_data;
+
+	/* Framework-passed (old) and real (new) min-max values */
+	old_max = 737;
+	old_min = 391;
+	new_max = 800;
+	new_min = 200;
+
+	/* Are we parking the lens? */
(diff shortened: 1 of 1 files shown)


kholk:
@sraase I don't know whether that was already known or not, but this appears
to be the same issue that we had on the previous k4.9 based CamX.

Tagging you to make you aware of the issue again.

sraase:
The actual drivers have not changed between CamX versions, so I did not expect any change in behaviour. I hope to be able to investigate this in more detail in the near future.

scripts/kconfig/lxdialog/check-lxdialog.sh
```diff
@@ -2,11 +2,25 @@
...
+# stub out pkg-config to avoid build system errors. Uff...
+if [ -n "${PRODUCT_SOONG_NAMESPACES}" ] && [ -n "${TARGET_PRODUCT}" ]; then
+pkgconfig()
+{
+	:
+}
+else
+pkgconfig()
+{
+	pkg-config $@
```
Thaodan: If you add command in front of running pkg-config you can just override pkg-config without renaming anything down the line. Or you just define this empty function when ever the above successedes with the name of pkg-config.

scripts/kconfig/lxdialog/check-lxdialog.sh
```diff
@@ -2,11 +2,25 @@
 # SPDX-License-Identifier: GPL-2.0
 # Check ncurses compatibility
 
+# Android Q doesn't allow us to call pkg-config for some reason, so
+# stub out pkg-config to avoid build system errors. Uff...
+if [ -n "${PRODUCT_SOONG_NAMESPACES}" ] && [ -n "${TARGET_PRODUCT}" ]; then
+pkgconfig()
```
Thaodan: Rename this to pkg-config and you don't need the other function and you don't need to rename the function calls in ldconfig.

scripts/kconfig/lxdialog/check-lxdialog.sh
Thaodan: If you use $@ you should quote is or this may break when it gets expanded.

Thaodan (changes requested):

jerpelea:
for now we add the patch while a solution is investigated

PR #2120, opened 2019-12-04
[7.1.r1] Fix clang build

kholk:
Now the kernel can be built with clang.
Tested on Clang 9.


nathanchance:
Hmmm disabling the stack protector shouldn’t be necessary... if you are getting that error, it usually means that there is a GCC only flag that has not been wrapped with `cc-option` so when the check happens for `-fstack-protector`, it unconditionally fails because there is a flag that has been added that it does not recognize. I have a way to debug this that I can share/test later.

Thaodan:
Please look at my comments on the other pr were you had the pkg-config changes.
You can just define pkg-config as function if you detect the android-build-system and don't need to change anything after that.

kholk:
Thanks for the feedback. I will look at your proposals.

By the way, making it clear, no I didn't get totally crazy. It's just about having done a code marathon today :D

Thaodan:
Great to hear :)

Just a little bit crazy?  Sometimes being crazy is 
a good thing.
But as long as you don't up ending like the 
original marathon runner everything is ok :D

jerpelea:
please remove the stack protector commit !

kholk:
@Thaodan Now that my brain "works" again: so... your proposal is the first thing I tried before doing it the way I did but - for some reason I didn't understand (and didn't want to dig up too much) - it didn't work as expected.

The lxdialog commit anyway has been modified to be a lil slimmer.

@nathanchance That stackprotector commit was an early test and ended up erroneously inside of this PR... it's not necessary, the intention was to keep it only locally for performance purposes.

@jerpelea this thingy has been forcepushed, you can go on

Thaodan:
@kholk 

```
if [ -n "${PRODUCT_SOONG_NAMESPACES}" ] && [ -n "${TARGET_PRODUCT}" ]; then
pkg-config()
{
        :
}
fi
```
Do something like this and skip the other changes in the commit.

kholk:
@Thaodan yes that's what I tried... and no, that seems to be forbidden on some versions of bash/sh..... specifically, some versions really dislike the dash in a function name

Thaodan:
It seems that POSIX doesn't allow dashes in function names. I tested this in dash and got:
```
$ pkg-config()
dash: 1: Syntax error: Bad function name
```

ix5:
@nathanchance you mean this?

```diff
--- Makefile
+++ Makefile
@@ -744,7 +744,7 @@ ifdef CONFIG_CC_STACKPROTECTOR
   stackp-path := $(srctree)/scripts/gcc-$(SRCARCH)_$(BITS)-has-stack-protector.sh
   stackp-check := $(wildcard $(stackp-path))
 endif
-KBUILD_CFLAGS += $(stackp-flag)
+KBUILD_CFLAGS += $(call cc-option,$(stackp-flag))
```

PR #2125, opened 2019-12-10
[LA.UM.7.1.r1] ARM64: dts: poplar: power the hybrid in-cell panel

Paulbouchara:
Previously, we had the full in-cell panel (9) being given a supply which
was needed for the hybrid in-cell panel (6). Probably a mistake during the
FBDEV -> DRM switch. Correct this.

Acked-by: MarijnS95 <marijns95@gmail.com>


Paulbouchara:
This commit have been ported from https://github.com/ZfSODP/kernel_sony_msm-4.9_kernel/commit/7142186615ce34cd5f54ad4e78418bd8c4a07240
Port it to 4.14.

This should fix https://github.com/sonyxperiadev/bug_tracker/issues/434.

MarijnS95:
Just noticed that I still had this change unstaged on my laptop - goes to show how long I haven't done kernel work from there :sweat_smile: 

Anyway, this has never been backported to 4.9 for those still using it, does anyone fancy PR'ing this one?

PR #2128, opened 2019-12-12
dts: lilac: add missing battery profiles

moralrecordings:
I have an XZ1C device (from about mid 2018) which, when flashed with Sony AOSP, would not show the correct charging percentage for the battery. Basically the type of battery was unknown, and it would always be stuck at 20% capacity.

```
130|lilac:/sys # cat ./devices/soc/800f000.qcom,spmi/spmi-0/spmi0-02/800f000.qcom,spmi:qcom,pmi8998@2:qpnp,fg/batt_info
Unknown Battery/331000/3/59 fa 07 13|00
lilac:/sys # cat ./devices/soc/800f000.qcom,spmi/spmi-0/spmi0-02/800f000.qcom,spmi:qcom,pmi8998@2:qpnp,fg/power_supply/bms/battery_type
Unknown Battery
lilac:/sys # cat ./devices/soc/800f000.qcom,spmi/spmi-0/spmi0-02/800f000.qcom,spmi:qcom,pmi8998@2:qcom,qpnp-smb2/power_supply/battery/capacity
20
```

After comparing the output of these queries on the stock kernel, I found out that there were two additional device tree profiles for 1308-1851 type batteries not included in the AOSP release. After ripping them out, converting them from dtb back to dts and adding them to the AOSP kernel device tree, the battery indicator works again as expected.

(The dtsi files in this commit aren't the original files; I extracted these from the stock Sony kernel.)

### DIFF
--- a/arch/arm/boot/dts/qcom/fg-gen3-batterydata-lilac-send-4245mv-2425mah.dtsi
+++ b/arch/arm/boot/dts/qcom/fg-gen3-batterydata-lilac-send-4245mv-2425mah.dtsi
@@ -1,0 +1,25 @@
+/*
+ * Copyright (c) 2017, The Linux Foundation. All rights reserved.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 2 and
+ * only version 2 as published by the Free Software Foundation.
+ *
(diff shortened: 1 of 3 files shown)


kholk:
Please PR this also to 7.1.r1.
Thanks!

PR #2130, opened 2019-12-12
[7.1.r1] xfs btrfs iversion fixes

Thaodan:
Add two commits that were missing to properly convert both file systems to the iversion API that was introduced earlier.


kholk:
Oops! Missed! Thanks!

PR #2132, opened 2019-12-14
[7.1.r1] dts: lilac: Add missing battery profiles

ix5:
See https://github.com/sonyxperiadev/kernel/pull/2128

Please test!


kholk:
There is really nothing to test on 4.14 that would be different from the test done already on the previous kernel. 

Thanks for the PR! 

Merge!

PR #2133, opened 2019-12-14
[LA.UM.7.1.r1] Boot, Ganges, boot!

MarijnS95:
The first small patches fix tiny quirks that prevented Ganges from booting at all, and making sure it stays alive.

The second set picks all the FPC/ET51x commits (#1912, #2010, #2017) from kernel 4.9, of which the sensor voltage bump to 3.3v is the most relevant for Ganges. Without it, the sensor is almost completely unresponsive and unusable.

Tested on:
Discovery RoW
Mermaid DSDS
Akatsuki DSDS (to make sure nothing in the FPC got borked)


kholk:
I totally approve!

PR #2134, opened 2019-12-15
[LA.UM.7.1.r1] [URGENT] SDM845: Disable inline rotator

MarijnS95:
The inline rotator was enabled but unfortunately freezes fullscreen video playback.

This reverts commit 87cde5ddf4b0274c432463b652781cd9254d083a.


jerpelea:
please leave it buggy so that we are forced to fix it properly

MarijnS95:
@jerpelea Sure, it was already disabled in userspace: https://github.com/sonyxperiadev/device-sony-common/pull/673/files#diff-8d1d4bc7ca3584dcde7d1ad9a9ff46d4R140

Not sure if we can or will ever fix this though.

PR #2145, opened 2019-12-21
[LA.UM.7.1.r1] arm64: DT: ganges-common: Upgrade to mainline WLED.

MarijnS95:
Ganges was accidentally left out in the conversion. Update the current
limit to the new prop name, and limit the strings to 2 (default for
pm660l is 3) to prevent nasty OVP interrupts as can be observed with:
grep wled /proc/interrupts


kholk:
GOGOGO :))))

PR #2011, opened 2019-06-16
[LE.UM.2.3.2.r1.4] arm64: DT: yoshino: Rename lilac/poplar pcc table nodes for SDE.

MarijnS95:
Fixes: https://github.com/sonyxperiadev/bug_tracker/issues/425

With the switch to SDE these nodes have been moved and renamed. The PCC
tables were overriding wrong/nonexistant nodes, leaving SDE panels
without PCC capabilities.

Note that Maple has the calibrations stored directly in the panel file.

### Before merging:
1. Are there other panels that need this treatment?
2. Do we want different behaviour when the UV read command is missing?

PR #2147, opened 2019-12-29
[LA.UM.7.1.r1] arm64: DT: yoshino: Rename lilac/poplar pcc table nodes for SDE.

stefanhh0:
@MarijnS95 this pull request pulls your PR #2011 for k4.14. See comment: https://github.com/sonyxperiadev/bug_tracker/issues/512#issuecomment-569419656. Would be great if you can confirm resp. comment on this.

Just did a kernel-build without flashing it, because lilac/yoshino is still not ready to my knowledge. So state is compiles without errors.

### DIFF
--- a/arch/arm64/boot/dts/qcom/dsi-panel-somc-lilac-id5_pcc.dtsi
+++ b/arch/arm64/boot/dts/qcom/dsi-panel-somc-lilac-id5_pcc.dtsi
@@ -18,8 +18,8 @@
  */
 
 /* pcc-table for Lilac LGD */
-&mdss_mdp {
-	dsi_5: somc,5_panel {
+&sde_kms {
+	sde_dsi_5: somc,5_panel {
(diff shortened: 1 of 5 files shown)


MarijnS95:
This would be necessary for the calibration, but shouldn't fix the other display issues Yoshino seems to be having.

That said, now that I know more about the DT, some things seem to be wrong (or at least slightly hiding the discrepancies):

- `dsi_5: somc,5_panel` and `8` equivalent define a new node (and get merged when a node with identical label is defined, afaik). This should instead be `&dsi_5` such that compilation actually fails when there is no node with that `dsi_5` label.
  That won't work with the PCC tables included _before_ the `*-base-sde.dtsi` file defines the actual nodes though..
- This breaks the non-SDE DT. Since it's unused, it might be best to remove those files altogether?

Pinging @kholk to provide answers :)

PR #2148, opened 2019-12-29
[VERIFYME][7.1.r1] fs: Improve eventpoll logging to stop indicting timerfd

ix5:
**VERIFY ME!!!**
Forwardport of https://github.com/sonyxperiadev/kernel/pull/1891

timerfd doesn't create any wakelocks; eventpoll can, and is creating the wakelocks we see called "[timerfd]".
eventpoll creates two kinds of wakelocks: a single top-level lock associated with the eventpoll fd itself, and one additional lock for each fd it is polling that needs such a lock (e.g. those using EPOLLWAKEUP).

Current code names the per-fd locks using the undecorated names of the fds' associated files (hence
"[timerfd]"), and is naming the top-level lock after the PID of the caller and the name of the file behind the first fd for which a per-fd lock is created.

### DIFF
--- a/fs/eventpoll.c
+++ b/fs/eventpoll.c
@@ -1394,15 +1394,23 @@ static int ep_create_wakeup_source(struct epitem *epi)
 {
 	const char *name;
 	struct wakeup_source *ws;
+	char task_comm_buf[TASK_COMM_LEN];
+	char buf[64];
+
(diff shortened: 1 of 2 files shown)


oshmoun:
tested, seems to work fine

PR #2151, opened 2020-01-02
[2.3.2.r1.4] dts: tone: Fix gpio_11->12 in somc_pinctrl_pmic

ix5:


kholk:
Please check if that's required on 4.14 as well (but I think I've fixed it there).

In the meanwhile, should be safe to merge

ix5:
You you already fixed it, that's how I discovered it in fact ;)

stefanhh0:
```
$ grep -l  ".*gpio_11\ .*gpio_11" *.dtsi
msm8996-mtp.dtsi
msm8996-tone-common.dtsi
```
Shouldn't it be fixed then also in the msm8996-mtp.dtsi as well? The file has the same problem there as well as in 4.14.

PR #2154, opened 2020-01-04
[2.3.2.r1.4] Fix missing POWER_SUPPLY_PROP_SDP_CURRENT_MAX getter

vknecht:
Should fix following message on Loire (and probably Tone too):
[    1.302239] power_supply usb: driver failed to report `sdp_current_max' property: -22
While not seemingly harmful on Android, on SailfishOS it led to log flood
and eventually reboot.


kholk:
Should be fine, as SDP is not for charger modes.

PR #2163, opened 2020-01-22
[7.1.r1] arm64: DT: MSM8998: Fix mailboxes for GLINK transports

kholk:
The GLINK transports had wrong mailbox cells declared: fix them.

Modem and DSP should now work on MSM8998.


stblr:
I just tested it, and it definitively fixes sonyxperiadev/bug_tracker#512.

PR #2157, opened 2020-01-13
[LA.UM.7.1.r1] Initial update to the latest 4.14 version

nathanchance:
This updates the Linux version from 4.14.84 to 4.14.164 (latest version at the time of writing this).

I do not see any build regressions with this using `clang-r353983c`. `aosp_loire_blanc_defconfig` and `aosp_yoshino_poplar_defconfig` do not build for me locally, not sure if that is the case for everyone. 

Unfortunately, just because everything builds fine doesn't mean that something won't happen at run time :) If there is some sort of run time issue, please bisect to find which version causes the issue since there are a LOT of changes here.

PR #2166, opened 2020-01-25
[URGENT] [LA.UM.7.1.r1] Fast forward to 4.14.160

MarijnS95:
This is an except from https://github.com/sonyxperiadev/kernel/pull/2157, which is merged to master in batches. Master is currently at .150 which breaks a couple of our most-stable devices.

I observe Akatsuki stuck on boot, and @kholk observes display problems. With no time to validate which patch exactly fixes it (and there really being no point, given that _it's fixed_), let us move to .160.


kholk:
Merge it instantly. Fixes breakages introduced by 4.14.150.

nathanchance:
Is there any reason I can't just rebase #2157 on Alin's cherry pick of 4.14.150 and just merge up to 4.14.166? Are there any more outstanding regressions?

kholk:
I don't know if there are other regressions, but the only daily usable platform right now is Tama, and we know that 160 makes it work again as it should.

For this reason, it's urgent to merge this. 

Kernel patch level upgrades should be tested very carefully so, while I agree that it'd be just great to directly go for 166, we don't know what regressions and what fixes will this bring on the many platforms that we support right now, while we know the status on 160.

In few words, strategy is "keep the quality as high as possible and do it slow to achieve that".

MarijnS95:
@nathanchance The main reason is timezones.

There are certain projects that rely on this exact kernel repo instead of creating a fork and maintaining their own update schedule.

.150 has apparently been accidentally deemed "okay enough" to be merged to master despite Angelo and me seeing different issues on Tama. Given that it was nighttime for you, and Alin likely not close to a (remote) machine the best we can do is submit a PR to fast-forward to a tested working version, that can be merged with a single click. As you can tell, that didn't happen in any timely manner though.

PR #2174, opened 2020-02-01
[LA.UM.7.1.r1] arm64: DT: sdm660-audio: Shorten cdc-vdd-mic-bias name length

MarijnS95:
The name of a regulator consumer consists of the combined name of the DT
node hierarchy and the name of the regulator property within this node,
concatenated with a dash.
Together they can (including null character) not exceed the max length
denoted by REG_STR_SIZE, currently 64 chars. However, the full path of
the analog-codec node: 800f000.qcom,spmi:qcom,pm660l@3:analog-codec@f000
together with cdc-vdd-mic-bias and a dash inbetween results in a name of
67 characters (including \0).

Shorten the length of the mic-bias supply name to fit within this limit,
making the driver probe correctly and vote for this supply which is
necessarry for audio recordings to have enough bias and not sound
muffled.

WARNING: The same change has to be applied in techpack/audio, which
looks for this specific name.

### DIFF
--- a/arch/arm64/boot/dts/qcom/sdm660-audio.dtsi
+++ b/arch/arm64/boot/dts/qcom/sdm660-audio.dtsi
@@ -159,16 +159,16 @@
 		qcom,cdc-vdd-pa-voltage = <2040000 2040000>;
 		qcom,cdc-vdd-pa-current = <260000>;
 
-		cdc-vdd-mic-bias-supply = <&pm660l_l7>;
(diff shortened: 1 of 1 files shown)


kholk:
Solves microphone gain issues on SDM630/660.

Merge!
P.S.: Don't forget https://github.com/sonyxperiadev/kernel-techpack-audio/pull/15

PR #2178, opened 2020-02-04
[7.1.r1] FPC wakeup fix

Haxk20:
This fixes the issue where IRQ signal is being masked by the driver resulting in device not waking up when FPC is pressed. This sadly does not fix calls not waking up the device tho.
@kholk Please look at this and tell me what you think of this.


kholk:
Eh, it's not a good idea, but it's not a bad one either. On 4.14 things are a little different.

Well, that should be tested carefully, as it may raise unwanted (and useless) wakeups... test it.

Haxk20:
@kholk Tested for 2 days. All calls are there as they should be and FPC wakes up the phone as it should.
But if you have better idea how to fix it please tell me.

Haxk20:
Tested. Battery seems untouched by this change.

kholk:
Let's merge it then :)))

Haxk20:
@kholk I think @jerpelea can help with that. :)

PR #2179, opened 2020-02-04
[7.1.r1] msm: vidc: Fix warnings

oshmoun:
A commit picked from 4.9 that adds the missing format descriptions to avoid having v4l2 spew out errors about unknown formats. It consists of a change in vidc_3x and a change in v4l2 core, but the former is already applied in the snapshot vidc_3x, so only the latter change is applied in this PR.


jerpelea:
for reference :
https://android.googlesource.com/platform/hardware/qcom/msm8x74/+/476f54d5431b8617cc52c020e6494b74e3db58cb/kernel-headers/linux/videodev2.h#274

PR #2181, opened 2020-02-08
[VERIFYME][7.1.r1] Forwardport HALT_WARNONLY and use BRANCH_HALT_WARNONLY for gpucc_gfx3d_clk

Haxk20:
Forward port BRANCH_HALT_WARNONLY and set it for gfx3d clock.
This has been set back to default state in the process of porting to k4.14
It seems its still needed as the clock halts and gets stuck in some scenarios.
@kholk As always say what you think about it.

### DIFF
--- a/drivers/clk/qcom/clk-branch.c
+++ b/drivers/clk/qcom/clk-branch.c
@@ -85,13 +85,6 @@ static int clk_branch_wait(const struct clk_branch *br, bool enabling,
 	const char *name = clk_hw_get_name(hw);
 
 	/*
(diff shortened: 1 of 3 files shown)


ix5:
For reference https://github.com/sonyxperiadev/kernel/pull/2001

Haxk20:
@ix5 Im more then aware of that PR. The thing is tho that we have white screen bug even without those commits and its the only difference between 4.9 and 4.14 in those files. (On 4.9 the clock was stable and fine)

Haxk20:
Not the solution we wanted and its not really a solution. Needs more looking into.

PR #2182, opened 2020-02-08
[2.3.2.r1.4] msm: kgsl: Three patches from the Feb Android bulletin

ix5:
https://source.android.com/security/bulletin/2020-02-01#qualcomm-components

4.14 seems unaffected, we're already patched up (thanks @nathanchance!)

State, as far as I could determine: https://sx.ix5.org/info/post/sodp-2020-02-05-security-state/

<strike>Also part of the Feb patches: https://github.com/sonyxperiadev/kernel/pull/2183</strike> -> Not necessary

---

The `dword` change in `adreno_ringbuffer.c` seems bad for legacy to me?

Build-tested for kagura.

### DIFF
--- a/drivers/gpu/msm/adreno_ringbuffer.c
+++ b/drivers/gpu/msm/adreno_ringbuffer.c
@@ -1,4 +1,4 @@
-/* Copyright (c) 2002,2007-2018, The Linux Foundation. All rights reserved.
+/* Copyright (c) 2002,2007-2018,2019, The Linux Foundation. All rights reserved.
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License version 2 and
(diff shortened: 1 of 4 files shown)


ix5:
Kagura boots fine

drivers/gpu/msm/adreno_ringbuffer.c
```diff
@@ -937,14 +973,16 @@ int adreno_ringbuffer_submitcmd(struct adreno_device *adreno_dev,
...
 
 		/*
-		 * REG_TO_MEM packet on A5xx and above needs another ordinal.
-		 * Add 2 more dwords since we do profiling before and after.
+		 * User side profiling uses two IB1s, one before with 4 dwords
+		 * per INDIRECT_BUFFER_PFE call
 		 */
-		if (!ADRENO_LEGACY_PM4(adreno_dev))
-			dwords += 2;
+		// FIXME: Felix: Removing this seems like it'll give legacy too many dwords...
```
ix5: Could use some guidance here?

kholk: Check 4.14... That's a nice find btw...

ix5:
```
/* Return true if the hardware uses the legacy (A4XX and older) PM4 format */
#define ADRENO_LEGACY_PM4(_d) (ADRENO_GPUREV(_d) < 500)
```
[adreno_pm4types.h](https://github.com/sonyxperiadev/kernel/blob/aosp/LA.UM.7.1.r1/drivers/gpu/msm_legacy/adreno_pm4types.h#L247-L248)
Guess we don't need to care about a4xx?

Will align with 4.14 then (for reference: https://github.com/sonyxperiadev/kernel/blob/aosp/LA.UM.7.1.r1/drivers/gpu/msm/adreno_ringbuffer.c#L1010)

stefanhh0:
Compiles, boots and runs fine on yoshino/lilac on top of latest 4.9.213

stefanhh0:
@ix5 [WIP] can be removed? Or still something open?

ix5:
@stefanhh0 I really wouldn't like to break other devices, but it seems this is ok to merge now.

PR #2187, opened 2020-02-11
[LA.UM.7.1.r1] Reintroduce merged PCCs for night light

### LINKED ISSUE #522: [tama] [apollo] kernel 4.14 Night Light crashes after sleep
After waking the device the colorization of the screen goes away till the Night Light mode gets reactivated. Issue present on both AOSP with f.lux installed and PixelExperience (built from current Xperia AOSP sources).

MarijnS95:
Fixes https://github.com/sonyxperiadev/bug_tracker/issues/522

As it turns out the implementation wasn't really caring too much about validation, running into some broken state. Fix these issues by adding more validation, fallbacks and getting rid of stupid code.

Regarding the u,v command, it could possibly be wrong, unless the panels/dric's found in Kumano really do not contain any calibration value. If so `somc,mdss-dsi-pcc-force-cal` _should_ be added (and we can shave down the calibration table size quite a bit :wink: ).

Tested on Akatsuki DSDS and Griffin DSDS.

### DIFF
--- a/drivers/gpu/drm/msm/dsi-staging/somc_panel/panel_color_manager.c
+++ b/drivers/gpu/drm/msm/dsi-staging/somc_panel/panel_color_manager.c
@@ -724,11 +724,11 @@ static int somc_panel_sde_crtc_set_property_override(struct drm_crtc *crtc,
 		pr_debug("Merging system calibration\n");
 		memcpy(&color_mgr->system_calibration_pcc, blob->data,
 				blob->length);
-		color_mgr->system_calibration_valid = true;
 
 		ret = somc_panel_update_merged_pcc_cache(color_mgr);
 		if (ret)
 			return ret;
+		color_mgr->system_calibration_valid = true;
 	}
 
 	// Copy (updated) cache to blob:
@@ -1335,6 +1335,7 @@ static int somc_panel_crtc_send_cached_pcc(struct dsi_display *display)
 static int somc_panel_send_pcc(struct dsi_display *display,
 			       int color_table_offset)
 {
+	int rc;
 	struct somc_panel_color_mgr *color_mgr =
 			display->panel->spec_pdata->color_mgr;
 
@@ -1348,7 +1349,9 @@ static int somc_panel_send_pcc(struct dsi_display *display,
 
 	color_mgr->pcc_profile = color_table_offset;
 
-	somc_panel_update_merged_pcc_cache(color_mgr);
+	rc = somc_panel_update_merged_pcc_cache(color_mgr);
+	if (rc)
+		return rc;
 
 apply_cached_pcc:
 	return somc_panel_crtc_send_cached_pcc(display);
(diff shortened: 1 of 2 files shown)


kholk:
Looks good. I'll check about the u,v commands on the new DrICs, but I suspect that there's nothing like this anymore.
The PCC table there is being used in a different way compared to the LCD implementations: now they have a calibration curve in function of display brightness to address the color drift that we saw on Akatsuki.

I solved that with a specific linear color curve (hue, value, contrast) variation that starts at low brightness values because that Akatsuki display is sort of fine with that, but they opted for a way more precise calibration done the other way around, which is nicer (but requires lots of time and right costful equipment to do, so that's why I didn't do that on Akatsuki).

We will have to switch to the new way on Kumano... Probably a DT property advertising this kind of table being in place, which will ignore the u,v read and vary the calibration code path accordingly (we can reuse, of course, the akatsuki style path, but instead we would be just reading and writing to the PCC block instead).
I mean, I already have implemented something listening for brightness changes in somc_panel, so it's very easy to write the new stuff...

Anyway, merge this!

MarijnS95:
> I solved that with a specific linear color curve (hue, value, contrast) variation that starts at low brightness values because that Akatsuki display is sort of fine with that, but they opted for a way more precise calibration done the other way around, which is nicer (but requires lots of time and right costful equipment to do, so that's why I didn't do that on Akatsuki).

Adacolor, through HSIC, right?
 
> We will have to switch to the new way on Kumano... Probably a DT property advertising this kind of table being in place, which will ignore the u,v read and vary the calibration code path accordingly (we can reuse, of course, the akatsuki style path, but instead we would be just reading and writing to the PCC block instead).

Besides such an implementation I guess we want whitepoint calibration back in ES?

```c
0x00 0x70 0x20 0x23 0x1C 0x1F 0x7300 0x7260 0x8000 /* 6000K */
0x00 0x71 0x1C 0x1F 0x1C 0x1F 0x8000 0x7AA0 0x6AD0 /* F6: 4150K */
0x00 0x72 0x18 0x1B 0x1C 0x1F 0x8000 0x7F00 0x7F00 /* D50: 5000K */
0x00 0x73 0x14 0x17 0x1C 0x1F 0x6FF0 0x6E00 0x8000 /* D65: 6500K */
```
That's the only interesting block in Akatsuki PCC DT, which is always selected based on the u,v our panels report. We would need to replicate something like that with u,v min and max set to 0 and 0xff respectively (and obviously different calibration colors).

Anyway that means another merge, as we'll get three components feeding their color transformation into the PCC block:
- Android system (night shift etc);
- Whitepoint selection;
- Kumano-specific calibration, as you say.

Sounds like fun. We'll need to abstract away and generalize the currently messy PCC codepaths :stuck_out_tongue_winking_eye:

kholk:
Yes it's what I've called ADAptiveCOLOR and yes it's through the HSIC.

And... Yes it's "" "fun" "".....

PR #2189, opened 2020-02-13
[7.1.r1] Various fixes for all

kholk:
Fix the ICP memory maps on Kumano, add sm8150 camcc variation from copylefts,
use a different supply for the DisplayPort USB QMP and fix softirqs while parking a
thread.

Needs to be tested on all platforms.


kholk:
Uhm...the chat is gone. Btw, commit added. Enjoy.

PR #2191, opened 2020-02-15
[7.1.r1] Fix IPA/GSI/WLAN and system freezes on SoMC SM8150 Kumano

kholk:
This patchset fixes FAST pagetables on SM8150, which fixes IPA, GSI, WLAN.
Also, fix the system freezes on this platform seen while "just sitting".

Please check for breakages on Tama and Nile platforms before merging.

Enjoy!


kholk:
Title and description changed after adding the solution to stop freezes on Kumano platform.

PR #2192, opened 2020-02-15
[2.3.2.r1.4] Linux 4.9.214

stefanhh0:
Applied treewide Linux 4.9.214. Only that nothing more.

~~while applying I came across a patch for pci-keystone.c. I realized then that the pci/dwc was pulled in from linux 4.17 two years ago. Since 4.17 is no longer maintained, I think it would be good to switch to a maintained version, and thus pulled in latest 4.14.171 specifically for pci/dwc. I am not sure if that is correct or breaks something. It compiled without errors in the kernel tree. I will test asap on yoshino/lilac.~~


stefanhh0:
Testing it currently on yoshino/lilac. Boots fine and runs fine for now.

kholk:
The dwc from 4.14 doesn't have good support for what we need.
Also as far as I remember I've had to make some changes in order to get the right
supplies (and sequence) for 8996.

In any case, this driver is being used only on Tone, so testing it on Yoshino is totally useless...

ix5:
<del>Tone's wifi seems broken with this.</del>
Wrong PR again, the change to the PCI-E drivers is the breaking one https://github.com/sonyxperiadev/kernel/pull/2195

stefanhh0:
I have investigated a little with gitk and can now see whats wrong, thanks for giving me the feedback.

I am currently investigating what has been done on k4.19 and it seems at first glance that pulling in the latest from using k4.19 seems to be the better choice. I am currently heavily diffing 
2.3.2.r1.4 `drivers/pci/dwc` and v4.19.104 `drivers/pci/controller/dwc` (yes, it was moved again). I'll force push something different, hopefully better soon.

I am a little time restricted, so it may take a few days...

stefanhh0:
Or even better I will make it a different PR, so that 4.9.214 can be tested and integrated independently from the dwc changes...

stefanhh0:
This patch runs stable on my phone for more than 10 days, do others could find some time to test it? If I am not the only one who tested it successfully, then it could be merged back, right?

kholk:
The only platform to be tested before deciding to merge this is Loire, due to some changes in this patch that should anyway be safe.

Please, anyone, test loire and report back so that we can merge this.

P.S.: Can you fix the PR description, since you've removed the pcie changes?

jerpelea:
@stefanhh0 loire test ?

stefanhh0:
I don't have a Loire, if someone else could help here that would be great.

@chris42 which kernel are you currently running on your Loire/Kugo, if you could give this PR a try and report here it would be awesome.

chris42:
Currently running the regular 4.9.213. Will test it and report back what I got.

chris42:
Just tried the PR on Kugo: Device boots up and reports 4.9.214, connects wifi etc. Could not spot anything unusual.
You want me to look out for something special?

stefanhh0:
@chris42 thanks a lot for testing. The functional testing you have done seems to me sufficient.
@jerpelea Then this one can be merged.

kholk:
There's nothing special to look for.
As long as the phone works as before, everything is fine for me.

This is most probably safe to merge. Go for it.

PR #2195, opened 2020-02-18
[2.3.2.r1.4] pci: dwc: Linux 4.19.104

stefanhh0:
An attempt to update dwc sources to 4.19.104. Reason for that is that dwc was pulled in originally from linux-4.17.y branch which is no longer maintained. So it would be good to switch to sources that are longterm maintained as it the case with linux-4.19.y branch. 

However I can't test it properly. Nor would I be able to fix any problems with it.

@ix5 Could you test it please? If it is not working as it should, then I would need help or in case no one can help I'd then just close this update attempt.

The change itself compiles fine.


ix5:
This breaks tone wifi

PR #2197, opened 2020-02-18
[LA.UM.7.1.r1] media: sde: rotator: Remove legacy swts code without hwts guard

MarijnS95:
The rotator driver attained hardware timestamp support on 4.14 but the
ported quirks for sdm6xx were not hidden behind this feature flag. This
meant a rogue sde_hw_rotator_swts_map would have executed on platforms
with hwts causing undefined behaviour such as panics due to accessing
uninitialized pointers, as sde_hw_rotator_swts_create is never called.

Merging the logic together causes sde_hw_rotator_swts_map to become
superfluous as sde_hw_rotator_swts_create - which maps the buffer too -
is always called when no HW timestamp support is available and the
swts_buf is not yet mapped.

Also remove the unused cpu-mapped swts_buffer from ported code,
which has been removed upstream from sde_hw_rotator_swts_create too [1].
Note that the struct and print statements retain this legacy field.

[1]: 6e2d3ce466b50ca43124f9e62177feefe9647bb5

---

This fixes a crash observed when opening Open Camera on Kumano, to test freshly added camera support in https://github.com/sonyxperiadev/kernel/pull/2196

### DIFF
--- a/drivers/media/platform/msm/sde/rotator/sde_rotator_r3.c
+++ b/drivers/media/platform/msm/sde/rotator/sde_rotator_r3.c
@@ -2525,14 +2525,6 @@ static int sde_hw_rotator_swts_map(struct sde_hw_rotator *rot)
 		goto err_unmap;
 	}
 
-	dma_buf_begin_cpu_access(data->srcp_dma_buf, DMA_FROM_DEVICE);
-	rot->swts_buffer = dma_buf_kmap(data->srcp_dma_buf, 0);
-	if (IS_ERR_OR_NULL(rot->swts_buffer)) {
-		SDEROT_ERR("ion kernel memory mapping failed\n");
-		rc = IS_ERR(rot->swts_buffer);
-		goto kmap_err;
-	}
-
 	data->mapped = true;
 	SDEROT_DBG("swts buffer mapped: %pad/%lx va:%p\n", &data->addr,
 			data->len, rot->swts_buffer);
@@ -2559,10 +2551,6 @@ static void sde_hw_rotator_swts_unmap(struct sde_hw_rotator *rot)
 	struct sde_mdp_img_data *data;
 
 	data = &rot->swts_buf;
-
-	dma_buf_end_cpu_access(data->srcp_dma_buf, DMA_FROM_DEVICE);
-	dma_buf_kunmap(data->srcp_dma_buf, 0, rot->swts_buffer);
-	rot->swts_buffer = NULL;
 
 	sde_smmu_unmap_dma_buf(data->srcp_table, SDE_IOMMU_DOMAIN_ROT_UNSECURE,
 			DMA_FROM_DEVICE, data->srcp_dma_buf);
--- a/drivers/media/platform/msm/sde/rotator/sde_rotator_r3.c
+++ b/drivers/media/platform/msm/sde/rotator/sde_rotator_r3.c
@@ -2449,28 +2449,36 @@ static int sde_hw_rotator_swts_create(struct sde_hw_rotator *rot)
 
 	data = &rot->swts_buf;
 	data->len = bufsize;
-	data->srcp_dma_buf = sde_rot_get_dmabuf(data);
+
 	if (!data->srcp_dma_buf) {
-		SDEROT_ERR("Fail dmabuf create\n");
-		return -ENOMEM;
+		data->srcp_dma_buf = sde_rot_get_dmabuf(data);
+		if (!data->srcp_dma_buf) {
+			SDEROT_ERR("Fail dmabuf create\n");
+			return -ENOMEM;
+		}
 	}
 
 	sde_smmu_ctrl(1);
 
-	data->srcp_attachment = sde_smmu_dma_buf_attach(data->srcp_dma_buf,
-			&rot->pdev->dev, SDE_IOMMU_DOMAIN_ROT_UNSECURE);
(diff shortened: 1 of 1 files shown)


kholk:
Eh, are you sure you're not breaking Nile?

Actually, the issue is that this is a misport... needs to be fixed properly....

MarijnS95:
> Eh, are you sure you're not breaking Nile?

I did not manage to break it (Discovery, Mermaid) with these changes. That is, I don't see how `sde_hw_rotator_free_ext` is triggered to share the SMMU with the MDP as indicated by the original port 6e2d3ce466b50ca43124f9e62177feefe9647bb5.

> Actually, the issue is that this is a misport... needs to be fixed properly....

I think I see what you mean. The code in 4.9 only creates swts resources when `iclient` is NULL, which is initialized once and never unset. Meaning a successive `sde_hw_rotator_free_ext` + `sde_hw_rotator_alloc_ext` only unmaps and remaps the dma_buf without ever detaching it in `sde_hw_rotator_swts_destroy`.

https://github.com/sonyxperiadev/kernel/blob/dbfc0e468e4edb9dec86c4dda75ddfc3b0b3f626/drivers/media/platform/msm/sde/rotator/sde_rotator_r3.c#L2443-L2453

On 4.14 the same happens, except that there is no `iclient` anymore. Instead, `sde_hw_rotator_alloc_ext` only checks if the buffer is unmapped, and reattaches the entire thing which is an artifact from [other](https://github.com/sonyxperiadev/kernel/commit/6e2d3ce466b50ca43124f9e62177feefe9647bb5) [changes](https://github.com/sonyxperiadev/kernel/commit/0931849cf9c764bde12cf8cf8c1960c19297a82b).

https://github.com/sonyxperiadev/kernel/blob/d4dc6670ed82258d2f0c068feac15588f39b369d/drivers/media/platform/msm/sde/rotator/sde_rotator_r3.c#L2762-L2774

This likely causes all sorts of breakage, so I think you want me to guard `sde_hw_rotator_swts_create` against attaching and pinning the smmu dma_buf again when it was not unpinned and detached first?

kholk:
Please keep this PR open for another 2-3 weeks....

ix5:
Merge conflict, is this PR still relevant?

MarijnS95:
It does some vital cleanup but has been "super"seded by another patch. I'll see if I can find time to rebase those changes off, leaving only the cleanup.

MarijnS95:
Force-pushed as requested. The first commit is relevant, the second one... Perhaps we should keep cleanup in `sde_hw_rotator_swts_{create,destroy}` but omit lazy creation in `sde_hw_rotator_swts_create` which was sort of superseded by #2234.

pablomh:
Hi @MarijnS95. This has been in my list of patches to apply and hasn't introduced any noticeable regression in my pioneer (nile platform).

drivers/media/platform/msm/sde/rotator/sde_rotator_r3.c
```diff
@@ -2449,28 +2449,36 @@ static int sde_hw_rotator_swts_create(struct sde_hw_rotator *rot)
...
 
 	sde_smmu_ctrl(1);
 
-	data->srcp_attachment = sde_smmu_dma_buf_attach(data->srcp_dma_buf,
-			&rot->pdev->dev, SDE_IOMMU_DOMAIN_ROT_UNSECURE);
-	if (IS_ERR_OR_NULL(data->srcp_attachment)) {
-		SDEROT_ERR("sde_smmu_dma_buf_attach error\n");
-		rc = -ENOMEM;
-		goto err_put;
+	if (!data->srcp_attachment) {
```
kholk: IS_ERR_OR_NULL check here

drivers/media/platform/msm/sde/rotator/sde_rotator_r3.c
```diff
@@ -2449,28 +2449,36 @@ static int sde_hw_rotator_swts_create(struct sde_hw_rotator *rot)
...
+		}
 	}
 
-	data->srcp_table = dma_buf_map_attachment(data->srcp_attachment,
-			DMA_BIDIRECTIONAL);
-	if (IS_ERR_OR_NULL(data->srcp_table)) {
-		SDEROT_ERR("dma_buf_map_attachment error\n");
-		rc = -ENOMEM;
-		goto err_detach;
+	if (!data->srcp_table) {
```
kholk: IS_ERR_OR_NULL check here too

kholk (commented):
Please apply the requested changes, then let's merge this one.

MarijnS95:
@kholk Fair point, but then again do we want to keep this lazy initialization? It doesn't seem strictly necessary anymore, but would prevent us from having unreferenced resources sitting around in the extremely unlikely event that calls to `create` and `destroy` are unbalanced.

kholk:
You have two choices:
1. Create synthetic test cases and make a strict validation of the entire tangled code; or
2. Better lazy-safe than sorry.

You got what I chose. Would you choose the option N.1? :p

MarijnS95:
Sure. The point being that this lazy initialization was ~ necessary ~ in the previous version of these patches, before #2234 came around. Now it's more-or-less cleanup than anything, we have no confirmation that this odd case I just made up ever happens.

Let's go for ⚠️ SAFETY ⚠️ 🎉 🎉 🎉 🎉 🎉

PR #2202, opened 2020-02-20
 [7.1.r1] arm64: dts: Add missing battery profiles for Poplar

### LINKED ISSUE #542: [Yoshino][Poplar] Battery stuck at 20% when flash sodp 9 / 10
**Platform:** Yoshino
**Device:** Poplar
**Kernel version:** 4.14 / 4.9
**Android version:** 10 / 9
**Software binaries version:** v3a / v9

**Previously working on** 
it work the first time i flash sodp build ( the battery display was normal)

**Description**
when i flash back to stock with newflasher and flash back to sodp the battery stuck at 20%

**Symptoms**
Write the symptoms here.

**How to reproduce**
flash sodp build, then flash stock with newflasher with presist partition flashed, then flash sodp again

**Additional context**
when i run logcat i found this:
11-29 14:45:13.168   621   621 E /vendor/bin/hw/android.hardware.health@2.0-service.sony: Cannot read battery capacity persist file from /mnt/vendor/persist/battery/battery_charge_full: No such file or directory

stblr:
Same thing as #2128, but for Poplar.

This may fix sonyxperiadev/bug_tracker#542, but needs to be tested (I don't have a Poplar).

### DIFF
--- a/arch/arm64/boot/dts/qcom/fg-gen3-batterydata-poplar-send-4245mv-2425mah.dtsi
+++ b/arch/arm64/boot/dts/qcom/fg-gen3-batterydata-poplar-send-4245mv-2425mah.dtsi
@@ -1,0 +1,88 @@
+/* Copyright (c) 2018, The Linux Foundation. All rights reserved.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 2 and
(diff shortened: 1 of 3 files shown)


yuezhenglingluan:
i will try to make a boot.img with this and try to see if the problem is fix soon

yuezhenglingluan:
Yes it fixed the issue

PR #2207, opened 2020-02-23
[7.1.r1] Various small performance improvements

kholk:
This patchset improves the performance of some crypto algorithms using AArch64 crypto extensions instructions.

Also, there are various generic and aa64 specific optimizations that aim to squeeze more performance. 

Tested on SoMC SDM845 Tama Akatsuki RoW.
Please, give it a test on SDM630/636 platform as well.

P.S.: Depends on my other PRs from yesterday (and I'm sorry, I'm on phone and copypasting references is not really great like this).


MarijnS95:
Copypaste: depends on https://github.com/sonyxperiadev/kernel/pull/2205 only.

It has been running fine on Discovery ROW, Mermaid DSDS, Bahamut DSDS and Griffin DSDS :)
And of course running fine for over a day on Akatsuki DSDS.

PR #2211, opened 2020-03-02
[7.1.r1] Linux 4.14.172

nathanchance:


kholk:
WHOA!! This is one of the most dangerous upgrades that I've ever seen here!!

Test this VERY carefully!

jerpelea:
MERGE MERGE MERGE :)

MarijnS95:
This seems to work fine on most devices (Akatsuki DSDS, Griffin DSDS, Mermaid DSDS, Discovery RoW), apart that `dmesg -w` / `cat /dev/kmsg` do not follow/wait for more output. After printing all lines the final `read()` never returns when more data becomes available. Would you mind checking why that happens or if it is fixed in .173?

nathanchance:
@MarijnS95 Sorry, forgot about that regression. I've pushed the fix that will be in 173 so that it does not regress.

PR #2212, opened 2020-03-04
[7.1.r1] Fix various issues, add more legacy support

kholk:
This rather huge patchset adds support code for legacy SoCs, such
as MSM8956/8996.
Also, there are some fixes for some of the SDM630/660 interrupts issues and
better support for LMH DCVS for MSM8998 and SDM630/636/660.
Last but not least, various sanitizations for all supported SoCs, including the
non-legacy ones.

Please test carefully on all platforms before merging, as this is very big and
I've got some time issues.

Enjoy!

PR #2213, opened 2020-03-06
[7.1.r1] Upgrade the camera_v2 driver

kholk:
This is an upgrade of the camera_v2 driver and it's the second part of
the split #2212 .

Splitting the pull request deemed to be necessary due to the fact that the
camera_v2 upgrade will take some days to get aligned.

Enjoy!


jerpelea:
tested with v6 and works as it should

PR #2214, opened 2020-03-07
[LA.UM.7.1][VERIFYME]Remove androidboot.mode=cei_charger flag from /proc/cmdline

Haxk20:
Xperia 10/10 Plus bootloader sets androidboot.mode=cei_charger, breaking offline charging since the flag isn't supported by AOSP. Removing it so that userspace will parse flag set by sony_param_warmboot() instead fixes it.

Change-Id: I8925d3b6e102902e0e31fda25ffe568fcd37185a
This commit has been tested by LuK on ganges for his port and works to fix the offline charging but this has to be yet tested on SODP. Which i will do tomorrow.

### DIFF
--- a/arch/arm64/kernel/setup.c
+++ b/arch/arm64/kernel/setup.c
@@ -283,6 +283,24 @@ static int __init sony_param_warmboot(char *p)
 	return 0;
 }
 early_param("warmboot", sony_param_warmboot);
+ 
+/*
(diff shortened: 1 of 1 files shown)


Haxk20:
Tested. While the phone has its screen always on and there is no LED indication of charging it looks like the phone is actually offline charging.

PR #2216, opened 2020-03-11
[2.3.2.r1.4] Remove problematic Android.mk file

chris42:
Follow @kholk changes to remove `drivers/input/touchscreen/siw/mon/Android.mk` as it produces errors on legacy build.
See [Build System Changes](https://android.googlesource.com/platform/build/+/master/Changes.md#phony_targets)

Should be merged together with [device-sony-common/PR704](https://github.com/sonyxperiadev/device-sony-common/pull/704)

Kernel are now prebuilt by script and driver apparently is build with kernel, so files are not needed.

I am currently testing this for Kugo/Loire. Other device/platform tests would be nice.

### DIFF
--- a/drivers/input/touchscreen/siw/mon/Android.mk
+++ b/drivers/input/touchscreen/siw/mon/Android.mk
@@ -1,6 +1,0 @@
-LOCAL_PATH := $(call my-dir)
-include $(CLEAR_VARS)
-LOCAL_MODULE              := ssw_mon.ko
-LOCAL_MODULE_TAGS         := optional
-LOCAL_MODULE_PATH         := $(KERNEL_MODULES_OUT)
-include $(DLKM_DIR)/AndroidKernelModule.mk


MarijnS95:
No need for tests on other platforms, this is a sane thing to do. However, it will remove inline kernel compilation for AOSP 9.0 entirely, so this might not be desired.

One can also set `BUILD_KERNEL := false` or never set it to true in the first place in common.

chris42:
Are the build system changes only affecting Android 10? I thought these are general changes, that would also affect previous Android inline builds? Hence you would need to prebuilt the kernel on 9 as well?

stefanhh0:
Yes the changes only affected Android-10. For Android-9 we still have inline kernel building available and it should be kept in shape for those that are still using it. This PR is not required to get the issue itself resolved, correct? If so I'd vote to not integrate this one and just close it.

chris42:
Happy to do so, would probably make sense to then set the inline build to false in default? Guess more people are looking into using prebuilt kernel via the gcc script, as Loire and others already had the green camera tint problems on Android 9.
Hence the inline clang should be rare? Do I assume correctly?

kholk:
Some Android 10 custom ROMs are building the kernel inline anyway.

Please remove only the SIW Android.mk for now.

chris42:
Ok, dropped the regular Android.mk out of the PR. Also set [kernel/PR704](https://github.com/sonyxperiadev/device-sony-common/pull/704) ready for review. I think they both should go together.

kholk:
This pr can be merged now

ix5:
We should probably remove the `Android.bp` files from 4.9 and 4.14 as well, they're unused and cause soong to throw errors because a module is defined twice.
And no it would be stupid to define a namespace for unused crap ;)

PR #2217, opened 2020-03-16
[LA.UM.7.1.r1] [REQUIRESTESTING] sdm660: Enable gfx3d_src parent clocks

### LINKED ISSUE #559: [kirin] [aosp 10] [kernel 4.14] GPU-related crashes hang the device
**Platform:** ganges
**Device:** kirin (i4113)
**Kernel version:** 4.14.172
**Android version:** android-10.0.0_r30
**Software binaries version:** 10.0.7.1_r1_v4c_ganges

**Previously working on**
Not sure, this is the first time I use an Android 10 / Kernel 4.14 build on this device for any amount of time significant enough to notice these GPU crashes.

**Description**
Sometimes, apparently in a random manner, after a period of time being in use, the phone becomes unresponsive to any inputs (power button, fingerprint reader) and the screen becomes fully white or grey. The application being used does not seem to matter, it has happened multiple times with different applications.

**Symptoms**
Fully white or grey screen, no response to input on physical buttons, device reachable via adb.

**How to reproduce**
Use the device normally for a few hours / days.

**Additional context**
Attached logcat with GPU crashes.
[logcat_gpu_crashes.txt](https://github.com/sonyxperiadev/bug_tracker/files/4339763/logcat_gpu_crashes.txt)

MarijnS95:
Might fix https://github.com/sonyxperiadev/bug_tracker/issues/559 and https://github.com/sonyxperiadev/bug_tracker/issues/510

Parent clocks on gfx3d_src were not enabled, which is hopefully the reason for failed RCG configuration all of a sudden. The hardware might reject switching to a parent clock that's not actually turned on.

This furthermore aligns with CAF 8.2.1 clocks by adding more debug information to sysfs/WARN_CLK through `.list_registers()`, as well as providing the option to disable the src clock.

Tested on Mermaid to not show any immediate breakage. Aforementioned issues only occur longer-term, which I cannot consistently reproduce.

### DIFF
--- a/drivers/clk/qcom/clk-rcg2.c
+++ b/drivers/clk/qcom/clk-rcg2.c
@@ -1543,6 +1543,8 @@ const struct clk_ops clk_gfx3d_src_ops = {
 	.set_rate = clk_gfx3d_set_rate,
 	.set_rate_and_parent = clk_gfx3d_src_set_rate_and_parent,
 	.determine_rate = clk_gfx3d_src_determine_rate,
+	.list_rate = clk_rcg2_list_rate,
+	.list_registers = clk_rcg2_list_registers,
 };
 EXPORT_SYMBOL_GPL(clk_gfx3d_src_ops);
 
--- a/drivers/clk/qcom/clk-rcg2.c
+++ b/drivers/clk/qcom/clk-rcg2.c
@@ -1536,6 +1536,8 @@ static int clk_gfx3d_src_set_rate_and_parent(struct clk_hw *hw,
 }
 
 const struct clk_ops clk_gfx3d_src_ops = {
+	.enable = clk_rcg2_enable,
+	.disable = clk_rcg2_disable,
 	.is_enabled = clk_rcg2_is_enabled,
 	.get_parent = clk_rcg2_get_parent,
 	.set_parent = clk_rcg2_set_parent,
--- a/drivers/clk/qcom/gpucc-sdm660.c
+++ b/drivers/clk/qcom/gpucc-sdm660.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2017, The Linux Foundation. All rights reserved.
+ * Copyright (c) 2016-2020, The Linux Foundation. All rights reserved.
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License version 2 and
@@ -144,7 +144,7 @@ static struct clk_init_data gpu_clks_init[] = {
 		.parent_names = gpucc_parent_names_1,
 		.num_parents = 4,
 		.ops = &clk_gfx3d_src_ops,
-		.flags = CLK_SET_RATE_PARENT,
+		.flags = CLK_SET_RATE_PARENT | CLK_OPS_PARENT_ENABLE,
 	},
 	[1] = {
 		.name = "gpucc_gfx3d_clk",
--- a/drivers/clk/qcom/gpucc-sdm660.c
+++ b/drivers/clk/qcom/gpucc-sdm660.c
@@ -210,7 +210,6 @@ static struct clk_rcg2 gfx3d_clk_src = {
 	.mnd_width = 0,
 	.hid_width = 5,
 	.freq_tbl = ftbl_gfx3d_clk_src,
-	.enable_safe_config = true,
 	.parent_map = gpucc_parent_map_1,
 	.flags = FORCE_ENABLE_RCG,
 	.clkr.hw.init = &gpu_clks_init[0],


Haxk20:
Mermaid seems stable with this so far. Will test more. Will report back when it crashes again or well it doesnt.

kholk:
Just to be paranoid, check gpu performance....

But anyway, this should be OK to merge.

sledges:
Hi, I've had couple of the same GPU-related rare random reboots with Xperia 10 Sailfish OS (Android 10 base) kernel 4.9 two weeks ago, but could not reproduce it since then, even during similar load case scenarios.

Your patch applies to 4.9 kernel, would it be safe to merge it there as well?

MarijnS95:
@sledges This should go right on top of what @krnlyng did for 4.9 in https://github.com/sonyxperiadev/kernel/pull/2046 (which was [picked for 4.14 too](https://github.com/sonyxperiadev/kernel/pull/2133/commits/4fecfaa803ba1d2676009aa359ea886fffbe87c8)). Just like you we found the same crashes but way less often and at completely random times, sometimes not reproducible for weeks. These patches are supposed to address it but @haxk20 reported the clocks failed to turn on after a while (rather than stuck while switching parents) which is a likely consequence of https://github.com/sonyxperiadev/kernel/pull/2217/commits/b43328204782291d3a865a39370ccc52fdf745ff. Provided logs did not show whether the GPU was actually in the sketched recovery scenario.

kholk:
The clock API is basically acting the same way in both 4.9 and 4.14.
Also, the clock drivers for sdm630/60 that you find in 4.14 are a direct port of the one that you see in 4.9.

This means that whatever change we apply here on 4.14 can be backported 99.9% safely to 4.9 and will have the same effect.

Also, be aware that, as Marijn said, we saw some issues with display clocks... Though, I'm mostly sure that these are not in any way related to the changes made in this patchset, but something that emerged later for other reasons (and I think I know why... And 4.9 should not be affected).

MarijnS95:
> Also, be aware that, as Marijn said, we saw some issues with display clocks... Though, I'm mostly sure that these are not in any way related to the changes made in this patchset, but something that emerged later for other reasons (and I think I know why... And 4.9 should not be affected).

I think you are referring to a different issue regarding `byte0_src`. After recompiling the kernel, including patch `.172` I cannot reproduce the issue: likely a fluke in my local tree.

`gfx3d_src` failing to turn on is totally related to this patchset as far as I'm aware :grin: (except that, as you said before, the clock consumer should not turn it off apart the very exceptional recovery scenario)

kholk:
Oh, well. Yeah.

Anyway, remember what I said about the kgsl commit that I've sent you blahblah and you couldn't find it?
That was directly related to gpu recovery :)))

By the way, if you guys can throw a log of the rcg enable failure I may be able to get some kind of idea to fix it.
In any case, remember: the RCG should always be set to a safe frequency and a safe parent in this kind of failure cases. This is due to how the GPU RTOS behaves.

jerpelea:
this patch should be safe to backport to 4.9

MarijnS95:
> Anyway, remember what I said about the kgsl commit that I've sent you blahblah and you couldn't find it?
> That was directly related to gpu recovery :)))

I still don't have it :joy: 

> By the way, if you guys can throw a log of the rcg enable failure I may be able to get some kind of idea to fix it.

@haxk20 again, you're up. See if you can get a dmesg _including_ the logs that come before the failure. I don't continuous stacktraces of it failing to enable are of any relevance:

[crashGPU.txt](https://github.com/sonyxperiadev/kernel/files/4374908/crashGPU.txt)

> In any case, remember: the RCG should always be set to a safe frequency and a safe parent in this kind of failure cases. This is due to how the GPU RTOS behaves.

`clk_rcg2_disable` doesn't seem to do so since `FORCE_ENABLE_RCG` is set for gfx3d_src. It'll only set `CMD_ROOT_EN` to 0 and return rather than parking, which requires `enable_safe_config` in addition.

jerpelea:
pushed on top of 4.9

Haxk20:
> > Anyway, remember what I said about the kgsl commit that I've sent you blahblah and you couldn't find it?
> > That was directly related to gpu recovery :)))
> 
> I still don't have it joy
> 
> > By the way, if you guys can throw a log of the rcg enable failure I may be able to get some kind of idea to fix it.
> 
> @Haxk20 again, you're up. See if you can get a dmesg _including_ the logs that come before the failure. I don't continuous stacktraces of it failing to enable are of any relevance:
> 
> [crashGPU.txt](https://github.com/sonyxperiadev/kernel/files/4374908/crashGPU.txt)
> 
> > In any case, remember: the RCG should always be set to a safe frequency and a safe parent in this kind of failure cases. This is due to how the GPU RTOS behaves.
> 
> `clk_rcg2_disable` doesn't seem to do so since `FORCE_ENABLE_RCG` is set for gfx3d_src. It'll only set `CMD_ROOT_EN` to 0 and return rather than parking, which requires `enable_safe_config` in addition.

I was just slightly late to grab the dmesg. Next time it crashes hopefully i can grab it ASAP.

sledges:
Thanks for the above, I hope this helps:
[dmesg_rcg_warnings_lead_to_reboot.txt](https://github.com/sonyxperiadev/kernel/files/4377892/dmesg_rcg_warnings_lead_to_reboot.txt)

MarijnS95:
@sledges what logs am I looking at? This looks like before the patches, unless you managed to reproduce the issue with these, too.

In case I did not make myself clear I'm looking for logs similar to those from @Haxk20 where the clock fails _to turn on_ after some time, which must be a direct consequence of enable/disable support being added by this PR. As @kholk said this RCG should be parked in a safe state when turned off, but that does not happen due to `FORCE_ENABLE_RCG` and a lack of `enable_safe_config`. I want to see if/when this happens and fix the clock first, before submitting more patches that might prevent a GPU recovery scenario in the first place (so that in the much less likely event that it _does_ need to recover, it doesn't die unconditionally).

sledges:
Sorry, my logs were indeed still without the patches.

MarijnS95:
@sledges Any updates? Did you manage to reproduce the failing clock enable and get a useful trace out of it?

I'd like to see if the mentioned parking fixes the recovery scenario, before applying a patch that might make such a scenario appear less frequently or not at all.

sledges:
Nothing so far. Should I monitor dmesg for gfx3d clk warnings? Or might I get rare random reboots/crashes?

MarijnS95:
@sledges Afaik the device just keeps going without display, continuously failing to turn the GPU clock back on like in @Haxk20's logs.This unfortunately makes it harder to catch the case as logs are likely expired by the time it is noticed, device is attached to adb etc.

One might also replace the `WARN()` with a `BUG_ON` and extract pstore after it happened. No guarantees that other clocks aren't stuck temporarily and crash the device instead.

sledges:
Hi, we updated the kernel to 4.9.215 (https://github.com/mer-hybris/droid-hal-sony-ganges-pie/pull/40).

Then randomly (but rarely) the device would show white screen (and nothing else helps but rebooting it manually via long-press power button).

Here are the logs, hopefully of a long-enough timespan: 
[dmesg_white_screen.txt](https://github.com/sonyxperiadev/kernel/files/4707751/dmesg_white_screen.txt)

Haxk20:
> Hi, we updated the kernel to 4.9.215 ([mer-hybris/droid-hal-sony-ganges-pie#40](https://github.com/mer-hybris/droid-hal-sony-ganges-pie/pull/40)).
> 
> Then randomly (but rarely) the device would show white screen (and nothing else helps but rebooting it manually via long-press power button).
> 
> Here are the logs, hopefully of a long-enough timespan:
> [dmesg_white_screen.txt](https://github.com/sonyxperiadev/kernel/files/4707751/dmesg_white_screen.txt)

Yes this is an rare rare issue that happens sometimes.
We havent looked into it that much. Also its weird that its happening at 4.9 where this issue was i think fixed somehow. We only see this on 4.14.

sledges:
Once the phone was in direct sunlight, and was under quite a bit of load software-wise too (GPS navigation etc). `healthd` was reporting ~52 degrees C, and device went to white screen.
I rebooted, but that time it went into white screen again rather quickly. So this could potentially be one way to reproduce I'd guess.

MarijnS95:
@sledges apologies for the late response, this unfortunate to hear. I "dug" into CAF again and found another interesting difference: we are selecting the last `gcc_gpu_gpll0_clk` parent too which CAF doesn't do. Can you apply https://github.com/MarijnS95/kernel/commit/d42ec19d9b2 and see if that fixes the issue for longer?

Otherwise, allow me to cook up a little patch that prints the last requested parent/configuration, hopefully allowing us to find a pattern.

@Haxk20 Tagging you too for aiding in debugging this issue :)

sledges:
Hi @MarijnS95, thank you for coming back.

I managed to provoke white screen in the same way with your [patch](https://github.com/MarijnS95/kernel/commit/d42ec19d9b2) on top of the https://github.com/mer-hybris/android_kernel_sony_msm/pull/63. I can try to update to the very latest of your version of 4.9 in case there has been any interim work and retry replicating the problem again.

Here are the logs: [whiteout-2020-06-07-1835.log](https://github.com/sonyxperiadev/kernel/files/4744577/whiteout-2020-06-07-1835.log)

MarijnS95:
@sledges That's disappointing, this was our last option before diving into the parent clocks themselves... If it wasn't for finding a mistake on 4.14: https://github.com/kholk/kernel/commit/4e36d97adadcd3bad4c1de146b5107f7163b83ac. Unfortunately this is a fixup for the refactor in 23934d017688df86fc32b88fc7a6b48bf3f4e84c which didn't make it to 4.9. If it ends up solving the case for @Haxk20 on 4.14 anyway we'll likely need to backport all the PLL changes to 4.9.

MarijnS95:
@sledges I totally forgot I had a tracing patch lying around when I last looked into this. It was right under my nose in my tree the entire time. Back then I just couldn't reproduce it anymore, with days of uptime.

Can you see if this applies on 4.9, and whether all `update_config` cases are covered? (even if `clk_gfx3d_src_set_rate_and_parent`, not to be confused with `clk_gfx3d_set_rate_and_parent`, is the only case we realistically have to care about):

https://github.com/MarijnS95/kernel/commit/c96459e84c3ca456658d618d7a06374586e4a3b0

sledges:
Thank you I'll check. So far I could not reproduce the white-out on 4.9.221 with your earlier patch atop (the one it got reproduced was on 4.9.215 + your patch).

MarijnS95:
> So far I could not reproduce the white-out on 4.9.221 with your earlier patch atop

That's the nature of this issue, it may take weeks to repro and impact from external factors (such as temperature) have yet to be proven. I doubt anything changed inbetween .215 and .221, especially given the fact that this issue has always remained even on 4.14.

https://github.com/sonyxperiadev/kernel/compare/80876c4bf29cce452a7a4503b3661adf86f31a10...aosp/LE.UM.2.3.2.r1.4

The only changes to RCG2 that I can find (GH UI is massively restrictive here) are the patches discussed and applied here :). PLL changes are irrelevant as we use `dsi-staging` which - AFAIK - doesn't refer back to `dsi`.

sledges:
Tested many times, and it looks like we might be facing the "debug printf()s solved the problem" conundrum.
I've backported your patch to 4.9 as https://github.com/sledges/android_kernel_sony_msm/commit/b99b66e7f4ceeb6ade737207fef78d4f84d2cec0, and I could never reproduce the white out.
Potential "solution" being reading clocks' old config... Which is present in 4.14 thanks to https://github.com/sonyxperiadev/kernel/commit/8dc79c03ade3a072fd2fcacb3f1039db09065352

kholk:
So basically you are suggesting that poking the hardware before actually doing stuff is *the* solution. Funny.

But.. It makes sense, really! I've had to do something like this in other contexts, like the SPMI on SDM845, where I was getting regulators stuck on because of a hardware bug that got solved basically exactly like that.

This is the kind of stuff that makes me cry....and laugh. Hysterically.
Good job tho! Looks like this nightmare has finally come to an end!

MarijnS95:
It must be the readback, as the debug prints are _only_ happening _after_ an update failure. Unless the failure still happens, but the print gives it long enough to recover (as the failure is simply "the bit didn't change within 500us").

kholk:
For sure it's not a timing issue. I mean, we aren't facing a situation in which the bit changes after more than 500us.
I know for sure because, well, I tried to give it plenty of time... And it wasn't the case. 

In this moment I really believe that the fix is seriously about poking it before setting....

sledges:
I'll do two more tests, one with readback, one without, and no printfs altogether. Just to eliminate any doubt including my build env :)

PR #2223, opened 2020-03-25
[URGENT] [7.1] sched: use rq_clock if WALT is not enabled

MarijnS95:
Use rq_clock to send timestamp to util update handler
if WALT is not enabled in the system.


kholk:
Fixes cpufreq transitions especially on RT and DL tasks.

This is URGENT.

PR #2225, opened 2020-03-26
[7.1.r1] Fix QUSB reg offsets for tama

artem:
k4.9 and k4.14 have different definition of "enum qusb_phy_reg",
leading to incorrect registers being used. By an unlucky coincidence,
USB2_PHY_REG_MAX values magically match in both versions, so driver did not
error out.


artem:
k4.9 has some weird quirks for sdm845 that use QUSB2PHY_SQ_CTRL{1,2}, idk if it is worth porting them over

kholk:
What the hell is going on here!!
And why did usb work without using SQ_CTRL?!?

Is it hard-coded now?

Good catch, by the way!

PR #2228, opened 2020-04-12
[7.1.r1] Update hw_breakpoints, bvec optimizations, GCC9.2 ready

kholk:
This patchset updates hw_breakpoints in order to solve some issues with GCC9.2.

Also push an optimization for bvec in order to get better asm generation (and slightly better performance) with some compilers, including older gcc and clang releases. 

Tested on SoMC Tama Akatsuki RoW, GCC 9.2.


kholk:
Stable. MERGE!

PR #2229, opened 2020-04-13
[2.3.2.r1.4] Linux 4.9.219

stefanhh0:
Following files required manual merge conflict resolutions:
- `drivers/char/random.c` -> Manually resolved the merge conflict. Applied "random: always use batched entropy for get_random_u{32,64}" accordingly.
- `drivers/clk/qcom/clk-rcg2.c` -> Manually resolved trivial merge conflict. Return `-EBUSY` instead of `0` in `update_config` when it failed.
- `drivers/gpu/drm/msm/msm_gem.c` -> Manually resolved the merge conflicts plus restored original upstream comment in `#get_pages`.
In `#put_pages` an if block was missing completely plus `dma_sync_sg_for_cpu` was not done at all in the previous version. Added the missing if-block accordingly.
 For details see:
 drm/msm: Use the correct dma_sync calls in msm_gem
 commit dca98889e8e548715668c7aadfb4a47d480f16f0
- `drivers/usb/dwc3/gadget.c` -> Manually resolved the trivial merge conflict: usb: dwc3: don't set gadget->is_otg flag
- `include/uapi/linux/coresight-stm.h` -> Not applying commit: "coresight: do not use the BIT() macro in the UAPI header" since currently macros aren't used at all in that file.

Please review. 

The kernel already runs fine on yoshino/lilac.


stefanhh0:
@nathanchance I've compared my changes with:
https://github.com/android-linux-stable/msm-4.9/commit/1e215e9df8312c7fe146a37762ab63672e33e1d8

There is a difference in: `drivers/gpu/drm/msm/msm_gem.c` to me it looks like you have skipped resolving the merge conflict by accident, could that be the case? Or was it done that way intentionally and if so why?

nathanchance:
I do not have a 4.9 device in an available state to test the merges that I do so I am overly cautious resolving conflicts that could directly affect Android devices. This was an intentional discard, it looks like Qualcomm has part of the series in their tree already and I trust that there are no crashes like the ones described in the commit messages in linux-stable.

If you have verified your change causes no issues, I would not worry about the differences between ALS.

stefanhh0:
Thanks for your explanation. Until now the kernel is running fine, at least on my device dmesg output looks as well clean. We will see the next days if that causes issues or is overall stable.

chris42:
Tested this one on Kugo and could not spot anything unusual yet.

jerpelea:
please rebase

PR #2231, opened 2020-04-16
[7.1.r1] [CRITICAL] LSE Atomics support, GIC fix for SDM630/660

kholk:
This patchset brings support for ARM64 LSE atomics (v8.1 instruction set), while
keeping 100% compatibility with LL/SC: the feature will not affect any functionality
unless it is enabled in the defconfig.

Also, this brings a fix for an interrupt mess on SDM630/660, while keeping
compatibility with MSM8996(/98) class SoCs, fixing the unability to wake up
the phone when receiving a call while in deep sleep, as reported by @haxk20
and implementing the solution found by @oshmoun.

This PR depends on #2228 and should be merged immediately as it brings
in a CRITICAL BUGFIX for Nile and Ganges platforms.


kholk:
CLICK IT :P

Paulbouchara:
MERGEME

PR #2235, opened 2020-04-21
[7.1.r1] leds: qti-tri-led: Write known value to TRILED HW at init time

kholk:
The bootloader may leave one (or more) of the LEDs up when booting
the HLOS: in that case, the LED will be stuck because we are
disallowing to write brightness when br_new==br_old.

We initialize the brightness to zero when initializing the driver,
so also initialize the hardware to known values: this way, the
driver init matches with the hardware status.

Tested on SoMC Kumano Griffin DSDS (@MarijnS95)


MarijnS95 (approved):
gets the stamp of approval from me, solves the stuck red led issue as expected :grin:

PR #2238, opened 2020-04-22
[7.1.r1] Import NFC PN553 driver for Kumano

MarijnS95:
This driver is needed to power the NFC chips found in Kumano devices. Also moves PN547 to retain a consistent folder structure. Build-tested for Akatsuki (which has PN547) with the driver being compiled in.

PN553 userspace support for Kumano devices hopefully follows shortly :tm:


kholk:
Finally we have it! Go go go! :D

PR #2241, opened 2020-04-30
[KERNEL/KUMANO] arm64: DT: SM8150: Flip mbhc detection switch

MarijnS95:
This is the only diff observed when comparing sm8150-audio-overlay.dtsi
to copyleft DT. Plugging in a wired headset results in a _disconnect_
event; unplugging it results in a _connect_ event, leading to the wrong
audio path in both cases.
Override the detection switch - just like stock - from 1 to 0 to achieve
the desired result.


kholk:
Go for it!

PR #2242, opened 2020-05-01
[2.3.2.r1.4] Linux 4.9.220

stefanhh0:
Following files required manual merge conflict resolutions:
- `drivers/of/base.c` -> Not applying https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?h=linux-4.9.y&id=8063482e1cbc547e496b1e5c62b9d5155c71f378 since in: https://github.com/sonyxperiadev/kernel/commit/fc00691916c11b7931e076c10236e8e4f21702ac the implementation of: `#__of_attach_node_sysfs` was moved to `drivers/of/kobj.c` and with the move the very same change was already done.
- `drivers/scsi/ufs/ufshcd.c` -> Not applying https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?h=linux-4.9.y&id=5ebfa3b9b37c926b4625b106ce854a277d2a6cb7 since in: https://github.com/sonyxperiadev/kernel/commit/9c8077087e81801753ce89322f47ce3d463ce0ad the return value of `ufshcd_sl_intr` is used, which might be important to do. The kernel.org commit does not use the return value of `ufshcd_sl_intr`, but sets simply `retval = IRQ_HANDLED;`

Please review.


chris42:
Skipped this in favour of .221
Let me know if an explicit test of this one is needed.

PR #2243, opened 2020-05-01
[KERNEL] [URGENT] Correct handling for new egistec sensors

kholk:
The drive strength for the ET580 VDD fixed-LDO was wrongly set
to LOW, even when the LDO was being pulled high: drive it
correctly to ensure low dropout noise, as this is going in a
chip that needs high precision and - beware - it is doing
capacitive readouts (so it's critical to have clean power here).

Also, push the correct powerup sequence (sleep times were
too short!) in the kernel driver for egistec sensors.

Tested on SoMC SM8150 Kumano Griffin DSDS


MarijnS95:
Great! We have conversed many times on this exact line, and it turns out it's was still wrong after all with the documentation and stock values not lining up at all :grin:

PR #2245, opened 2020-05-04
[7.1.r1] Implement Seamless Dynamic Mode Setting from continuous splash

kholk:
On some devices we want to change the display resolution during
continuous splash - before the userspace kicks in -.

Implement code path variations in order to seamlessly change
the resolution of the display and accordingly set the DSI PHY:
this way, we get a smooth transition from the continuous splash
logo to the first frame that's being sent from the userspace.

To achieve this, add a new property and, based on finding that,
temporarily disallow pushing frames, then force reprogramming
the DSI PHY registers with the new timings while in ON state
and carefully also reset the PPS settings (for the eventual
Display Stream Comnpression display).

Resume accepting any frame only after the setup is completely
over, so that we're sure that no image corruption happens in
the event of the userspace sending them before.
In the event of a very long init process, the frames will be
dropped on expiry, instead of semaphore waiting forever.

This behavior gets activated by setting the property
"somc,splash-dms-switch-to-this-timing" on the specific
timing that is meant to be target of this DMS during splash
mechanism.

Example:
disp-timing@0 {
	timings....
};
disp-timing@1 {
	somc,splash-dms-switch-to-this-timing;
	timings....
};

This method is now applied to the SoMC SM8150 Kumano Griffin
device, allowing it to work at its panel native resolution.

Tested on SoMC SM8150 Kumano Griffin DSDS.

### DIFF
--- a/arch/arm64/boot/dts/qcom/dsi-panel-somc-sec-3840p3288-cmd.dtsi
+++ b/arch/arm64/boot/dts/qcom/dsi-panel-somc-sec-3840p3288-cmd.dtsi
@@ -183,6 +183,7 @@
 				somc,mdss-dsi-flm2-off-command-state = "dsi_lp_mode";
 			};
 			timing@1 {
+				somc,splash-dms-switch-to-this-timing;
 				qcom,mdss-dsi-panel-width = <1644>;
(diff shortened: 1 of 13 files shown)


kholk:
What are we waiting for? :P

PR #2246, opened 2020-05-04
[7.1.r1] CAF Security

Haxk20:
Saw this was missing from [thanks to this](https://sx.ix5.org/info/post/sodp-2020-04-05-security-state/) so lets get it in the kernel.


ix5:
Now do the other patches as well! Qcom media and wlan are still open.

Haxk20:
> Now do the other patches as well! Qcom media and wlan are still open.

Will do later.

PR #2253, opened 2020-05-07
[7.1.r1] Cleanup compiler warnings

stefanhh0:
Several cleanups and a functional change because I think the macro redefinition is not correct.

Please review if the suggested changes are okay that way or if I should change some things.

Build-Tested and verified that with this changeset the warnings previously reported for
these files are no longer being reported.

### DIFF
--- a/drivers/soc/qcom/peripheral-loader.c
+++ b/drivers/soc/qcom/peripheral-loader.c
@@ -1655,7 +1655,9 @@ static int __init msm_pil_init(void)
 	struct device_node *np;
 	struct resource res;
 	int i;
+#ifndef TARGET_NO_MINIDUMP_SUPPORT
 	size_t size;
+#endif
 
 	np = of_find_compatible_node(NULL, NULL, "qcom,msm-imem-pil");
 	if (!np) {
--- a/drivers/gpu/drm/msm/dsi-staging/dsi_ctrl.c
+++ b/drivers/gpu/drm/msm/dsi-staging/dsi_ctrl.c
@@ -41,7 +41,6 @@
 
 #define CEIL(x, y)              (((x) + ((y)-1)) / (y))
 
-#define TICKS_IN_MICRO_SECOND    1000000
 
 struct dsi_ctrl_list_item {
 	struct dsi_ctrl *ctrl;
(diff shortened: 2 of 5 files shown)


drivers/soc/qcom/peripheral-loader.c
```diff
@@ -1678,8 +1677,10 @@ static int __init msm_pil_init(void)
 	for (i = 0; i < resource_size(&res)/sizeof(u32); i++)
 		writel_relaxed(0, pil_info_base + (i * sizeof(u32)));
 
-	/* Get Global minidump ToC*/
 #ifndef TARGET_NO_MINIDUMP_SUPPORT
+	/* Get Global minidump ToC*/
+	size_t size;
+
```
MarijnS95: ISO C90 disallows mixed declarations and code, undefining `TARGET_NO_MINIDUMP_SUPPORT` should throw a build error here. Iirc kernel guys commonly use:
```c
#else
	(void)size;
#endif
```
at the bottom, but don't quote me on that.

Another one might be to replicate
```c
#ifndef TARGET_NO_MINIDUMP_SUPPORT
	size_t size;
#endif
```
above, right after the other declarations.

UPDATE: That warning is not enabled as error, so that'll replace the current warning with:
```
kernel/sony/msm-4.14/kernel/drivers/soc/qcom/peripheral-loader.c:1683:2: warning: ISO C90 forbids mixed declarations and code [-Wdeclaration-after-statement]
  size_t size;
  ^~~~~~
```

stefanhh0: I will choose your second suggestion.

Part of my build-test was to check the build log for occurrences of "[-W". Don't know why it didn't show up.

MarijnS95: It doesn't show up because `TARGET_NO_MINIDUMP_SUPPORT` is defined by default for all SODP devices, at the top of the file ;)

I have manually undefined it to show you the effects, and remind myself that it's C90, not C99 :wink:

stefanhh0: Ah thought it would be then a warning from the preprocessor. Tbh I don't know much about C... But now I learned something :-)

stefanhh0:
Amended the commit message for:
power: supply: BATT_MISS_SOC macro must not be redefined

PR #2257, opened 2020-05-08
[7.1.r1] [!!TOTALLY-URGENT!!] Fix CPU scaling on legacy and early kernel panic

kholk:
This patchset fixes CPU DVFS on MSM and SDM630/660.
This is REALLY urgent, as CPU stays at bootloader configured rates right now and risks
overheating!!!!!!

Tested on SoMC MSM8998 Yoshino Maple RoW


oshmoun:
can we not remove the config and simply disable it in defconfig? sounds like a more sane solution :grimacing:

kholk:
The interactive governor uses slack timers and needs to be probed later in order to not generate a kernel panic: doing this means that you have to move the entire cpufreq framework to a late init, which is a big no. 

For this reason, it is sane to remove the support for interactive as default governor. 
The configurations will be cleaned later.

PR #2262, opened 2020-05-15
[LE.UM.2.3.2.r1.4] Remove msm8956-v1.1 1958 MHz node

LuizPauloDamaceno:
The msm8956 runs with 1.8 GHz in the BIG cluster, if you want to run with more clock you should overvolt it. This clock is cause of random reboots in AOSP roms. Sol, remove it if you want to fix it. Or, if you want to overclock, add 50 ~ 75 mV in the regulator node of the BIG cpu corresponding to this clock.


ix5:
That's a neat find, if it turns out to be true. Why didn't you choose to overvolt instead?

Regarding your commit message, it requires some formatting. Please update it with this message that I've edited for you:

```
arm64: dts: msm8956-v1.1: Remove 1958MHz node

The msm8956 chip runs at 1.8 GHz in the BIG cluster; running
with a higher clock rate requires overvolting.

This clock has been observed to cause "random" reboots when
set to 1958MHz.

In case overclocking is desired, add 50 ~ 75 mV in the
regulator node of the BIG CPU corresponding to this clock.

Signed-off-by: electronicslover <your-email>
```
(substitute `<your-email>` for your real email but keep the `<>` signs)

LuizPauloDamaceno:
Thank you for the orientations, is my first time here :) 

Yes, you can also overvolt it, i'll do some pushes soon for my repo of this source to suzu that is running with a stable OC of 2.3 GHz in big cores... For some reason i cant overclock LITTLE cores, but i think i shouldn't. 

I think overvolting to sustain these high frequencies (see in regulator .dtsi that 8956 seems to have 2.016 GHz clock frequency for big cluster. Can run normally with the proper overvolting. for 2.3GHz here i'm running with 1.1V core (normal is 0.99) so is a big change (+135 mV) and results in a bigger power consumption, also heating. I dont think that is the goal (of course we can have also better performance, like i've already seen here) of this community in AOSP kernel for example.


Thank you for the help in the commit message and i'm free and very excited to discuss hardware things if you want :) 

I've not seen any option to edit the commit message (i'm noob here), should i do a new pull request with the new format, right?

danielgusvt:
> I've not seen any option to edit the commit message (i'm noob here), should i do a new pull request with the new format, right?

Just do a `git commit --amend` and then `git push -f`

kholk:
This clock was added to uniform 8956 and 8976, as they are the same chip. Add 50mV to this step.

oshmoun:
How sure are you this is the sole cause of reboots on loire? I remember testing without this frequency, i.e setting maximum frequency of BIG cluster to a lower value, and device would still reboot while running 3dmark for example.

EDIT: I should state that it's been a while since I last tested, and that things might have changed between now and then. Still, could you please consider running 3dmark multiple times back-to-back with this PR to confirm it really does not cause a reboot anymore?

LuizPauloDamaceno:
As @kholk saw is better to add 1958 MHz anyway with 50 mV step? Considering this isn't better to also add 2016 MHz step instead 1958 MHz with 75 mV step, that i've already tested and its stable? As i say for now i'm running with 2300 MHz without random reboots but 125 mV more. This is causing a lot of overheating that i'm solving with a "fan cooled joystick" while gaming.

@oshmoun i can confirm to you that is the cause of random reboots (tested in AOSP r46). Removing this frequency node and / or adding higher core voltage to this step it is stable.

chris42:
@LuizPauloDamaceno how long are you running this configuration?
My Kugo/Loire will reboot only once in a while and I couldn't pin it to high load. I ran ~20 benchmark tests on a test device and got no crash even without any changes.
So are we talking about the same reboots here? Random sometime reboots vs. verifiable under load?

LuizPauloDamaceno:
@chris42 I'm running overclocked / overvolted for three or for days and also playing games with it, with a cooled joystick... Like these ones we buy on wish for example. Under high loads with 65 ~ 70 ºC per core cpu temperature. 

Yes, you're correct, verifiable under load is the better definition. I say random because sometimes android uses a lot of CPU and bumps clock to compress / decompress something or open some application in background for notifications.  But verifiable under load is better to describe it. 

My device is Suzu. I think there's no big hardware (cpu at least) difference between yours. Dont?

oshmoun:
ok thanks for the confirmation. Does this finally mean that loire is stable on 4.9? woohoo
@harryharryharry this might interest you ;)

chris42:
I'll put a test build on my daily driver. Then I can see over the next weeks, if I get reboots. During that we can decide, if/how overclocking should happen in the regular setup

LuizPauloDamaceno:
Okay! Test it. I'm testing different voltage configurations today and i illalso share the results here. Remember: the 2016 MHz step is also present in msm8956-regulator.dtsi. If you decide to enable it to test should be arround 75 ~ 85 mV higher than the stock. Good testing!

kholk:
No, don't push new frequencies.

chris42:
So far the device runs stable with the removed frequency. However normally it crashes right after I post something like this... So just putting it here and will look a few more days. ;-)

LuizPauloDamaceno:
Cool! Try also to add increase by 50 mV the voltage level corresponding to this higher clock (from 990 mV to 1040 mV) in the msm8956-regulator. I'm running here with 2.2 GHz in the BIG core without problems. With some over voltage, of course. https://github.com/sonyxperiadev/kernel/pull/2266/commits/52bc1ce83655f1ed7a84f63f467e6891f5c27b93#diff-39ccae056a70b43e76843acfa89f0795

mbc07:
> This clock was added to uniform 8956 and 8976, as they are the same chip. Add 50mV to this step.

@kholk may I ask why a 1958 MHz clock was added for these SoCs? All documentation I could find about both MSM8956 (Snapdragon 618/650) and MSM8976 (Snapdragon 620/652) indicates a maximum frequency of 1.8 GHz on the BIG cores, so anything above that is through overclocking.

Also, correct me if I'm wrong but I'm not aware of any SODP supported device that uses the MSM8976, it looks rather weird to have a slightly overclocked step for the MSM8956 (which is actually used on SODP supported devices) just to make both uniform. Is there any particular reason for that?

artem:
> Also, correct me if I'm wrong but I'm not aware of any SODP supported device that uses the MSM8976, it looks rather weird to have a slightly overclocked step for the MSM8956 (which is actually used on SODP supported devices) just to make both uniform. Is there any particular reason for that?

You are correct.
If I'm not mistaken, QCOM just decided to label MSM8976 units with low-quality silicons as MSM8956 (with two gold cores disabled).

jerpelea:
please do not overclock/overvolt !

ix5:
@LuizPauloDamaceno please change the title of this PR to `[LE.UM.2.3.2.r1.4] Remove msm8956-v1.1 1958 MHz node` so that we know it's for 4.9 when searching..

PR #2263, opened 2020-05-16
[7.1.r1] [URGENT] Revert "arch: dtsi: tama-common: panel: set min brightness to 1.2mA"

kholk:
This reverts commit 162f8d73d8cd36fffbc9af9581b538988d355168.

This commit disables continuous splash on SoMC Tama Akatsuki.
Reverting it brings cont splash back!

Tested on SoMC SDM845 Tama Akatsuki RoW.

PR #2264, opened 2020-05-17
 [7.1.r1] Fixes: Override panel current to 10mA for akatsuki

sjllls:
We could override the panel current on XZ3, but this commit needs to be tested. I'm  not sure if it works fine.
#2263


MarijnS95:
Tested on Akatsuki DSDS, this fixes the issue :slightly_smiling_face:

PR #2266, opened 2020-05-19
arm64: dts: msm8956-regulator: add overvoltage

LuizPauloDamaceno:
The msm8956 chip runs at 1.8 GHz in the BIG cluster. 
Running with a higher clock rate requires overvolting, 
when set to 1958MHz under high loads with the 990 mV level we have reboots.

In case overclocking is desired, add 50 mV in the regulator node 
of the BIG CPU corresponding to this clock. 50 mV will be added to
the two last voltage levels of BIG CPU. If 2 GHz want to be reached 
we can test stability with this first voltage, if not stable add more 
25 mV to this voltage level.

Signed-off-by: electronicslover <luizpaulodamaceno@live.com>


kholk:
It's not overclocking... And that's not overvolting, not in that sense anyway.

The aging algo is broken here, so that's why we need this 50mV raise :)))

LuizPauloDamaceno:
So, the 1.95 GHz is an default frequency for this soc? Why the old kernel has only 1.8 GHz?

PR #2269, opened 2020-05-24
arm64: dts: Squash common Seine code into one DTSI

konradybcio:
According to the filenames, this code is common for all
Seine devices. Let's not scatter it around different device
trees then.

Signed-off-by: Konrad Dybcio <konradybcio@gmail.com>


pablomh:
Hi @konradybcio, would you mind prepending the branch name in the title?

kholk:
No no no no. If you want to do something like this, please rebase your changes over my PR.

jerpelea:
please do not cleanup at this stage ! It will make out later work harder

PR #2277, opened 2020-06-02
[LA.UM.7.1] drivers: regulator: implement debug suspend

oshmoun:
The clock framework has support for debugging clocks during suspend.
Implement similar debugging for regulators.


kholk:
Can we please get a click here?

PR #2278, opened 2020-06-03
[LA.UM.7.1.r1] arm64: DT: Seine: Correct bootloader-panel-detect

ix5:
The driver looks for `somc,bootloader-panel-detect`. Correct this typo.


ix5:
Just for posterity, something really strange is going on with the commit history. This commit should not have been necessary.

PR #2279, opened 2020-06-04
[7.1.r1] Revert "arm64: DT: Yoshino: Globally disable continuous splash feature"

stefanhh0:
Since with this commit, yoshino lilac is unable to reactivate the panel
once it went off for more than ~5sec it has been reverted. As soon as
the root cause has been found and the underlying problem has been fixed
this revert should be reverted.

This reverts commit b71379c4717912dd945637291217aa438d5b71d2.

For further details, see as well https://github.com/sonyxperiadev/bug_tracker/issues/590


jerpelea:
@kholk you should not disable cont splash

kholk:
The continuous splash was disabled for a good reason, I'm not crazy :p

This commit will return... :p :p

PR #2280, opened 2020-06-04
[7.1.r1] [URGENT] supply: qcom: smb-lib: Disallow HVDCP detection and authentication

kholk:
The HVDCP authentication requires a proprietary userspace
component: disable HVDCP detection and force it as a standard
5V charger to stop disconnect-reconnect issue and still
allow the user to use legacy QuickCharge chargers to at least
charge their device normally.

P.S.: All Tama devices *do* support USB Power Delivery float
chargers on Open Devices, so it is still possible to fast charge
the battery: just use a proper charger.

Tested on SoMC Tama Akatsuki RoW, Apollo DSDS.

### DIFF
--- a/drivers/power/supply/qcom/smb-lib.c
+++ b/drivers/power/supply/qcom/smb-lib.c
@@ -244,6 +244,24 @@ enum {
 	MAX_TYPES
 };
 
+/*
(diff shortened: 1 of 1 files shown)


MartinX3:
Tested on SoMC Tama Akari DSDS.

PR #2284, opened 2020-06-08
[7.1.r1] KGSL upgrade, Fabia PLL recalibration

kholk:
The Fabia PLL likes to lose its calibration values and giving unstable clock
output every now and then: let's force calibration everytime we reset the
PLL or in any case everytime it loses the calibration value, so that we
guarantee an expected and stable clock output.

Also, upgrade KGSL in order to fully support the new platforms.

Please wait for testing before merge.


kholk:
Pushed more commits to support HaloCore DSP and CS35L41 I2C/I2S amp,
various updates, Xperia 10 II initial defconfig.

MarijnS95:
So far running fine on Akatsuki DSDS after an afternoon and night of use. Both Kumanos seemed happy too, but were not checked overnight. No zombies or dead clocks 😀

PR #2289, opened 2020-06-17
[7.1.r1] Ganges secondary cameras fixes

Haxk20:
This PR includes few changes.
Lets change fast_plus to the values from stock. They are nearly the same just some improvements.
And change the location and sensor mode of the secondary camera dts node.

### DIFF
--- a/arch/arm64/boot/dts/qcom/sdm630-ganges-kirin-camera.dtsi
+++ b/arch/arm64/boot/dts/qcom/sdm630-ganges-kirin-camera.dtsi
@@ -183,8 +183,8 @@
 					  "CAM_VDIG",
 					  "CAM_VANA",
 					  "CAM_VIO";
-		qcom,sensor-position = <1>;
-		qcom,sensor-mode = <1>;
+		qcom,sensor-position = <0>;
+		qcom,sensor-mode = <0>;
 		qcom,cci-master = <1>;
 		status = "ok";
 		clocks = <&clock_mmss MCLK3_CLK_SRC>,
--- a/arch/arm64/boot/dts/qcom/sdm630-ganges-kirin-camera.dtsi
+++ b/arch/arm64/boot/dts/qcom/sdm630-ganges-kirin-camera.dtsi
@@ -252,8 +252,8 @@
 		qcom,hw-tsu-sta = <18>;
 		qcom,hw-thd-dat = <16>;
 		qcom,hw-thd-sta = <15>;
-		qcom,hw-tbuf = <19>;
-		qcom,hw-scl-stretch-en = <1>;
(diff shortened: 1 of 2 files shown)


ix5:
- `ARM64: dts: mermaid: Correct camera location and sensor mode`
- Line length
- Signoff!

Haxk20:
>     * `ARM64: dts: mermaid: Correct camera location and sensor mode`
> 
>     * Line length
> 
>     * Signoff!

Should be fine now ?

stefanhh0:
Signed-off-by with contents like:

Signed-off-by: Haxk20 <16738302+Haxk20@users.noreply.github.com>
Signed-off-by: ix5 <ix5@users.noreply.github.com>

is not what is wanted. At least the real name is required and probably also a valid contact address to be able to contact the person in charge, see as well:

https://stackoverflow.com/questions/1962094/what-is-the-sign-off-feature-in-git-for

If you don't want to disclose such information better do not sign-off at all. E.g. Angelo seems not to sign-off his commits, however Marijn is doing it correctly, so the group has no common understanding whether or not to sign-off and how, maybe you should clarify that :-)

stefanhh0:
See as well:

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/process/submitting-patches.rst?id=HEAD

Section:
11) Sign your work - the Developer's Certificate of Origin

Haxk20:
Angelo does not sign off commits at all only when pushing stuff to mainline. And yes I should get some real email address for this but I do not see the point of doing it here I will do it in future PRs rho

stefanhh0:
In case the commit will be mainlined, it has to be edited afterwards, which is not ideal, additionally even if it won't go to mainline then the sign-off should be at least removed, since with its current content it is wrong and adds no useful information and would be a bad role model for others.

So it would be nice from you if you would reconsider your decision and choosing instead one of the two options:
- fixing it
- removing it

kholk:
I sign-off my commits only when I push to upstream, or to projects which rules require that you sign them.

As for now, here, there is no such rule, so you are not obliged to add sign-off to your commits BUT, if you want to, then at least do it properly.

P.S.: I agree with Stefan about requiring the full name on your sign-off. For any other organizational matter, I want to remind you that this is not the right place, nor the right topic, to start a discussion.

Haxk20:
I will blame ix5 for this. He told me to signoff. When I did he said I did
it OK. oK no signoff from now on then until mainline.

kholk:
There is noone to blame.... :)))

Haxk20:
@kholk

Haxk20:
@jerpelea please do your thingy

PR #2294, opened 2020-06-25
[7.1.r1] Cleanups for SM8150 and IPA

kholk:
This patchset adds some cleanup for SM8150, some updates around
and removes the hard dependency on the PCI subsystem for IPA.

Tested on Sony SM8150 Kumano Griffin DSDS.


MarijnS95 (approved):
Works! This is a hard requirement to build Kumano ~on .184 (https://github.com/sonyxperiadev/kernel/pull/2287)~ with the new defconfig changes.

PR #2295, opened 2020-06-26
[KERNEL] [7.1.r1] wlan-qc: Update qcacld submodule hashes to 8.1

MarijnS95:
Get rid of these "unstaged changes" in our local tree, even if we don't
have a .gitmodules setup and having these hashes checked in being pretty
much useless.

See https://github.com/sonyxperiadev/local_manifests/pull/102


pablomh:
Is there any reason why we don't have .gitmodules set up?

MarijnS95:
@pablomh No idea really; looks like this was a merge oversight when these repos were added to the tree. Techpack for example has a [`.gitignore`](https://github.com/sonyxperiadev/kernel/blob/aosp/LA.UM.7.1.r1/techpack/.gitignore) to make sure we don't track any of these.

The repos are anyway to be synced with `repo sync`, but adding a `.gitmodules` file should make it easier to do true "out of tree" builds. If we do, we should include the two techpack repos in there as well. I can set it up if you want to try it out, it shouldn't conflict with `repo` as long as we keep updating these hashes whenever the submodules change.

Since kernel is @kholk s domain, opinions?

kholk:
We had git modules.... Then we removed it because of some mess happening with Android's repo tool going completely nuts.

If it works again, I would be pleased to reintroduce the git submodules for the wifi...

MarijnS95:
On second thought it might be related to how submodules are stored. To `repo` these projects (as listed in the manifest) are just another git repository, with stuff in `.git/` symlinked to the project and object store in `.repo/`. When a submodule however, these files are stored in `.git/modules/` in the top-level repo (kernel) instead, with `.git` in the submodules an arbitrary file containing `gitdir: ../../../../.git/modules/drivers/staging/wlan-qc/qcacld-3.0` for example.

We have to check if repo supports this properly one way or another before setting it up again.

stefanhh0:
This one can be closed since #2302 was merged back already.

PR #2302, opened 2020-06-30
[7.1.r1] Add support for audio suspend during activity, updates

kholk:
This patchset adds support to suspend the audio card while some PCM
is open: this is a way to save power during audio pauses.
Also, update the UART driver to address a small quirk on error conditions
so that it's able to resume operation faster and sneak in a commit update
on the wifi drivers, because git gets really sad during development if we don't.

Tested on SM8150 Kumano Griffin DSDS.


MarijnS95:
> and sneak in a commit update
> on the wifi drivers, because git gets really sad during development if we don't.

https://github.com/sonyxperiadev/kernel/pull/2295

Yeah, I have no idea where the clickety-click went :joy:

kholk:
Ah, well.... I... Didn't notice :D

PR #2303, opened 2020-07-01
[7.1.r1] Linux 4.14.187

stefanhh0:
No manual conflict resolutions were required also no fuzz merges occurred. 

Please review and test.


stefanhh0:
Technically spoken, this is not a draft but finished, thus removed the draft status. However we should at least have a couple more devices tested on the kernel update before merging back.

PR #2308, opened 2020-07-03
[7.1.r1] Ganges - Solve display FB issues, fix back aux cameras, unbalanced vreg ena 

kholk:
This patchset gives the following improvements on the Sony SDM630/636 Ganges platform:
- Ability to use digitizer when no firmware has been provided for upgrade
- Solved unbalanced disables for LCDB regulator when powering down
   the display for the first time after continuous splash
- Bad, ugly, but working and effective hack to solve issues with the Android
   recovery and offline charging UI (by crashing this UI on purpose with a bad
   framebuffer operation and letting it reinitialize the display).
- Back auxilliary camera (second camera) is now declaring the expected power
   GPIOs that userspace sends in the power sequence

Test:
* Recovery has display
* Boot faliure screen works
* Can use digitizer in recovery environment
* Back AUX camera works

Tested on SDM630 Kirin DSDS.

### DIFF
--- a/arch/arm64/boot/dts/qcom/sdm630-ganges-common-sde-overlay.dtsi
+++ b/arch/arm64/boot/dts/qcom/sdm630-ganges-common-sde-overlay.dtsi
@@ -101,3 +101,11 @@
 &mdss_dsi0 {
 	qcom,cont-splash-enabled;
 };
+
+&lcdb_ldo_vreg {
+	vreg-cont-splash-enabled;
+};
+
+&lcdb_ncp_vreg {
+	vreg-cont-splash-enabled;
+};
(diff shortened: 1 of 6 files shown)


Haxk20:
@kholk Bad and ugly here is understatement.

kholk:
You're welcome to find a better and proper in-kernel solution in the small time that I've had (25 minutes), or in whatever time you like.

Once you can get it fixed in a more beautiful way (anyone can - issue is finding time), then we can revert that single commit and push yours, when time will come.

In the meanwhile, the only solution that we have for a working Ganges platform is what I've pushed here, so this has to be merged asap in order to provide full functionality.

P.S.: Don't forget the schedule! We are rushing like crazy here!

kholk:
Added AUX camera solution for Ganges platform.

PR #2309, opened 2020-07-04
[7.1.r1] Fixup critical clocks for MSM8998 Yoshino

kholk:
This patchset solves issues with platform freezes happening especially when suspending (deep sleep) the device.

Tested on MSM8998 Lilac RoW by @stefanhh0


stefanhh0:
Great job Angelo! Thanks for providing that fix!

kholk:
Much love, much Yoshino :D

PR #2310, opened 2020-07-06
[7.1.r1] MSM8998 clock tree partial rewrite and stability fixes

kholk:
This patchset includes a partial rewrite of the clock parents and PLLs,
aiming to get the MSM8998 Yoshino platform towards the stable world.

Also, a SDM845 fix sneaked in because - while looking at 8998 - I've
found something that seemed to stink a lil.. and it did! :)))
EDIT: Oh! There's also a msm-bus fix for SDM660's GNOC.. :P

Tested on MSM8998 Yoshino Maple RoW
@stefanhh0 partially tested on MSM8998 Yoshino Lilac
@MarijnS95 tested on SDM845 Tama Akatsuki RoW


MarijnS95:
Yup, and it didn't become phantom Akatsuki overnight :grin:

PR #2314, opened 2020-07-10
[7.1.r1] Linux 4.14.188

stefanhh0:
No merge conflicts, no fuzz merges. Please review and test.


pablomh:
Hasn't introduced any noticeable regression in my pioneer since it was submitted.

PR #2315, opened 2020-07-13
[DNM] [2.3.2.r1.4] Linux 4.9.222

chris42:
source kernel.org
https://cdn.kernel.org/pub/linux/kernel/v4.x/incr/patch-4.9.221-222.xz

Patch applied only with repositioning
Builds and device boots, audio, bluetooth/nfc, wifi, phone working

Tested on Kugo further test on other legacy devices would be appreciated.


stefanhh0:
Current version is 230, it would be great to have all patches up to 230 available, then it could be tested in one go. Each version should be in its own commit, either in this PR or split up into individual PRs.

Look out for fuzz merges, those need to be double-checked explicitly when applying patches.

chris42:
Already wanted to write you, to check these and the following PRs in detail, as I am new to this. 
I intend to go up to .230, however testing will take some time. Will test .224 for a longer run, as I had merge conflicts to solve.

Also @nathanchance could have a look?

Haxk20:
As @stefanhh0 said. It would be great to have them all in one PR just each version in 1 commit. That way its not making mess in history.

stefanhh0:
@Haxk20 even having 8 PRs must not end up in having 8 merge commits. It depends on how it is being done and how it is merged back.

The basic options to avoid merge commits are:
1. Let each PR branch depend on the predecessor PR branch and then only merge back the very last PR branch, that creates a single merge commit and implicitly closes all predecessor PR branches in github (with the prerequisite that the predecessors have no merge conflict though, however such a conflict can be resolved even after the merge is done).

2. Instead of creating merge commits at all, merge and rebase strategy can be chosen instead, which avoids merge commits completely. In fact for some repositories we are using that strategy. I can't say why we don't use that strategy for the kernel. There is a slight drawback in that approach, the commits are not weaved in by creation-timestamp and thus when looking at the history the commits are not ordered in a monotonically increasing way creation-timestamp-wise. I personally don't think that it is a problem at all. It is not unusual to use merge and rebase strategy for short-lived development branches.

When supporting several long-lived version branches in parallel, and bugs are fixed on the oldest supported and affected branch,  then the merge commit strategy is a very convenient and recommendable approach to propagate the changes from the affected branch to the most recent one (e.g. master). This approach can be found often in commercial contexts, but for sure creates a lot of merge commits. The number of merge commits depends then on the number of release branches that have been merged in order to propagate a bugfix.

Haxk20:
I'm aware @stefanhh0 
The main concern wasn't mainly git history but GitHub UI and having comments for the PRs in different ones. Either way it's now in 1.

stefanhh0:
> That way its not making mess in history.

vs.
> The main concern wasn't mainly git history but GitHub UI and having comments for the PRs in different ones

The sentences are contradictory. Anyway having a couple of PRs open as e.g. 8 different PRs for 8 changes building up on each other on github wouldn't be a problem at all for the Github UI and wouldn't be a mess. There are pros and cons for it, as always. I just wanted to elaborate a little on the git mechanics.

MarijnS95:
>  (with the prerequisite that the predecessors have no merge conflict though, however such a conflict can be resolved even after the merge is done)

If predecessors have a merge conflict that should not have been resolved in one of the followup PRs, resulting in no single commit inherently having any conflict with the base branch. Needless to say that creating PRs on top of the tracking branch rather than branching off an older HEAD is always the better option, and 4.9 doesn't see much if any development anyway.

The pros of merge commits are to have clear branching overview: When did the commits separate from the target branch, where and when did they join again, and what's the attached PR? (Rebased) cherry-picks don't exhibit any such information, requireing careful search in GitHub PR history instead.

Timestamp weaving should be a non issue anyway as the contributors to this project aren't shy of picking external commits or reordering their own for clarity, inherently changing _author_ timestamps. Only GitHub is **extremely stupid** to sort commits in PRs by time rather than parent-pointer relations 🤷

chris42:
Nice that I could spark such lively discussion. :-D How about some testing/checking of the .230 PR ;-)

stefanhh0:
@chris42 I'd test it if I had a device with that I could do it. Unfortunately I don't have. Maybe Felix has some time to test that PR on his device, I think he has a Kagura/Tone.

> If predecessors have a merge conflict that should not have been resolved in one of the followup PRs, resulting in no single commit inherently having any conflict with the base branch.

I agree, conflicts have to be resolved in the first place where the conflict occurs. What happened in my case was that I didn't force-push a local predecessor branch. Merging back the top most PR, resulted in github refusing to close all the older PRs with merge conflicts. After force pushing my local branches to github the problem was resolved and the remaining PRs have been then closed automatically by github.

> The pros of merge commits are to have clear branching overview

That is true, I didn't thought it through deep enough. Especially how the kernel is developed in the context of SODP. As a matter of fact the PRs contain important information that can't be found in the commit message, additionally it is the only way currently to summarize commits having the same topic or belonging together in a story/epic like context. That also means the developer branches have to remain in the git repository.

Another way to do it and as it is being done in other contexts, is using a change management tool. Those are capable modelling more complex workflows. Then the git commits contain a change ID and also the change management tool lists all the commits belonging to that change. Then it is no longer required to have the development branches in the git repository and in fact those can then be pruned from the repository and in consequence merge commits for integrating development branches make no longer sense. But sure we don't have such an environment, so it is right to continue with the merge commits on this repository.

> Only GitHub is **extremely stupid** to sort commits in PRs by time rather than parent-pointer relations

I don't think that is a bad decision at all. Both git log and github show the *history* and the most natural sorting criteria of a history is by definition a timestamp, don't you think so? In case of git log that is per default the commit timestamps, I guess for github it is the same?! The command line tool `git log` has some more options and more flexibility compared to the github user interface. So what you want is sorting instead via `--topo-order`, right? I wonder if a github repo can be configured to show topo-order as well. Seems not to be possible, at least not for community accounts.

MarijnS95:
> What happened in my case was that I didn't force-push a local predecessor branch.

Meaning that the "same" commits, in the topmost PR, had a different hash, correct?

> Merging back the top most PR, resulted in github refusing to close all the older PRs with merge conflicts. After force pushing my local branches to github the problem was resolved and the remaining PRs have been then closed automatically by github.

If these were rebased on top of the target branch before force pushing then the net commits are zero, leading to a close. If you made them point to the right commit that is reachable from the target branch then indeed GH does said auto-close which is pretty cool :)

> As a matter of fact the PRs contain important information that can't be found in the commit message

We decided to leave the kernel as-is, in part with this reasoning in mind. Lots of external commits are picked that may not contain relevant information as to why this was picked. A PR provides this information, but the merger has to manually copypaste it to the field otherwise the only back-reference is the PR ID.
For the other repositories (in particular device repos) we are fully in control (there's nothing to pick) and stuff the commit description chock-full with in formation. In fact I find it hard to write anything besides a quick summary and expect the reader to read individual commits for _contextual_ information on distinct changes.

> Then it is no longer required to have the development branches in the git repository and in fact those can then be pruned from the repository and in consequence merge commits for integrating development branches make no longer sense.

Unfortunately those projects have automated tools that merge and pick changesets to various branches and tags, leading to completely unreadable history unless merges are filtered out.

> I don't think that is a bad decision at all. Both git log and github show the _history_ and the most natural sorting criteria of a history is by definition a timestamp, don't you think so? In case of git log that is per default the commit timestamps,

I don't have any special settings on top of `git log` and it displays history linearly according to parent-child relationships. Commits from parallel (potentially merged) tracks are generally weaved in according to time, but also preserve their respective relationships. `--topo-order`, according to the documentation, does not display commits out of order to preserve timestamps but `Show no parents before all of its children are shown, and avoid showing commits on multiple lines of history intermixed`.

GitHub _should_ show commits in the same order as git log (eg. preserving parent relationships) in the `Commits` view, but arguably so interleaves them in a sorted manner with messages and events in the `Conversation` view (where we're reading and writing these messages now). A more intricate sorting algorithm that preserves commit chains within this view is desired IMO.
[Checking up on a recent PR shows GH seems to get some ordering right: forcepushed commits now show up after the last review messages, with a `Marijn added x commits on May something`, well before the last messages. I recall individual commits within this block displayed out of order in the event that they were reordered or new commits inserted during a rebase.]

Important to note that I'm speaking about _author_ timestamps here, not _committer_ timestamps. As one can only point to a parent commit (hash) if the parent already exists, the _commit_ timestamp is inherently newer than the parent (unless intentionally overriding this timestamp). [Perhaps all the above you said is true and isn't due to relationships but _commit_ timestamps, I haven't validated that.]
GitHub seems to have opted to sort by author rather than committer timestamps, which may or may not make sense depending on the scenario. As always, there's no one-size-fits-all :shrug:

stefanhh0:
> Meaning that the "same" commits, in the topmost PR, had a different hash, correct?

Yes, exactly

> Unfortunately those projects have automated tools that merge and pick changesets to various branches and tags, leading to completely unreadable history unless merges are filtered out.

I meant doing really no merges but always rebase after merge back the development branch into the target branch, so that all merge commits are gone, we are doing it that way in the company I currently work for and it works out nicely giving us a clean linear history and all the development branches are deleted after merge back/rebase to the target branch. If you want to add additional context or so we have in the commit description just the issue number of the issue tracker, however we have no source code integration from our issue tracker currently, which is not so super cool. Anyway, I wouldn't like to have tools that automate merges and doing a bad job like you have seen for some projects.

> I don't have any special settings on top of git log and it displays history linearly according to parent-child relationships.

Yes you are right, I was wrong :-) It shows by default parent-child relationships. I read too much documentation and made a wrong assumption.

>  GitHub should show commits in the same order as git log

Yes, I agree, I wasn't aware of that they show it sometimes differently. So kernel history is currently differently shown on github than in git log? Or are you referring only what you experienced in the past with PRs in conjunction with rebases?

> Commits from parallel (potentially merged) tracks are generally weaved in according to time

Right, and the time that is used is the committer timestamp. With `--topo-order` the ordering is changed to retain the parent-child ordering also for merges if I understood correctly, However I don't understand the documentation fully:

It says:

--- 

For example, in a commit history like this:

```
    ---1----2----4----7
	\	       \
	 3----5----6----8---
```

where the numbers denote the order of commit timestamps, git rev-list and friends with --date-order show the commits in the timestamp order: 8 7 6 5 4 3 2 1.

With `--topo-order`, they would show 8 6 5 3 7 4 2 1 (or 8 7 4 2 6 5 3 1); some older commits are shown before newer ones in order to avoid showing the commits from two parallel development track mixed together.

---

I don't understand why or when (under which circumstances) it should show the alternative ordering mentioned in braces: `8 7 4 2 6 5 3 1` Or is it the explanation after the semi-colon? If so then I would need to understand that sentence as well, maybe a little bit of context is missing for me to understand it fully.

MarijnS95:
> I meant doing really no merges but always rebase after merge back the development branch into the target branch

Rebasing feature branches on top of development branches, and development branches on top of the main branch is a very common strategy that you'll see a lot, even around here. We use it to keep (long living) PRs up to date and prevent merge conflicts. This is fine when running a single release branch like most projects, but when keeping multiple releases (ie. android versions) up to date  while commits are cherry-picked all around creates an inconsistent mess that's hard to read in log. Likewise for all the merges paired with it.
Having CI is definitely a requirement, especially when developing cross-platform. It just doesn't work for our Android case where everything requires the full fat environment, it's hard to implement that efficiently for every PR - not to mention the lack of tests (that are impossible to validate regardless as a lot is device/hardware specific).

> Yes you are right, I was wrong :-) It shows by default parent-child relationships. I read too much documentation and made a wrong assumption.

I'm not too sure yet, perhaps it's sorted by _commit_ timestamp, though creating commits with an older timestamp pointing to a newer one deliberately... sounds like a possible case when messing around with computer time but may simply be classified as undefined behaviour.

> Yes, I agree, I wasn't aware of that they show it sometimes differently. So kernel history is currently differently shown on github than in git log? Or are you referring only what you experienced in the past with PRs in conjunction with rebases?

GitHub used to sort commits according to _author timestamp_ when shown in the _Conversation_ view, but as mentioned it didn't appear on recent PRs anymore. Perhaps this was rectified.
(fwiw, author timestamps get out of order when reordering/inserting commits using a rebase, yes)



> With `--topo-order` the ordering is changed to retain the parent-child ordering also for merges if I understood correctly

parent-child ordering is always enforced, but `--topo-order` determines whether commits from parallel branches are allowed to be interleaved. It is perhaps more clear to look at git log with `--graph` which shows parallel tracks so that you can actually tell commits from two distinct ones apart.

> I don't understand why or when (under which circumstances) it should show the alternative ordering mentioned in braces: `8 7 4 2 6 5 3 1` Or is it the explanation after the semi-colon?

The part after the semicolon explains in generic terms why some commits, not the entire track, are displayed out of order date-wise - doesn't have anything to do do with the specific ordering between parentheses.
I assume in this case git shows the entire left or right track first based on another flag: either show the track with the most recent commit (eg. `7 4 2`) first, or show the left (`^1`) track first, before the right (`^2`) one.
There's probably not much use visual-wise, but might be useful in a specific scripting scenario when using plumbing commands like the mentioned `rev-list`.

stefanhh0:
Maybe worth to play a little around with commits and the different command line options to figure out how it behaves in praxis. Thanks for the good discussion, it helps me to understand such things much better :-)

PR #2319, opened 2020-07-14
[1/3] [KERNEL/Seine] Enable and configure Egistec driver

MarijnS95:
Update the DT configuration and defconfig to build and use our et51x driver. This is pretty much identical to Kumano except different GPIOs.

Tested on XQ-AU52.


MarijnS95:
@kholk When are we planning to move the defconfig to https://github.com/sonyxperiadev/kernel-defconfig? I don't think defining `CONFIG_FPR_FPC` was an accident, more like importing some default :wink:

PR #2325, opened 2020-07-27
[7.1.r1] media: msm: camera_v2: Do not drop reconfigure requests 630/660

### LINKED ISSUE #574: [Pioneer][Nile][10] Camera is broken with new 8.11r1 hal.
**Platform:**  Nile
**Device:**  Pioneer
**Kernel version:** 4.14 (Prebuilt)
**Android version:** 10
**Software binaries version:** SW_binaries_for_Xperia_Android_10.0.7.1_r1_v6a_nile.img

**Previously working on**
Has this always been a problem, or is it a new issue?
New issue, Clean and new build on 2020.5.5 with latest sources.

**Description**
Camera app reported " Can't connect to the camera"

**How to reproduce**
Build with 8.11r1 hal.

**Log**
```
01-31 11:39:16.193     0     0 W qcom,camera-flash soc: qcom,camera-flash@0: Led led_flash0 renamed to led_flash0_1 due to name collision
01-31 11:39:16.196     0     0 E CAM-SOC msm_camera_get_reg_base: 905 err: no valid mem region for device:camss_cpp
01-31 11:39:16.196     0     0 E MSM-CPP cpp_probe: 4640 optional: camss_cpp_base not present
01-31 11:39:16.200     0     0 I MSM-CPP cpp_init_hardware: 1135 CPP HW Version: 0x60010002
01-31 11:39:16.200     0     0 E MSM-CPP cpp_init_hardware: 1153 stream_cnt:0
01-31 11:39:16.201     0     0 I MSM-CPP cpp_release_hardware: 1216 cpp hw release done
01-31 11:39:16.202     0     0 E msm_vfe48_get_dual_sync_platform_data: dual-vfe-irq not supported !
01-31 11:39:16.203     0     0 E CAM-SOC msm_camera_get_reg_base: 894 err: mem resource msm-cam not found
```
`01-31 11:39:20.383     0     0 E init    : couldn't write 642 to /dev/cpuset/camera-daemon/tasks: No such file or directory`

**Additional context**
[pioneer-cam.txt](https://github.com/sonyxperiadev/bug_tracker/files/4578309/pioneer-cam.txt)
[dmesg-pioneer-cam.txt](https://github.com/sonyxperiadev/bug_tracker/files/4578310/dmesg-pioneer-cam.txt)

kholk:
Fixes https://github.com/sonyxperiadev/bug_tracker/issues/574

ISP reconfigure requests may come while the MCT is still
processing some frame: in this case Qualcomm tells us to
drop the request until the MCT processing is done.

That ain't working fine at all on (old only?) platforms
that are using a Dual-VFE configuration, as the ISP seems
to slow down a whole lot during error processing, so the
MCT will be *always* signaling that's processing frames,
never actually giving a chance to this driver to actually
reconfigure and/or recover from the bad hardware state.
For this reason, just never drop reconfigure requests,
process them and do the job in any case: there are no
measurable drawbacks to this and the user experience is
anyway even improved.

While another cleaner solution to this issue does indeed
exist, it is way more complicated than whatever gets done
here.. and it doesn't give *any* improvement compared to
this patch anyway, so let's keep it simple and short.

Solves camera issues on SDM630 Nile Pioneer, Nile Discovery.
Tested on SDM630 Nile Pioneer.

### DIFF
--- a/drivers/media/platform/msm/camera_v2/isp/msm_isp_axi_util.c
+++ b/drivers/media/platform/msm/camera_v2/isp/msm_isp_axi_util.c
@@ -3717,7 +3717,7 @@ static int msm_isp_request_frame(struct vfe_device *vfe_dev,
 	enum msm_vfe_input_src frame_src = 0;
 	int k;
 	uint32_t wm_mask = 0;
-	int vfe_idx;
+	int is_vfe48_660 = 0, vfe_idx;
 	uint32_t pingpong_bit = 0;
 
 	if (!vfe_dev || !stream_info) {
@@ -3746,6 +3746,11 @@ static int msm_isp_request_frame(struct vfe_device *vfe_dev,
 	pingpong_status = vfe_dev->hw_info->
 		vfe_ops.axi_ops.get_pingpong_status(vfe_dev);
 
+	/* Force no drop_reconfig on SDM630/660 for dual-vfe issue */
+	is_vfe48_660 = msm_vfe_is_vfe48_660(vfe_dev);
+	if (is_vfe48_660)
+		vfe_dev->isp_page->drop_reconfig = 0;
+
 	/* As MCT is still processing it, need to drop the additional requests*/
 	if (vfe_dev->isp_page->drop_reconfig) {
 		pr_err("%s: MCT has not yet delayed %d drop request %d\n",
@@ -3771,7 +3776,8 @@ static int msm_isp_request_frame(struct vfe_device *vfe_dev,
 			__func__, __LINE__, frame_id);
 		vfe_dev->isp_page->drop_reconfig = 1;
 		return 0;
-	} else if ((vfe_dev->axi_data.src_info[frame_src].active) &&
+	} else if (!is_vfe48_660 &&
+		   (vfe_dev->axi_data.src_info[frame_src].active) &&
 			((frame_id ==
 			vfe_dev->axi_data.src_info[frame_src].frame_id) ||
 			(frame_id == vfe_dev->irq_sof_id)) &&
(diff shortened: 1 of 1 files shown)


MarijnS95:
```
Fixes https://github.com/sonyxperiadev/bug_tracker/issues/574
```
Please :grin:

MarijnS95:
> Solves camera issues on SDM630 Nile Pioneer, Nile Discovery.

FWIW it was only broken on Pioneer... Discovery - for some reason - is fine :confused:

Haxk20:
Tested on Nile Pioneer.
Works just fine now.

sjllls:
> Fixes [sonyxperiadev/bug_tracker#574](https://github.com/sonyxperiadev/bug_tracker/issues/574)
> 
> ISP reconfigure requests may come while the MCT is still
> processing some frame: in this case Qualcomm tells us to
> drop the request until the MCT processing is done.
> 
> That ain't working fine at all on (old only?) platforms
> that are using a Dual-VFE configuration, as the ISP seems
> to slow down a whole lot during error processing, so the
> MCT will be _always_ signaling that's processing frames,
> never actually giving a chance to this driver to actually
> reconfigure and/or recover from the bad hardware state.
> For this reason, just never drop reconfigure requests,
> process them and do the job in any case: there are no
> measurable drawbacks to this and the user experience is
> anyway even improved.
> 
> While another cleaner solution to this issue does indeed
> exist, it is way more complicated than whatever gets done
> here.. and it doesn't give _any_ improvement compared to
> this patch anyway, so let's keep it simple and short.
> 
> Solves camera issues on SDM630 Nile Pioneer, Nile Discovery.
> Tested on SDM630 Nile Pioneer.

Thanks a lot! :D

PR #2331, opened 2020-08-01
[7.1.r1] Fixes for MSM8998 GSI and non-GSI targets

kholk:
DWC3 may access GSI register(s) on non-GSI targets (so, an unexistant iospace, "hopefully"), so let's stop this.... moreover, add some more logic to the MSM8998 firmware GSI workaround: it is appropriate to free the channel memory when we are in a strange situation in which the firmware states that the channel is not allocated... also, instead of just returning success.. "perhaps" it's also mentally sane to zero out the channel refcount, so that we actually (uh, hopefully, too) stop IPA3 failures during the deinitialization path, with yet another hope to stop a kernel panic during subsystem restart phase.

### DIFF
--- a/drivers/platform/msm/gsi/gsi.c
+++ b/drivers/platform/msm/gsi/gsi.c
@@ -3014,6 +3014,11 @@ int gsi_dealloc_channel(unsigned long chan_hdl)
 		    ctx->state == GSI_CHAN_STATE_NOT_ALLOCATED) {
 			GSIERR("MSM8998 workaround for gsi firmware: avoiding "
 			       "kernel panic for already not allocated CH.\n");
+			if (ctx->allocated) {
+				devm_kfree(gsi_ctx->dev, ctx->user_data);
+				ctx->allocated = false;
+			}
+			atomic_set(&ctx->evtr->chan_ref_cnt, 0);
 			return GSI_STATUS_SUCCESS;
 		}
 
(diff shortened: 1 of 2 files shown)


kholk:
@stefanhh0 That's for you ;)

stefanhh0:
Cool, I have that running now on my phone. Again I had a startup where the modem failed, but then it recovered, which was never the case before. So this fix is helpful and improves the situation. I additionally observed that the wlan was going away for a short amount of time and then coming back. That is expected?

Here the [kernel.log](https://github.com/sonyxperiadev/kernel/files/5014376/kernel.log).

kholk:
OH YEAH! Your phone has experienced that same kind of modem crash that used to tear down the entire system in pieces and reboot last time (see the GSI MSM8998 workaround messages? :D) so that kind of crash is now finally fixed!

Regarding the wlan getting restarted, may seem strange, but it's actually totally expected: on Qualcomm platforms, the modem is actually a core subsystem, so when that one gets restarted, the SSR procedure will reboot (almost, if not) all of the remote subsystems.

Big achievement! MERGE!

ShujathMohd:
@kholk , Thanks for the fix.

There is some weird issue going on with Yoshino devices about the mobile network and wifi(2.4 ghz).

WiFI(2.4 ghz)  keeps disconnecting if the mobile network is set to 4g, however if I downgrade it to 3g(Or set airplane mode) and reboot I am not seeing any disconnections and both wifi and mobile network works great.
 
Requires your expertise here please :)

PR #2333, opened 2020-08-11
[6.4.r1] Update to Linux 4.4.222

sjllls:
I met some errors about IPv6 router when upgrade to 4.4.223, so only updated to 4.4.222.
Tested on Kagura, **this PR needs to be verified on other devices**.
Don't give up tone and loire, we are still alive on antique kernel. XD
Screenshot_20200811-141135_Settings


Thaodan:
Testing that with nile-discovery it looks like everything is fine. Merge?

Thaodan:
Ping @jerpelea

jerpelea:
@Thaodan should be tested on all platforms

Thaodan:
> @Thaodan should be tested on all platforms
Ah sorry that is something I can't do.

jerpelea:
this is a legacy branch and we should not accept patches for it 
Thanks for your understanding

PR #2341, opened 2020-09-14
[7.1.r1] Welcome, Android 11!

kholk:
This patchset introduces the new changes from Google for Android 11, in particular, it's bringing support for BinderFS and Incremental FS, other than a number of fixes around.

Please note, Android 11 is introducing a new fscrypt v2 API, which couldn't be applied cleanly (nor in any "reliable" dirty way), so that part was ignored. In any case, that Android version should be still supporting the old v1 fscrypto, so there shouldn't be any issue in running it on the "old" API.

Perhaps we can try to upgrade the fscrypt later, but that will bring a lot of headaches, due to our support for older platforms in this kernel (otherwise, we could've just replaced the implementation completely and called it a day).

This patchset was only compile-tested with SDM845 Tama Akatsuki defconfig (plus, obviously, CONFIG_ANDROID_BINDERFS and CONFIG_INCREMENTAL_FS were activated in order to compile-test the new entries).
Please test it on-device.


MarijnS95:
> This patchset was only compile-tested with SDM845 Tama Akatsuki defconfig (plus, obviously, CONFIG_ANDROID_BINDERFS and CONFIG_INCREMENTAL_FS were activated in order to compile-test the new entries).
Please test it on-device.

I'll give this an actual run-test on 11 tonight :grin:. Thanks for picking it all!

(Fwiw crypto seems fine on Griffin)

PR #2347, opened 2020-10-31
[6.4.r1] Linux stable 4.4.y inludes fixes for Bleedingtooth CVE-2020-12351, CVE-2020-1235 

Thaodan:
Please test first

### DIFF
--- a/net/bluetooth/a2mp.c
+++ b/net/bluetooth/a2mp.c
@@ -233,6 +233,9 @@ static int a2mp_discover_rsp(struct amp_mgr *mgr, struct sk_buff *skb,
 			struct a2mp_info_req req;
 
 			found = true;
+
+			memset(&req, 0, sizeof(req));
+
 			req.id = cl->id;
 			a2mp_send(mgr, A2MP_GETINFO_REQ, __next_ident(mgr),
 				  sizeof(req), &req);
@@ -312,6 +315,8 @@ static int a2mp_getinfo_req(struct amp_mgr *mgr, struct sk_buff *skb,
 	if (!hdev || hdev->dev_type != HCI_AMP) {
 		struct a2mp_info_rsp rsp;
 
+		memset(&rsp, 0, sizeof(rsp));
+
 		rsp.id = req->id;
 		rsp.status = A2MP_STATUS_INVALID_CTRL_ID;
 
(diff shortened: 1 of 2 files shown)


stefanhh0:
Hey @Thaodan 

thanks for providing the security fixes for all the sodp kernel versions. Normally we don't merge kernel.org directly into the sodp kernel but do it like this: https://github.com/sonyxperiadev/kernel/pull/2333. It would be really good if you could do it like that, ideally this PR should depend on https://github.com/sonyxperiadev/kernel/pull/2333 since @sjllls has already patched it up to  v4.4.222.

Best regards
Stefan

Thaodan:
Can you explain it a bit? It sounds more logical to keep the list of changes. I will rebase it on top of the changes request on monday.

stefanhh0:
I don't know the original thoughts why it was decided to do it this way. Probably to have a simpler history. It simplifies recognizing what was changed in the sodp kernel only, what are cherry-picks from various sources (like e.g. CAF or newer kernel.org branches) and which changes are simply patches from kernel.org. It helps the people more closely involved in this project to "understand" the history. In case there is a problem within a specific patch we always can expand the patch to the single commit list and then can do a git bisect to identify the commit that introduce the problem.

I don't know if that is a strict policy or just a best practice from that it is okay to deviate this time - I'd prefer to not deviate from that.

On a personal note, I don't think it is in general more logical to merge all commits. It is more a question of how the sources are organized. As always there are pros and cons for both approaches and none of the approaches is more logical than the other a priori. I have often seen that the "logical" argument was used to favour one approach over the other in avoidance to deliver pros and cons and coming then to an informed conclusion/decision.

jerpelea:
@stefanhh0 when there is a CVE we need to patch the issue as fast as possible and then we can merge the regular way.
In this case the patch must be cleaned and rebased 

@Thaodan Thanks for the fixes

MarijnS95:
@jerpelea Except that this CVE PR is littered with "regular" kernel upgrades/duplicates from #2333.

EDIT: Ah, you edited that in as well.

jerpelea:
@MarijnS95 this is why I asked to be cleaned

Thaodan:
> @jerpelea Except that this CVE PR is littered with "regular" kernel upgrades/duplicates from #2333.

Because I thought I could just pull the regular patches to while I was at it.

Now it just includes the fixes for the CVE. Tested on nile now way to trigger the CVEs anymore.

PR #2353, opened 2020-11-16
[DNM][7.1.r1] Add CLK_IGNORE_UNUSED to aggre1_ufs_axi_hw_ctl_clk flags

Haxk20:
There seems to be a race condition where when clk_disable_unused() is run it sees both the clocks.
This is happening because both gcc_aggre1_ufs_axi_clk and gcc_aggre1_ufs_axi_hw_ctl_clk use the same enable_reg and enable_mask.
Thus they are aliased.
This can be fixed by adding CLK_IGNORE_UNUSED to the hw_ctl_clk flags.

Huge thanks goes to Tootea who reported this bug and also made the idea for this fix.
See: https://github.com/sonyxperiadev/bug_tracker/issues/654

### DIFF
--- a/drivers/clk/qcom/gcc-msm8998.c
+++ b/drivers/clk/qcom/gcc-msm8998.c
@@ -1399,7 +1399,7 @@ static struct clk_branch gcc_aggre1_ufs_axi_hw_ctl_clk = {
 				"gcc_aggre1_ufs_axi_clk"
 			},
 			.num_parents = 1,
-			.flags = CLK_SET_RATE_PARENT,
+			.flags = CLK_SET_RATE_PARENT | CLK_IGNORE_UNUSED,
 			.ops = &clk_branch2_hw_ctl_ops,
 		},
 	},


MarijnS95:
Perhaps you should see why we have two clock structs shadowing the same register instead?!?!

Haxk20:
> Perhaps you should see why we have two clock structs shadowing the same register instead?!?!

I tried. But the same thing is there in 4.9 branch also. So im kinda confused on that.

MarijnS95:
> But the same thing is there in 4.9 branch also.

https://github.com/sonyxperiadev/kernel/blob/aosp/LE.UM.2.3.2.r1.4/drivers/clk/qcom/gcc-msm8998.c#L1450-L1485 Can you show me?

Haxk20:
> > But the same thing is there in 4.9 branch also.
> 
> https://github.com/sonyxperiadev/kernel/blob/aosp/LE.UM.2.3.2.r1.4/drivers/clk/qcom/gcc-msm8998.c#L1450-L1485 Can you show me?

I didnt mean that we have UNUSED in 4.9. I meant that the regs are the same in 4.9 as well.

Thaodan:
Please fix the title of the PR to include the kernel you want to merge to.

kholk:
This is *seriously* wrong, and the fix for this is easier than that.
Just remove the gcc_aggre1_ufs_axi_*hw_ctl*_clk and change all users to its parent: on k4.14 the HWCG is being accounted-for and I have even added the required configuration to the parent, hence this hack is not needed anymore.

Besides that, it feels strange to call gcc_aggre1_ufs_axi_clk a parent of the hw_ctl one: they're effectively the exact same clock.

Haxk20:
@kholk Thank you. Will do.

MarijnS95:
@kholk would you mind explaining what this "hack" was? Were the HWCG and halt registers previously coded in separate clocks? Now that registers appear to be duplicated here and in a ton other gcc- files, things way more wrong than ever imagined :D

Haxk20:
I will close this as its wrong. I will delete the occurances of that clock when i get the time to do so.

PR #2354, opened 2020-12-03
[LA.UM.7.1.r1] Import SDAM LPG support for Seine

MarijnS95:
Seine currently does not have any breath or pattern support on the LED; the `breath` file is missing.

This is easily fixed by importing the relevant SDAM support code from the downstream kernel (`59.0.B.2.328`) and reverting _stupid_ bits of code. Now `breath` shows up, and writing `1` to it gives the desired pattern.

Note that, despite `qcom,sync-channel-ids = <1 2 3>;` being set in `trinket-seine-common.dtsi` none of the SDAM codepaths use this (`lpg_group`). This makes it easily possible to create desynced patterns.

CC @simonmicro

### DIFF
--- a/drivers/pwm/pwm-qti-lpg.c
+++ b/drivers/pwm/pwm-qti-lpg.c
@@ -19,10 +19,12 @@
 #include <linux/kernel.h>
 #include <linux/module.h>
 #include <linux/mutex.h>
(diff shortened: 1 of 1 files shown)


simonmicro:
A little bit too late: Yes, it indeed works.

Or at least: Does not panic on usage.

PR #2356, opened 2020-12-14
[7.1.r1] drm/msm/sde: sde_hw_top: Fix UBWC version detection for SDM630

konradybcio:
SDM630 (and by extension 636/660 and probably 8998) don't
seem to have a working UBWC version register, but the UBWC
version is already specified in the DT and then read out
by SDE, so let's make use of that instead.

-------------------

Please test on tama/kumano!

`dmesg | sed 3000q | grep -i uwbc`

should return nothing. In case it returns

`Unsupported UBWC version 0xsomenumber`

please paste the output here so that I can fix it up.

### DIFF
--- a/drivers/gpu/drm/msm/sde/sde_hw_top.c
+++ b/drivers/gpu/drm/msm/sde/sde_hw_top.c
@@ -387,6 +387,10 @@ void sde_hw_reset_ubwc(struct sde_hw_mdp *mdp, struct sde_mdss_cfg *m)
 {
 	struct sde_hw_blk_reg_map c;
 	u32 ubwc_version;
+	u32 reg = m->mdp[0].ubwc_static |
+		(m->mdp[0].ubwc_swizzle & 0x1) |
+		((m->mdp[0].highest_bank_bit & 0x3) << 4) |
+		((m->macrotile_mode & 0x1) << 12);
 
 	if (!mdp || !m)
 		return;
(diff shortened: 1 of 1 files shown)


drivers/gpu/drm/msm/sde/sde_hw_top.c
```diff
@@ -387,31 +387,36 @@ void sde_hw_reset_ubwc(struct sde_hw_mdp *mdp, struct sde_mdss_cfg *m)
...
 
 	if (!mdp || !m)
 		return;
 
 	/* force blk offset to zero to access beginning of register region */
 	c = mdp->hw;
 	c.blk_off = 0x0;
-	ubwc_version = SDE_REG_READ(&c, UBWC_DEC_HW_VERSION);
+	/* 630 doesn't seem to have the UBWC version register */
+	//ubwc_version = SDE_REG_READ(&c, UBWC_DEC_HW_VERSION);
```
pablomh, line 402: Please, remove the line instead of commenting it.

konradybcio: I thought keeping it for reference is okay, newer platforms have this feature.

MarijnS95: I think this is WIP code. The original below reads both the register in `ubwc_version` and what was stored in DT as `m->ubwc_version`. Seems like it's done on purpose, I doubt this is correct. @konradybcio What do they do on mainline?

pablomh: FWIW, we can see in `drivers/gpu/drm/msm/sde/sde_hw_catalog.h`:

` * @ubwc_version       UBWC feature version (0x0 for not supported)`

konradybcio: @Marijns95 on mainline the version is hardcoded per platform.

PR #2369, opened 2021-01-04
[LA.UM.7.1.r1] 4.14.213

pablomh:
Everything applied cleanly except:

- https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable-rc.git/commit/drivers/scsi/ufs/ufshcd.c?h=linux-4.14.y&id=29ad0fa9733c27f8497d16fd4fa003f13ec751df
- https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable-rc.git/commit/drivers/soc/qcom/smp2p.c?h=linux-4.14.y&id=fc7f65e394728ac1d7491b82441b29131d1c73bc
- https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable-rc.git/commit/drivers/usb/gadget/function/f_fs.c?h=linux-4.14.y&id=27e1b18af55f85c57dddaf4c0d4d99d507a00b84

They needed to be applied manually (no manual editing apart from copying/pasting at the right offsets).


MarijnS95:
@pablomh Thanks!

Would it be possible next time to commit every conflicted commit separately before `treewide` (after all, you already have the original commit separate from the treewide diff) so that we can more easily compare the changes, and see that this commit was applied manually when browsing commit/file history later on?

pablomh:
I think I'll even start doing it for this one... Stay tuned!

drivers/scsi/ufs/ufshcd.c
```diff
@@ -1949,8 +1949,16 @@ static int ufshcd_devfreq_target(struct device *dev,
...
+	pm_runtime_get_noresume(hba->dev);
+	if (!pm_runtime_active(hba->dev)) {
+		pm_runtime_put_noidle(hba->dev);
+		ret = -EAGAIN;
+		goto out;
+	}
 	start = ktime_get();
 	ret = ufshcd_devfreq_scale(hba, scale_up);
+	pm_runtime_put(hba->dev);
+
```
pablomh, line 1961: I've added this space to match `-stable` code.

pablomh:
It boots and works without apparent regressions in both `pdx201 (seine platform)` (Android 11) and `pioneer (nile platform)` (Android 10).

PR #2373, opened 2021-01-19
arm64: configs: pdx206: Enable PINCTRL_SOMC

konradybcio:


MarijnS95:
We're not using https://github.com/sonyxperiadev/kernel-defconfig anymore I suppose?

jerpelea:
@konradybcio please use https://github.com/sonyxperiadev/kernel-defconfig

PR #2376, opened 2021-01-21
[LA.UM.9.12.r1] Needed fixes for SIM detect

pablomh:


MarijnS95:
I don't recall newer platforms with newer RIL like Kumano and Seine use `sim-detect`, perhaps Edo does not either? The `sim-detect` switch is _as far as I know_ only used for its namesake userspace application to show a "sim removed, reboot" dialog, but these devices all support hotswapping.

I could be totally wrong though.

konradybcio:
Well we either disable/drop it or import these fixes, the kernel doesn't compile as of current HEAD with the driver set to =y. Enabling it on edo might not be necessary though, as you said.

oshmoun:
I think it's still used on stock edo, regardless of hotswapping, to restart the device upon changing the telephony config.

MarijnS95:
Well, @konradybcio can _probably_ check if SIM hotwapping works without these patches - iirc it does, but my mind isn't even a sieve anymore - just a big gaping hole.

kholk:
I can definitely confirm that SIM hotswapping works fine without this driver on Xperia 1 II and 5 II.

pablomh:
Angelo, if it works without it, I'll drop the PR then :)

PR #2377, opened 2021-01-25
[LA.UM.7.1.r1] 4.14.217

pablomh:
This one applied cleanly.


pablomh:
It works without apparent regressions on both `pioneer` (`nile` platform) and `pdx201` (`seine` platform).

jerpelea:
cherry-picked

PR #2380, opened 2021-01-27
[7.1.r1] arm64: vdso32: Use full path to Clang instead of relying on PATH

ix5:
Using (cc-name) directly leads to selection of first (cc-name) in path without respecting (CC) - in practice, this means a fallback to host clang under /usr/bin/clang.

32-bit cross compilers will still be respected.

**EDIT: ** Apparently Nathan already fixed this and it was not ported over: [FROMLIST: arm64: vdso32: Use full path to Clang instead of relying on PATH](https://android.googlesource.com/kernel/msm/+/a01c2fe09643caff8c96b8153de5b1ef488fab20)

### DIFF
--- a/arch/arm64/kernel/vdso32/Makefile
+++ b/arch/arm64/kernel/vdso32/Makefile
@@ -5,7 +5,7 @@
 # A mix between the arm64 and arm vDSO Makefiles.
 
 ifeq ($(cc-name),clang)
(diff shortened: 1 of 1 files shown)


Haxk20 (approved):
I have nothing to nitpick or change here. LGTM

PR #2384, opened 2021-02-01
[LA.UM.7.1.r1] Raise ICNSS voltages on Nile and Ganges

konradybcio:
The values were waaay lower than what we had on stock.


pablomh:
@konradybcio, please, prepend the title with the branch name :)

PR #2392, opened 2021-02-08
[9.12.r1] Updates to PCIe, MHI for new WiFi and modem power savings

kholk:
This patchset enables compiling the newest WiFi stack and adds some
code that will also allow some modem power savings on the PCIe link.


kholk:
another github bug is hitting us, apparently, closing and reopening....

PR #2395, opened 2021-02-14
[9.12.r1] arm64: dts: edo: Finish removing Sony camera configuration

kholk:
Almost there... but one (now stale) phandle was left, breaking the
DTBOverlay for PDX203 completely.

Fixes: e5d0a237f6ce ("dts: edo: remove Sony camera configuration")


MarijnS95:
`camrea`?

PR #2397, opened 2021-02-15
[9.12.r1] Camera CMIC, cnss bugfixes, fingerprint sensor

kholk:
This patchset adds the camera CMIC (slg51000), fixes a bug in cnss2 when MEMORY_DUMP_V2 is disabled, adds support for the fingerprint sensor in OSS land.

### DIFF
--- a/arch/arm64/boot/dts/somc/kona-edo-common.dtsi
+++ b/arch/arm64/boot/dts/somc/kona-edo-common.dtsi
@@ -226,26 +226,32 @@
 		};
 	};
 
+	fpc_vana_vreg: fpc_vana_vreg {
+		compatible = "regulator-fixed";
+		regulator-name = "fpc_vana_fixed_reg";
+		regulator-min-microvolt = <3350000>;
+		regulator-max-microvolt = <3350000>;
+		startup-delay-us = <2000>;
+		enable-active-high;
+		gpio = <&pm8150l_gpios 8 0>;
+		vin-supply = <&pm8150a_bob>;
+	};
+
 	et6xx: et6xx {
-		status = "ok";
 		compatible = "egistec,et580";
 
-		egistec,gpio_irq = <&tlmm 23 0x0>;
+		et51x,gpio_irq = <&tlmm 23 0x0>;
+		et51x,no-low-voltage-probe;
+		et51x,vana-voltage = <3350000>;
+		vdd_ana-supply = <&fpc_vana_vreg>;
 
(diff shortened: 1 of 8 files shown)


MarijnS95 (changes requested):
No, we should remove CEI detection instead of falsely reporting Egistec may someone ever want to venture porting Nile and running into obscure issues like this.

arch/arm64/boot/dts/somc/kona-edo-common.dtsi
```diff
@@ -226,26 +226,32 @@
...
-		egistec,gpio_irq = <&tlmm 23 0x0>;
-
-		pinctrl-names = "et6xx_reset_reset",
-				"et6xx_reset_active",
-				"et6xx_ldo_low",
-				"et6xx_ldo_high",
-				"et6xx_irq_active";
+		et51x,gpio_irq = <&tlmm 23 0x0>;
+		et51x,no-low-voltage-probe;
+		et51x,vana-voltage = <3350000>;
```
MarijnS95, line 245: Regulator-fixed, is this just to prevent errors when the driver tries to set it to 3v3 flat? Gross.

kholk: If you have another cleaner alternative that doesn't break the whole driver logic and without hardcoding voltages again inside, you're welcome.

kholk:
> No, we should remove CEI detection instead of falsely reporting Egistec may someone ever want to venture porting Nile and running into obscure issues like this.

I don't like breaking drivers and losing functionality around.

MarijnS95:
> I don't like breaking drivers and losing functionality around.

So you opted to break the driver on FPC and make it loose functionality on FPC at the same time? +1 

:stuck_out_tongue:

kholk:
> > I don't like breaking drivers and losing functionality around.
> 
> So you opted to break the driver on FPC and make it loose functionality on FPC at the same time? +1

Last time I checked, the logic that I have wrapped up is used only in case `check_sensor_type` is enabled, which is *definitely not the case* unless you are running it on the Nile platform, which is anyway NOT SUPPORTED in kernel 4.19.

MarijnS95:
I am again saying that someone _might_ attempt porting Nile to 4.19. Why, I have no idea, but this would shoot them in the foot. Either remove CEI detection entirely or make it a compiler failure (`#error`).

jerpelea:
CEI detection will be handled in a future commit

PR #2400, opened 2021-02-17
[9.12.r1] Display optimizations and new timings

kholk:
This patchset sets up some optimizations for the Edo displays in order to reduce jitter and frame times for various usecases.
Detailed informations are present in the commits - so if you want to know more, just look at the commit descriptions.


kholk:
depends on https://github.com/sonyxperiadev/kernel-techpack-display-driver/pull/3

PR #2401, opened 2021-02-17
[kernel] [9.12] Remove nonsensical Android.mk for kernel drivers

MarijnS95:
We don't use any of the modules defined in the kernel (ie.  drivers/staging/greybus/tools), and they result in duplication conflicts with 4.14 - also checked out in the AOSP tree. Simply creating an empty Android.mk (that itself would be responsible for recursively including more makefiles) prevents the build system from looking beyond this path.


kholk:
No Marijn. I mean, I agree, but not this way.

*** RUDIMENTARY INSTRUMENTS WARNING ***
find .| grep "Android.mk" | xargs rm ; git commit -m "Remove nonsensical Android.mk for kernel drivers"

... Should be enough :P

MarijnS95:
Yeah, that's rudimentary...

```console
foo:kernel/ $ find -iname Android.mk -delete
```

There's only one makefile anyway, the greybus one that is causing trouble :grimacing:

PR #2405, opened 2021-02-25
[LA.UM.7.1.r1] 4.14.222

pablomh:
This one applied cleanly except for some bits, mainly due to the inclusion of this two Android commits not in "regular" -stable trees:

- 9abcee5156e4 (FROMLIST: kbuild: fix dynamic ftrace with clang LTO, 2017-06-16)
- af1f94d0afc1 (FROMLIST: kbuild: add support for clang LTO, 2017-11-28)


pablomh:
I probably won't be able to test it in any device for at least some weeks (new machine at home and ENOTIME to set it up), so if someone does it it would help a lot :(

pablomh:
Rebased.

PR #2406, opened 2021-03-04
[LA.UM.7.1.r1] 4.14.223

pablomh:
This one applied cleanly.


pablomh:
I probably won't be able to test it in any device for at least some weeks (new machine at home and ENOTIME to set it up), so if someone does it it would help a lot :(

When someone can confirm that it doesn't break her/his device, I'll remove the [Draft] status.

jerpelea:
cherry-picked

PR #2414, opened 2021-03-17
[LA.UM.7.1.r1] 4.14.223 + fix build

pablomh:


MarijnS95:
- I hate, **hate** to see the word "fix". Its synonyms aren't much better. Mention you're adding this header, and if it fits: why.
- Always indent things like logs or other things that go over line limit, to make this clear.
- Despite the double mention, add `Fixes:` at the bottom.
- Mention that the linked commit _removed_ the transitive `icmpv6` header inclusion from a header included by `qmi_rmnet` - _that is the reason why it needs to be readded here_.

Finally, something we have no time for: investigate whether upstream fixed this properly by not accessing `icmp6_hdr` either, as that seems to be the reason for its removal from `net: icmp: pass zeroed opts from icmp{,v6}_ndo_send before sending`. Perhaps our driver is not supposed to access or know about it either!

pablomh:
Hi @MarijnS95, thanks for the comments!! Quick answers while I amend it:

- I somehow agree with you regarding the "Fix" word, but in this situation and due to the hurry in having a buildable tree I didn't pay much attention and wanted to have something ready and not "Mainline-quality" :)
- I'll fix the indent, thanks for the advice.
- I had a `Fixes:` tag added but removed it because to me the aformentioned commit didn't introduced a bug, only showed that the existing QCOM code relied (probably wrongly) on that transitive header inclusion.
- As a bonus point, and seeing that the commit ID is not recognized by our repo, I'll include a link to it.
- All the code I've seen upstream continues using the same API: `header = icmp6_hdr(skb)`.

MarijnS95:
@pablomh Rushing is never a good excuse to shortcut a title; if anything it's the most important place to capture all thoughts and changes when in a hurry.

> * I had a `Fixes:` tag added but removed it because to me the aformentioned commit didn't introduced a bug, only showed that the existing QCOM code relied (probably wrongly) on that transitive header inclusion.

That is a fair point - the commit doesn't introduce a bug per-se, it doesn't even know about qcom code in the first place. Not sure what upstream strategy is used to link to the commit that introduced the breakage on our side, I guess in-line is good enough then.

> - As a bonus point, and seeing that the commit ID is not recognized by our repo, I'll include a link to it.

Having the link is nice, we're indeed not having that commit referenced anywhere inside this fork besides being squashed into `.223`. However, since it's `android-4.14-stable` you're better off linking straight to [the original source](https://android.googlesource.com/kernel/common/+/refs/heads/android-4.14-stable) directly: https://android.googlesource.com/kernel/common/+/372fb8e270446213ba51b11bfaa7b9632c0efa90 instead of the massive CAF link - that may make it seem like the breaking change was made **or picked** by CAF as well (in turn leaving readers wondering why there is no CAF commit fixing this exact thing again - at least I would).

> * All the code I've seen upstream continues using the same API: `header = icmp6_hdr(skb)`.

Correct; I could not find any recent tags having this commit picked into it - or it was reverted.

PR #2415, opened 2021-03-22
[LA.UM.7.1.r1] 4.14.225

pablomh:
This one applied cleanly except for some bits, mainly due to the inclusion of the following commit(s) not in "regular" -stable trees:

- 1611298c81197a4109ebfa9c6245a8cef6a8ac39 (ANDROID: dm table: propagate inline encryption flag to dm devices)


jerpelea:
cherry-picked

PR #2424, opened 2021-06-01
[9.12.r1] Kernel updates for Android 12

kholk:
This patchset brings up Android 12 specific kernel updates from the Google common kernel repos.

Please note that this set was only compile-tested.


G8342 (approved):

G8342 (approved):
- @G8342

PR #2435, opened 2021-07-27
[LA.UM.9.12.r1] Add SoMC Lena PDX213 support

konradybcio:
Added all the necessary code to make it compile, only compile-tested.


kholk:
`Reviewed-by: AngeloGioacchino Del Regno <angelogioacchino.delregno@somainline.org>`

arch/arm64/configs/aosp_lena_pdx213_defconfig
```diff
@@ -0,0 +1,814 @@
+CONFIG_LOCALVERSION="-perf"
+# CONFIG_LOCALVERSION_AUTO is not set
```
MarijnS95: We can't have nice things?

kholk: are you really complaining about a first-ever defconfig used for testing before auto-config implementation in the config generator repo?

konradybcio: We can, it's an omission from the stock config

MarijnS95: > are you really complaining about a first-ever defconfig used for testing before auto-config implementation in the config generator repo?

Why not use the config generator cleanly from the get-go instead of running around with incomplete defconfigs for half a year until someone is bothered enough?

PR #2442, opened 2021-09-09
[9.12.r1] Add SDM660 family SoCs and Nile/Ganges platform devices support

bartcubbins:
**Welcome SoMC Nile and Ganges platforms on 4.19 kernel :)**
This is the complete SDM660/636/630 SoCs porting patch set, including SoMC Nile and Ganges platforms.

This **HUGE** pull request contains all the necessary changes for the relatively stable operation of this hardware on the Linux 4.19 kernel. In order to keep things as clean as possible from quirks, legacy platforms and large portions of unused code, most of the things were reimplemented from scratch, some of them came from the downstream SODP 4.14 kernel, some from the kernel-copyleft archives and some from my head :)

**There are a few important things worth clarifying.** The current kernel contains the change that came with commit https://github.com/sonyxperiadev/kernel/commit/21dc7da573b1282e5471e1b14ceb1130676ee2d0 that requires us to have the latest Adreno firmware. During porting, I used the firmware (a530_pfp.fw and a530_pm4.fw) found in the https://git.kernel.org/pub/scm/linux/kernel/git/firmware/linux-firmware.git/ repository. I also used the old display HAL (LA.UM.8.1.r1 branch) since the new one needs new blobs (I really hope they will work).

This patch set has been successfully tested on Nile/Pioneer (SDM630) and Ganges/Mermaid (SDM636) devices.
The compilation tests for the Edo/PDX203/PDX206 and Lena/PDX213 devices were also successfully performed. But also these platforms must be tested with this patch set to prevent unwanted regressions!


kholk:
Adreno's new WHEREAMI also requires blobs support, so don't ever try old adreno blobs, they'll be unstable.

PR #2444, opened 2021-09-15
Revert "Revert "net: qrtr: Only update pkts with broadcast node ID""

Haxk20:
This reverts commit 49d71996332a9adcb63de875014391fcfacc5072.

I reverted this as upstream patch would not  apply cleanly as i thought the fix was in upstream already.
This wasnt the case and caused our phones to freeze after few minutes of uptime.


MarijnS95:
That patch is basically https://github.com/sonyxperiadev/kernel/commit/f51ba60e432a103201b6bcf349a78d581281d064#diff-724dabe0ec23dafa1f85dfa831dd79eb8ff6048b3be50cd23a2ea5fc5ce3958f, but ever so slightly different. That upstream patch should have simply kept the `hdr->dst_port_id` assignments outside of if-else as was done previously; at that point the only difference becomes the check against `sq_node` instead of `sq_port`. I doubt this causes current freezes, but who knows - downstream has different assumptions than upstream.

PR #2448, opened 2021-10-09
[7.1] techpack/audio: Delete Makefile that is already checked in to subrepo

MarijnS95:
This file is also checked in to the techpack-audio [1] repository and
causes unnecessary conflicts when it's managed by two git repositories
at the same time.  It's not necessary to replace it with a placeholder
to keep the folder structure alive as `repo` will create a directory
when syncing.

Unlike wlan-qc, no submodules are used for any of the techpack
repositories.

[1]: https://github.com/sonyxperiadev/kernel-techpack-audio

Signed-off-by: Marijn Suijten <marijn.suijten@somainline.org>


MarijnS95:
Note that everyone will have to `repo sync` / `git reset` their techpack/audio subfolder after this. Unlike locally running `git rm --cached`, applying this commit actually deletes the file from disk (and marking techpack/audio as dirty).

PR #2451, opened 2021-10-22
[7.1] wlan-qc: Use cc-option to check if -Wno-error excludes are available

MarijnS95:
Unfortunately people are still on ancient compilers that are unaware of these specific warnings (hence why -Werror is not a problem to them) but this also means that the compiler fails on "unknown warning option".  Solve that by running through cc-option, which only passes the option if it's available.

Fixes: 22e47a65f69a ("wlan-qc: Disable unfixable CAF warnings-as-errors for newer compilers")

CC @voidanix


voidanix (approved):
LGTM, but clang 13 has _just_ been released and it also does not understand these flags

PR #2454, opened 2021-10-29
{LA.UM.7.1.r1] Revert "selinux: Relocate ss_initialized and selinux_enforcing to sep…

Thaodan:
…arate 4k"

This reverts commit d4807bd60dc166a6445c3b13684f1d1afb722bc5.

That out of tree patch causes the resulting kernel image to be too
large, causing ld.lld to error; likely due to the additional section and
alignment requirements.  Android and floral don't host virtual targets
and thus do not run at EL2 on arm64.

Bug: 63740206
Change-Id: I357bf4d5c7e29230746eacd51fef413acec06067
Signed-off-by: Nick Desaulniers <ndesaulniers@google.com>

See also: https://github.com/kdrag0n/proton-clang/issues/12
I discovered while building for kumano against GCC 11.2.

PR #2455, opened 2021-10-30
[LA.UM.9.12.r1] Revert "selinux: Relocate ss_initialized and selinux_enforcing to sep…

Thaodan:
…arate 4k"

This reverts commit 50240fa8416f5fd183cb147fd3c66aeaeb74a1f8.

That out of tree patch causes the resulting kernel image to be too
large, causing ld.lld to error; since .bss is specified twice.

Fixes the observed error:
ld.lld: error: output file too large: 18446743524330100280 bytes

when using LLD without LTO enabled.

See also: https://github.com/kdrag0n/proton-clang/issues/12

### DIFF
--- a/arch/arm64/kernel/vmlinux.lds.S
+++ b/arch/arm64/kernel/vmlinux.lds.S
@@ -76,10 +76,6 @@ jiffies = jiffies_64;
 #define TRAMP_TEXT
 #endif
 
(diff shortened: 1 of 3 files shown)


Thaodan:
This mirrors: https://github.com/sonyxperiadev/kernel/pull/2454

PR #2460, opened 2021-11-18
[LA.UM.9.12.r1] wlan-qc: Disable unfixable CAF warnings-as-errors for newer compilers

Thaodan:
Newer compilers detect more and more mistakes in these vast codebases,
and they're disallowed with -Werror.  It is infeasible and a waste of
our time to fix them all, but we'd still like to use newer GCC.

### DIFF
--- a/drivers/staging/wlan-qc/Makefile
+++ b/drivers/staging/wlan-qc/Makefile
@@ -1,1 +1,7 @@
 obj-$(CONFIG_QCA_CLD_WLAN)	+= qcacld-3.0/
+
+ccflags_no-error_array_parameter := $(call cc-option,-Wno-error=array-parameter)
+ccflags_no-error_maybe_uninitialized := $(call cc-option,-Wno-error=maybe-uninitialized)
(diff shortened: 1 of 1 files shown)


MarijnS95:
Doing a --reset-author is really poor, no matter the patch contents.

Thaodan:
I didn't use the original patch but just copied the changes since it revert a previous change.

MarijnS95:
This is literally a squash of my two patches, including the title and description.

Thaodan:
A copied the change and the message then commit the changes. I can do --reset-author so you can claim the authorship..

MarijnS95:
That's better.

Hold off merging this for a bit though, I didn't submit this change to 4.19 because it broke Make and/or the compiler and I have to check again what that was or if it's just a fluke.

Thaodan:
It's not I'm working on fixing it right now.

MarijnS95:
:joy:

Thaodan:
The last one works but only sets the first option, now idea how to fix.

Thaodan:
This one works fine for me.

PR #2464, opened 2021-11-24
lena camera dts changes

sraase:


Thaodan:
Why is wl2864c there then if it servers no purpose?

jerpelea:
there is usually a difference between a QCOM reference board and a final product

sraase:
> Why is wl2864c there then if it servers no purpose?

There is no regulator driver for it. But we can use its default configuration and save a lot of headache.

MarijnS95:
In other words, in the current tree there's no driver that handles `qualcomm,gpio_wl2864c` nor `qualcomm,i2c_wl2864c`. @sraase what default are you referring to, the default this regulator provides when it is being turned on through `gpios = <&tlmm 58 GPIO_ACTIVE_HIGH>;` without being - supposedly - configured over i2c?

sraase:
Correct. The wl2864c datasheet is publicly available from OmniVision.

PR #2465, opened 2021-12-04
[9.12.r1] misc: sony_ext_uim_ctrl: Add quirks for SoMC Lena

konradybcio:
Lena only uses one of the GPIOs and sets a different value on probe.
Account for this brokennes to make the secondary SIM slot work..

### DIFF
--- a/drivers/misc/sony_ext_uim_ctrl.c
+++ b/drivers/misc/sony_ext_uim_ctrl.c
@@ -32,7 +32,9 @@ struct sony_ext_uim_ctrl_drvdata {
 	struct device *device;
 	struct mutex lock;
 	int uim2_detect_en_gpio;
+#if !defined(CONFIG_ARCH_SONY_LENA)
 	int uim2_select_gpio;
+#endif
 	bool type;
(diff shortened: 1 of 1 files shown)


MarijnS95:
> and sets a different value on probe

Isn't the DT perhaps inverting GPIO polarity?

jerpelea:
@konradybcio 
current 
 gpio75  : out low  **func0** 2mA no pull
 gpio76  : out low  **func0** 2mA no pull
 gpio77  : out low  **func0** 2mA no pull

expected 
gpio75  : out low  **func1** 2mA no pull
gpio76  : out low  **func1** 2mA no pull
gpio77  : out low  **func1** 2mA no pull

Thaodan:
Would it make more sense to import the newer Sony driver from this that 
already has the workarounds needed for the QN1613D?

jerpelea:
NO! we should fix the existing code

Thaodan:
OK. So what's missing after this PR? From  what I see the newer driver applied the workaround slightly different but similar.

PR #2468, opened 2021-12-10
[9.14.r1] Fix compilation

konradybcio:


jerpelea:
needed by userspace!

MarijnS95:
```c
  HDRTEST usr/include/linux/netfilter_ipv6/ip6_tables.h
In file included from <built-in>:1:
In file included from ./usr/include/linux/netfilter_ipv6/ip6_tables.h:21:
In file included from ./usr/include/linux/if.h:28:
In file included from /usr/include/sys/socket.h:33:
/usr/include/bits/socket.h:191:8: error: redefinition of '__kernel_sockaddr_storage'
struct sockaddr_storage
       ^
./usr/include/linux/socket.h:29:27: note: expanded from macro 'sockaddr_storage'
 #define sockaddr_storage __kernel_sockaddr_storage
                          ^
./usr/include/linux/socket.h:16:8: note: previous definition is here
struct __kernel_sockaddr_storage {
       ^
1 error generated.
```

Notice that there is no leading `.` for `/usr/include/bits/socket.h` - in fact there doesn't appear to be a `bits/socket.h` file anywhere in 9.14 nor the `struct sockaddr_storage` definition. `HDRTEST` appears to be mixing and matching host headers with those found inside the kernel source tree.

```c
#ifndef __KERNEL__
#include <sys/socket.h>			/* for struct sockaddr.		*/
#endif
```

Is expanded to remove `__KERNEL__`, and this external header is provided by glibc. In other words, the commit message of this PR is wrong, as it depends on the target system.

PR #2474, opened 2021-12-28
[LA.UM.9.12.r1] Backport msm pci changes from LA.UM.9.14.r1 

Thaodan:
This PR backports changes from msm pci LA.UM.9.14.r1 to fix issues releasing IRQs
making the reverting of a later mainline redundant.

### DIFF
--- a/drivers/pci/controller/pci-msm-msi.c
+++ b/drivers/pci/controller/pci-msm-msi.c
@@ -28,6 +28,7 @@ struct msm_msi {
 	struct irq_domain *inner_domain; /* parent domain; gen irq related */
 	struct irq_domain *msi_domain; /* child domain; pci related */
 	phys_addr_t msi_addr;
+	u32 msi_addr_size;
 };
 
 /* structure for each client of MSI controller */
@@ -106,7 +107,7 @@ static int msm_msi_domain_prepare(struct irq_domain *domain, struct device *dev,
 	client->msi = msi;
 	client->dev = dev;
 	client->msi_addr = dma_map_resource(client->dev, msi->msi_addr,
-					PAGE_SIZE, DMA_FROM_DEVICE, 0);
+				msi->msi_addr_size, DMA_FROM_DEVICE, 0);
 	if (dma_mapping_error(client->dev, client->msi_addr)) {
 		dev_err(msi->dev, "MSI: failed to map msi address\n");
 		client->msi_addr = 0;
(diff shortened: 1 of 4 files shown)


jerpelea:
cherry-picked

MarijnS95:
Is there a particular advantage history-wise to revert your own revert-commit, within the same PR?

Thaodan:
> Is there a particular advantage history-wise to revert your own
> revert-commit, within the same PR?
Yes I don't have to force push to revert my commit?

MarijnS95:
> Yes I don't have to force push to revert my commit?

Do consider that, while this may be easier for yourself, it introduces extraneous overhead and noise upstream.

Thaodan:
> > Yes I don't have to force push to revert my commit?
> 
> Do consider that, while this may be easier for yourself, it introduces more
> overhead and noise upstream.
> 
I can't do that and this can create issues downstream e.g. In our/hybris 
kernel tree.

MarijnS95:
> I can't do that and this can create issues downstream e.g. In our/hybris kernel tree.

That doesn't set a great precedent for upstream contributions, where the maintainer generally picks (rebase-merges) changes on top. Presuming the upstream project doesn't force push (which we as smaller project try not to do, but is sometimes necessary to combat "accidents"), downstream projects are expected to simply merge in the upstream branch or better: rebase their own patches on top so that they can be smoothly reworked for submission (where necessary) too.

MarijnS95:
It looks like this PR and the picking of it created the wrong outcome now. Your original revert was applied a while ago in e718cae4b3eea (#2471), and now the patches were picked in the wrong order so that it was first reverted again in 7aa55fbaed1c0, and then reapplied again in 4401449fe74f1).

In other words, this PR (you can just create a new branch, nothing that should affect downstream hybris) should have been based on top of at least e718cae4b3eea, and should have never included e4fab65af86562c87ae783714c82d75825b5a66e as it was already applied.

Case in point.

jerpelea:
I will force push the cleanup ASAP

Thaodan:
>It looks like this PR and the picking of it created the wrong outcome now. Your original revert was applied a while ago in e718cae (#2471), and now the patches were picked in the wrong order so that it was first reverted again in 7aa55fb, and then reapplied again in 4401449).

The revert was because the revert was redundant now by the backports applied to the kernel so you dropping the revert is fine.

PR #2475, opened 2021-12-28
[9.14.r1] Add appended DTB support 

konradybcio:


konradybcio:
required configs:

```
CONFIG_BUILD_ARM64_APPENDED_DTB_IMAGE=y
CONFIG_BUILD_ARM64_APPENDED_DTB_IMAGE_NAMES="qcom/lahaina qcom/lahaina-v2 qcom/lahaina-v2.1"
```

arch/arm64/Makefile
```diff
@@ -145,9 +145,15 @@ endif
...
 boot		:= arch/arm64/boot
-KBUILD_IMAGE	:= $(boot)/$(KBUILD_TARGET)
 
-all:	$(KBUILD_TARGET)
+ifeq ($(CONFIG_BUILD_ARM64_APPENDED_DTB_IMAGE),y)
+# Warning! Uncompressed kernel + appended DTB + DTBO WILL NOT WORK!
+KBUILD_IMAGE   := $(KBUILD_TARGET)-dtb
+else
+KBUILD_IMAGE   := $(KBUILD_TARGET)
+endif
```
MarijnS95, line 154: The original commit has this inside `ifeq ($(CONFIG_BUILD_ARM64_KERNEL_COMPRESSION_GZIP),y)`.

arch/arm64/boot/dts/Makefile
```diff
@@ -31,7 +31,21 @@ subdir-y += xilinx
...
+ifneq ($(DTB_NAMES),)
+DTB_LIST := $(addsuffix .dtb,$(DTB_NAMES))
+else
+DTB_LIST := $(dtb-y)
 endif
+targets += $(DTB_LIST)
+
+dtbs: $(addprefix $(obj)/, $(DTB_LIST))
+
+clean-files := dts/*.dtb *.dtb
```
MarijnS95, line 51: This was moved to `arch/arm64/boot/Makefile`.

arch/arm64/Makefile
konradybcio: this way it will work with uncompressed kernel without dtbo

MarijnS95: I dout uncompressed kernel + appended dtb works? Does it?

jerpelea: It was working on 4.19

MarijnS95:
With `vendor_boot` working, I don't think we need this anymore (pending some refactoring of the kernel setup and build before I can PR the entire thing). Unless someone likes spending their time porting other devices to 5.4, we could wait with these patches - at least not set aforementioned `CONFIG_` flags for Sagami.

konradybcio:
setting as draft in case somebody wants to pick this up for new (old) devices

PR #2476, opened 2021-12-31
input: ts: sec_ts: Limit spam from event buffer empty

Thaodan:
On some devices like Xperia 10 II this can happen dozens of items
just by touching the screen limit the output of it.

### DIFF
--- a/drivers/input/touchscreen/sec_ts/sec_ts.c
+++ b/drivers/input/touchscreen/sec_ts/sec_ts.c
@@ -680,7 +680,7 @@ static void sec_ts_read_event(struct sec_ts_data *ts)
 				read_event_buff[0][6], read_event_buff[0][7]);
 
 	if (read_event_buff[0][0] == 0) {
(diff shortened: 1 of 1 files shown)


bartcubbins (approved):

Thaodan:
Original issue is still present after this:
```
 i2c_geni 4a88000.i2c: i2c error :-107
```


Should we push this to other branches too?

bartcubbins:
Let's better fix it properly instead of hiding the symptoms. Do we have a case on the bug tracker describing the current problem and containing the kernel logs?

Thaodan:
As requested:
https://github.com/sonyxperiadev/bug_tracker/issues/738

MarijnS95:
Looks like the link broke in the mail reply - it should be https://github.com/sonyxperiadev/bug_tracker/issues/738

bartcubbins:
@Thaodan thanks!

PR #2477, opened 2022-01-04
[9.14.r1] arm64: dts: qcom: lahaina: Enable 16b ECC

konradybcio:
This improves pstore reliability a bit :wink:


MarijnS95 (approved):
If only dt-bindings checks existed downstream to "see" (or well, document - which gets out of sync with drivers anyway) that `cc-size` is not used anywhere :joy:

PR #2482, opened 2022-01-14
[9.12.r1] USB stack and misc updates

bartcubbins:
The patch set includes updates for USB stack drivers, GIC save/restore state API required for SDM845, and other updates for SoC specific drivers.
With the patch set, we have more stable deep sleep on SDM845, more stable modem operation on SDM845 and SDM630, etc.

**This patch set MUST be tested on Edo and Lena platforms before merging!**

This patch set has been successfully tested on:
- Xperia XZ2 (Akari, SDM845)
- Xperia XA2 (Pioneer, SDM630)
- Xperia 10+ (Mermaid, SDM636)
- Xperia 10 II (Seine, SM6125) (Tests on this upcoming platform were made using my internal branch which will be sent out shortly)


bartcubbins:
PR rebased on the current kernel HEAD

bartcubbins:
PR rebased on the current kernel HEAD.
All SDM845/Tama specific commits have been excluded and will be added in a separate PR as they are quite urgent.

PR #2485, opened 2022-01-18
[9.12.r1] leds: qti-tri-led: Remove LED_KEEP_TRIGGER flag

bartcubbins:
Light HAL in the user space resets the "timer" trigger by setting
the "none" trigger to avoid states when the "timer" trigger will
not be cleared automatically, and the LED would be still in working
blink mode.

To allow it to set the "none" trigger LED_KEEP_TRIGGER flag must not
be set.

This change is required for an upcoming Lights HAL update.

### DIFF
--- a/drivers/leds/leds-qti-tri-led.c
+++ b/drivers/leds/leds-qti-tri-led.c
@@ -411,7 +411,6 @@ static int qpnp_tri_led_register(struct qpnp_tri_led_chip *chip)
 		led->cdev.blink_set = qpnp_tri_led_set_blink;
 		led->cdev.default_trigger = led->default_trigger;
 		led->cdev.brightness = LED_OFF;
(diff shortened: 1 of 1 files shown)


MarijnS95:
> Light HAL in the user space resets the "timer"

The lights HAL is totally broken anyway so you might as well change this.

> This change is required for an upcoming Lights HAL update.

:thinking:

bartcubbins:
I have the lights HAL update since October last year, I just can't get around to PR it😅

MarijnS95:
I haven't PRd most of my changes either, while using them for the better part of 2 years: fix one phone, break the next.

The drivers as just so extremely bad, I ended up waiting for and contributing to (review-wise) the LPG driver upstream and hope to just backport it some day. Until then, discoball notification led FTW :woman_dancing:

PR #2488, opened 2022-02-04
ssusb-redriver-nb7vpq904m.c: Remove leftover #endif

Haxk20:
In commit f7d0ae632da4d51033e413137e13266d506b7ab6
#ifdef that used this #endif was removed but endif was left over

Remove it


MarijnS95 (approved):
Nice catch!

PR #2492, opened 2022-02-13
[9.12.r1] SoMC Kumano audio support

bartcubbins:
This patch set contains audio support for the Kumano platform.
There is also a fix for the screen to take a long time to turn on after pressing the power button.


Haxk20:
@jerpelea Can we get this merged. Have it running on pdx203 for few days now. No issues so far.

PR #2493, opened 2022-02-14
[9.12.r1] arm64: dts: SoMC: Tama: Update wled scale current limit for Apollo

bartcubbins:
The WLED driver found in the current kernel version only supports
setting power values that are predefined in the driver itself in the
wled_fs_current_values array. so let's use 20000 uA for Apollo as well
as Akari. Also move the configuration to Tama common dts and disable the
WLED driver for Akatsuki device because it has an OLED panel.

### DIFF
--- a/arch/arm64/boot/dts/somc/sdm845-tama-akari_common.dtsi
+++ b/arch/arm64/boot/dts/somc/sdm845-tama-akari_common.dtsi
@@ -61,10 +61,6 @@
 	};
 };
 
-&pmi8998_wled {
-	qcom,string-cfg = <7>;
-};
-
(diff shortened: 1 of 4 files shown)


MarijnS95:
Can we switch over to the mainline `wled` driver already? :D

bartcubbins:
@MarijnS95 I will only be glad:)

MarijnS95:
@bartcubbins My patches for it got picked, I think they're in the 5.17 rc's :D

Then the `string-cfg` field can be converted into something that actually makes sense :tada:

jerpelea:
@bartcubbins please fix the typo in PR message

PR #2495, opened 2022-02-21
arm64: dts: somc: kumano: add camera dt support

sraase:
Add camera device tree support for kumano.
Remove the unused sony_camera device nodes.

Do not inherit from QCOM MTP, because our hardware
configuration is different anyway. Results in a
cleaner, separate camera dt.

Untested, handle with care.


bartcubbins:
Camera framework will not work until bandwidth voting logic for CPAS is updated 
Reference:
https://github.com/sonyxperiadev/kernel/commit/a56cd66b6755ee0a28228eb68b0af26da90c3f01
https://android.googlesource.com/kernel/msm-extra/devicetree/+/4c66cf9960b9180ec5f42a1e188f49a588f24ba3

sraase:
I know. This is why this patch is untested. :)
Thanks for the reference, this seems to be exactly what I was missing.

bartcubbins:
Without this you will get a kernel panic, everything else should be fine. I have a similar configuration in my local repository so I can tell that it has already been tested:)

sraase:
The reference is for kona (edo), but the pull request is for sm8150 (kumano), so it doesn't work. Adding a minimal/empty "camera-bus-nodes" only causes problems down the line. So we need to correctly update sm8150-camera.dtsi, but I can't find any patch to do so.

bartcubbins:
I posted this just as an example. I will write the configuration myself as soon as possible or someone else can do it before me

sraase:
My knowledge of doing git is apparently insufficient, so I made a new pull request instead. Sigh.

MarijnS95:
@sraase the new PR was created from the same branch name, did you - as Github says - delete and recreate the fork?

PR #2499, opened 2022-03-03
arm64: dts: qcom: pmi8998: enable automatic string configuration

faenil:
This fixes pmi8998 triggering huge amounts of OVP irqs on Apollo,
causing the ovp IRQ handler to take 50+% CPU when the screen is on,
leading to terrible UI performance.

More info:
https://github.com/sonyxperiadev/bug_tracker/issues/757

### DIFF
--- a/arch/arm64/boot/dts/qcom/pmi8998.dtsi
+++ b/arch/arm64/boot/dts/qcom/pmi8998.dtsi
@@ -543,6 +543,7 @@
 			qcom,current-boost-limit = <970>;
 			qcom,switching-freq = <800>; /* 800KHz */
 			qcom,enabled-strings = <0 1 2>;
+			qcom,auto-string-detection;
 
 			qcom,pmic-revid = <&pmi8998_revid>;
 			status = "disabled";


MarijnS95:
That's a "terrible" commit message again. "Terrible" UI performance is just a symptom, not the issue you're trying to fix. Powering detached strings _may_ be terrible for the hardware in an electrical sense.

You should instead determine what strings are present and what are not, and correct if _for Apollo only_:

https://github.com/sonyxperiadev/kernel/blob/00470f22d64483283ef6fb824dc397647faf5cec/arch/arm64/boot/dts/qcom/sdm845-tama-apollo_common.dtsi#L57-L66

I'm just surprised no-one else ran into this issue - maybe your board or panel is damaged? This should be rechecked with downstream in case an Apollo with different panel lottery exists.

jerpelea:
@MarijnS95  AFAIK there is only 1 panel for apollo

faenil:
> You should instead determine what strings are present and what are not, and correct if for Apollo only:

Indeed, I checked other kernel sources and the strings seemed to match, however I currently have little insight over how this driver works, so it's totally possible that I missed something 😉

If you have any proposal on what to look for exactly, I am happy to have another look as I get the time 👍

> That's a "terrible" commit message again. 

Thanks again for the friendliness 🙂

Unfortunately I don't currently have a lot of insight into how the driver works and the exhaustive list of consequences this change has.
It is a late night effort to try and improve the usability of the current version of the firmware.

I did as much as I could in the given time, cross checked against 4.9 and 4.19 dts, drivers, and they all seem to rely on the automatic string detection.

Given that this is a late-night voluntary effort, I would really appreciate more support, especially for new contributors who are sacrificing their time in an effort to contribute something meaningful.

I tried to include all the context I had in the commit.

If you have more, I am very happy to reword it as your propose.

No hard feelings and let's be more supportive of each other, especially when people are making a real effort 🙂👏

MartinX3:
I agree that some people in the IT environment sound unnecessarily harsh.
But we IT people are not known for our social competence either. :)

faenil:
>But we IT people are not known for our social competence either. :)

Indeed 🙂 just thought I would explicitly state it as cultural differences play a role too, and one might not even be aware of how the message might have looked like from a different perspective 🙂

We can just benefit from being more supportive 👍

Managing and dealing with OSS communities can be mentally challenging indeed 😄

MartinX3:
My perspective is from germany :)
And the more explicitly state it, the bigger the learning effect will be. :)

MarijnS95:
@faenil You're free to share your findings on issues if looking for more support and unsure how to proceed.

Things to consider and add to the commit message:

- What does auto-detection do/mean?
- Why do you think it should be enabled?
- Why isn't `qcom,enabled-strings` enough?
- Why can't we pre-determine what strings to enable?
- Why do you think it should be enabled for _every user of `pmi8998`_, rather than just Tama, or just Apollo?
- As mentioned in the issue for this PR: you apparently took this approach from downstream/other sources, yet the commit message contains no single mention of / reference to this source?
- Finally, there's a tiny bit of room at the bottom of the page to mention that it fixes a side-effect of sluggish UI.

Don't worry about "sacrificing time to contribute something meaningful", we've had our fair share of that over the years.

faenil:
>- What does auto-detection do/mean?
>- Why do you think it should be enabled?
>- Why isn't qcom,enabled-strings enough?

All valid points @MarijnS95, and believe me I'd love to be able to study to get an answer for all of them, as I like investigating and I had the same questions myself.
However, there might be other low hanging fruits that would be a more efficient use of the little time available.

If this were for mainline I wouldn't risk it, but for this project (with a much more restricted scope) I thought the compromise would make sense and would allow us to iterate faster.

This really is a best-effort trial-and-error fix that is based on comparison with Stock 4.9 and SODP4.19 kernels, with the assumption that at least the Stock kernel should be doing the sort-of-right thing (hopefully :grimacing: ).
They both seem to have auto-calibration enabled, so in order to fix the really slow UI on my Apollo, I just went for the quick fix, i.e. align with what is being done on other kernel versions.

Still happy to reword the commit to make it less about symptoms and more about the feature itself, anyway :+1: 

>As mentioned in the issue for this PR: you apparently took this approach from downstream/other sources, yet the commit message contains no single mention of / reference to this source?

I have posted the permalinks to the dts of Stock 4.9 and SODP4.19 in the issue thread  :+1: 
Will add them to the commit message once we have agreed on a solution.

If anyone has more insight and is confident this fix is not the right way to go, I am happy to investigate more :)

jerpelea:
@faenil Thanks for digging into the issue

Since the same workaround is used in the copylefts I think that we should use the same since we do not have all the HW specifications available.

Please update the commit message with this information

faenil:
@jerpelea thanks :)

I will first dig a bit more if that's okay, as I have noticed some flickering when the Sony logo shows at boot, which I guess is the auto detection doing its job.
I did not have that on stock fw, so I first want to verify that's okay 👍 

I will keep you guys posted and update the PR once done 👍

faenil:
Ouch, it seems this got merged, in the end 😅 

I was still planning to do more investigation to make sure this is correct, given the flickering it generates while the logo is showing (at least on my device), just haven't had the time yet

Also the commit was not updated.

Was the merge intentional?

jerpelea:
@faenil for now the patch fixes the existing issue and the devices work better than before. 

When you conclude then investigation we can improve the situation with a new PR, if needed.

faenil:
Okay, thanks for confirming @jerpelea 👍

PR #2506, opened 2022-04-20
[9.12.r1] dts: lena: Fix display brightness control

sledges:
PDX213 panel only regards the first 10 bits of brightness value.

Adjust max brightness to match the workaround mentioned in
https://github.com/sonyxperiadev/kernel/commit/af1ddf3c0940dc3990a764d224fd8be53653e07e

Fixes https://github.com/sonyxperiadev/bug_tracker/issues/764

### DIFF
--- a/arch/arm64/boot/dts/qcom/lagoon-qrd.dtsi
+++ b/arch/arm64/boot/dts/qcom/lagoon-qrd.dtsi
@@ -140,8 +140,8 @@
 	qcom,panel-supply-entries = <&dsi_panel_pwr_supply>;
 	qcom,mdss-dsi-bl-pmic-control-type = "bl_ctrl_dcs";
 	qcom,mdss-dsi-bl-min-level = <1>;
-	qcom,mdss-dsi-bl-max-level = <4095>;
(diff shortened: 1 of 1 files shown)


MarijnS95:
As suggested by https://github.com/sonyxperiadev/bug_tracker/issues/764#issuecomment-1092577996, but GH doesn't allow me to leave a positive review comment :-1:

MarijnS95:
@konradybcio @sledges That workaround is utter garbage though. Just be honest about:

1. Big-endian instead of little-endian (I'm smacking the PDX213 panel with a hammer now for being against DSI_DCS spec, this doesn't deserve light of day);
2. limiting the range programatically to never exceed `1023`, which causes the rollovers @sledges has been seeing in the bug report.

It's nothing special, really.

(I wish this kind of info could be worked into the commit message but it's already been ninja-merged, alas)

Thaodan:
> (I wish this kind of info could be worked into the commit message but it's
> already been ninja-merged, alas)
Add a PR with a comment to that line?

MarijnS95:
I rather have it in the commit(s, both...).

PR #2507, opened 2022-04-22
[9.12.r1] Add Tama specific Synaptics TCM touchscreen driver

bartcubbins:
While porting Tama platform, my hands did not get to this driver, so I am doing it only now:)


MarijnS95:
From patch 2:

> Also defer touch driver probe because of timing requirement
between display and touchscreen.

But it seems you have moved this to a separate patch 3?

PR #2512, opened 2022-06-18
[WIP] Add initial support for Xperia 1 IV (PDX223)

Haxk20:
This platform is a bit different as it only supports overlaying DTBOs.

Copyleft also included lahaina DTS so i included it as a way to speedup later bringup of 1III
(I would prefer having 1IV booting first and then porting it)

Signed-off-by: Martin Botka <martin.botka@somainline.org>


arch/arm64/Kconfig.platforms
```diff
@@ -208,6 +208,24 @@ config ARCH_LAHAINA
...
+	  which is based on QCOM WAIPIO chipset.
+	  If you enable this config,
+	  please use SONY Mobile device tree.
+
+config MACH_SONY_PDX223
+	bool "Sony Mobile PDX-223"
+	depends on ARCH_SONY_NAGARA
+	help
+	  Support for the SONY Mobile PDX-223 board
+	  which is based on SONY Nagaro platform.
```
MarijnS95, line 225: `NagarA`, not `NagarO`, everywhere.

MarijnS95 (changes requested):

arch/arm64/Kconfig.platforms
MarijnS95: Why was this marked as resolved?

jerpelea: He force pushed. I could not grep the typo in the commit.

MarijnS95: The typo was fixed in the `config` name, but not in the documentation which is what this comment was about:

image

Haxk20:
Dammit. Will fix the typo when im home

PR #2514, opened 2022-06-20
[LA.UM.9.14.r1] Update SONY UFS quirks

Haxk20:
Murray platform has an updated UFS quirk
and we need to follow these changes in our kernel
as well as to avoid any damage.

DO NOT BOOT DEVICE WITHOUT THIS!!!

[WRONG]


MarijnS95:
This doesn't look like the original ufs quirk at all, just some extra (odd-looking) logic to write in chunks if the size of the data exceeds the size of a HW sector; perhaps our ufs chip doesn't support that but I would hope the UFS subsystem copes with that at a higher level?

PR #2398, opened 2021-02-16
[LA.UM.7.1.r1] arm: dts: trinket: use qcom,auto-suspend-disable to avoid suspend failures.

krnlyng:
commit 9a0b2664ae23e6cc54736113789fc82e290fa7f9 made a change to NOT depend
on linux pm runtime, this however fails on (at least) seine.

The failure is as below:
android.system.syspend reads from /sys/power/wakeup_count
android.system.suspend writes to /sys/power/wakeup_count
android.system.suspend writes mem to /sys/power/state

The system tries to suspend as follows:
Feb 14 19:22:33 Xperia10II-DualSIM kernel: PM: PM: suspend entry 2021-02-14 17:22:33.365245561 UTC
Feb 14 19:22:33 Xperia10II-DualSIM kernel: PM: suspend entry (deep)
Feb 14 19:22:33 Xperia10II-DualSIM kernel: PM: Syncing filesystems ... done.
Feb 14 19:22:33 Xperia10II-DualSIM kernel: Freezing user space processes ... (elapsed 0.023 seconds) done.
Feb 14 19:22:33 Xperia10II-DualSIM kernel: OOM killer disabled.
Feb 14 19:22:33 Xperia10II-DualSIM kernel: Freezing remaining freezable tasks ... (elapsed 0.003 seconds) done.
Feb 14 19:22:33 Xperia10II-DualSIM kernel: Suspending console(s) (use no_console_suspend to debug)
Feb 14 19:22:33 Xperia10II-DualSIM kernel: msm_geni_serial 4c90000.qcom,qup_uart: msm_geni_serial_stop_tx.Device is suspended.
^ suspend failure here

Feb 14 19:22:33 Xperia10II-DualSIM kernel: PM: Wakeup pending, aborting suspend
Feb 14 19:22:33 Xperia10II-DualSIM kernel: PM: noirq suspend of devices failed
Feb 14 19:22:33 Xperia10II-DualSIM kernel: OOM killer enabled.
Feb 14 19:22:33 Xperia10II-DualSIM kernel: Restarting tasks ...

At this point the 4c90000.qcom,qup_uart driver is in limbo and future
suspend cycles will fail as below:
android.system.suspend reads from /sys/power/wakeup_count
... and becomes stuck in read().

So after the first suspend cycle the system fails to suspend in future cycles
(and does not even attempt to do so because mem is never written to /sys/power/state anymore).

PR #2516, opened 2022-06-23
[7.1.r1] Avoid suspend failures on sm8150

voidanix:
Just like with tama, seine and lena, kumano does not go to sleep without this.

FWIW this patch would probably also have to be applied on 4.19

### DIFF
--- a/arch/arm64/boot/dts/qcom/sdm845-tama-common.dtsi
+++ b/arch/arm64/boot/dts/qcom/sdm845-tama-common.dtsi
@@ -286,6 +286,7 @@
 	qcom,qup_uart@898000 {
 		clock-frequency = <4000000>;
 		status = "okay";
+		qcom,auto-suspend-disable = <1>;
 	};
 
 	/* SPI: QUP8 */
--- a/arch/arm64/boot/dts/qcom/sm8150-kumano-common.dtsi
+++ b/arch/arm64/boot/dts/qcom/sm8150-kumano-common.dtsi
@@ -566,6 +566,7 @@
 	/* UART: QUP13 */
 	qcom,qup_uart@c8c000 {
 		status = "okay";
+		qcom,auto-suspend-disable = <1>;
 	};
 
 	qcom,wdt@17c10000 {


Thaodan:
> Just like with tama, seine and lena, kumano does not go to sleep without this.

I tried this earlier but I wasn't sure if this was needed, good that you checked.

> FWIW this patch would probably also have to be applied on 4.19

Please also create a PR on 4.19 (you can cherry-pick this patch to 4.19).

Thaodan:
Can you use a proper email address to create your commit instead of users.noreply.github.com?

Thaodan:
I might also sense to cherry-pick  to cherry-pick https://github.com/mer-hybris/android_kernel_sony_msm/commit/9e170078f901a6907db6a1382548b8ca394ada48 at the same time.

jerpelea:
@Thaodan  please do

voidanix:
> Can you use a proper email address to create your commit instead of users.noreply.github.com?

Okay.

> I might also sense to cherry-pick to cherry-pick mer-hybris@9e17007 at the same time.

Does that mean in this PR for 4.14 (which means cherry-picking the patch for the platforms mentioned above and updating the 4.19 one) or are _you_ going to do so in separate PRs?

Thaodan:
> > I might also sense to cherry-pick to cherry-pick ***@***.*** at the
> > same time.
> Does that mean in this PR for 4.14 (which means cherry-picking the patch for
> the platforms mentioned above and updating the 4.19 one) or are _you_ going
> to do so in separate PRs?
> 
> 
I can do so however cherry-picking that ref should be easy and avoid 
additional PRs.

arch/arm64/boot/dts/qcom/sdm845-qupv3.dtsi
```diff
@@ -49,6 +49,7 @@
 		interrupts-extended = <&pdc GIC_SPI 607 0>,
 				<&tlmm 48 0>;
 		status = "disabled";
+		qcom,auto-suspend-disable = <1>;
```
Thaodan: The source of this is https://github.com/sailfishos-sony-tama/android_kernel_sony_msm/commit/c85507a577683ca26b764121bbdf669999e4bbbb I'm not sure if the commit should be taken instead since this commit message is better.

MarijnS95: https://github.com/sonyxperiadev/kernel/pull/2398/commits/8f378eeadaf6d8d3398fe35659a9a71e2f07bf64 Yes, please :), better message and maintaining authorship.

Thaodan: > [8f378ee](https://github.com/sonyxperiadev/kernel/commit/8f378eeadaf6d8d3398fe35659a9a71e2f07bf64) Yes, please :)

Agreed.

voidanix: Please note that https://github.com/sailfishos-sony-tama/android_kernel_sony_msm/commit/c85507a577683ca26b764121bbdf669999e4bbbb does this in the tama specific overlay: should the patches go onto the qcom one or not?

If yes, will just reassign authorship.

Thaodan:
As the others went to the qcom overlay I think this one also should go there.

arch/arm64/boot/dts/qcom/sdm845-qupv3.dtsi
jerpelea: it should go into overlay like the others

voidanix: @jerpelea I assume this is what you meant? 👀

PR #2530, opened 2022-08-19
[K.P] Revert "dts: move Sony devices to a separate folder"

konradybcio:
This reverts commit 02ab40677f40ac01cf8206b884b869d93afca7fb.

It makes sense to keep all the qcom things in one place and the
"ease of rebasing" mentioned in the original commit makes little
sense considering the nature of DTB OSS drops, the differences
between SODP and stock kernel and the hacks required to build the
DTB/Os properly without some magic internal qc tools. The only
real way to update these against new releases is careful manual
diffing and examination of what should and shouldn't be merged in.


Thaodan:
> DTB/Os properly without some magic internal qc tools.

Like? It would make it more clear for someone from the outside to understand the diff in DTB's.

That DTB OSS are just drops sounds concerning.

jerpelea:
@Thaodan 
https://github.com/sonyxperiadev/kernel-copyleft-dts/

konradybcio:
>> DTB/Os properly without some magic internal qc tools.

>Like? It would make it more clear for someone from the outside to understand the diff in DTB's

starting with msm-5.10 device trees are split into many overlays, but our bootloader (as far as I am
aware) only supports choosing one for the SoC and one for the specific machine, instead of the 13 or so
required... stock does effectively what I did by hacking in #includes, but I suspect qualcomm has some
internal tools to merge these.. fdtoverlay works for the SoC dt, but it does not for overlaying DTBOs
on top of DTBOs on the device side

>That DTB OSS are just drops sounds concerning.

you only get the final product, no commit history from the oem or qualcomm

Thaodan:
> > > DTB/Os properly without some magic internal qc tools.
> 
> > Like? It would make it more clear for someone from the outside to understand the diff in DTB's
> 
> starting with msm-5.10 device trees are split into many overlays, but our bootloader (as far as I am aware) only supports choosing one for the SoC and one for the specific machine, instead of the 13 or so required... stock does effectively what I did by hacking in #includes, but I suspect qualcomm has some internal tools to merge these.. fdtoverlay works for the SoC dt, but it does not for overlaying DTBOs on top of DTBOs on the device side
> 

Ah thank you that makes sense then. This is not the first time I hear that there's a new feature that the bootloader doesn't support.

> > That DTB OSS are just drops sounds concerning.
> you only get the final product, no commit history from the oem or qualcomm

Yeah that's the issue, context is missing. 
We can only guess from the commit history of the Codeaurora branches if available.

konradybcio:
>We can only guess from the commit history of the Codeaurora branches if available.

since msm-4.19 the device trees are proprietary and not a part of codeaurora (now known as codelinaro) releases

Thaodan:
> > We can only guess from the commit history of the Codeaurora branches if available.
> 
> since msm-4.19 the device trees are proprietary and not a part of codeaurora (now known as codelinaro) releases

Ah the blessing of permissive licenses, to bad :/

Thaodan:
> required... stock does effectively what I did by hacking in #includes, but I suspect qualcomm has some
> internal tools to merge these.. fdtoverlay works for the SoC dt, but it does not for overlaying DTBOs
> on top of DTBOs on the device side

Is that SOMC specific or QCOM? If it's the later I suspect that for this there are better solutions.

PR #2532, opened 2022-08-22
[9.12.r1] Import more CNSS changes from 9.14.r1

Thaodan:


Haxk20:
Can you please provide description of what the PR is even for ? Kinda hard to follow otherwise as "importing CNSS changes" doesnt really explain the reason for why they are needed.

Thaodan:
The idea was that it could help prevent the wifi module from triggering `pci_pm_suspend_noirq(): cnss_pci_suspend_noirq+0x0/0x48 returns -11`  as suggested by @voidanix.
Also to make the suspend/resume work better.

The context from IRC is missing here of course.

jerpelea:
cherry-picked

PR #2536, opened 2022-09-12
[9.12.r1] Fix the bounds checking in snd_soc_put_volsw_sx and snd_soc_put_xr_sx

Thaodan:
The issue is what most of the Qualcomm codecs set the volume levels of controls like this:
-- SOC_SINGLE_SX_TLV("IIR1 INP1 Volume", LPASS_CDC_IIR1_GAIN_B1_CTL, 0,  
-84, 40, digital_gain) --
which causes the values from the caller to be rejected incorrectly on 
the put callback function.

This in turn creates an issue where some Android phones have extremely high amplification on the sidetone mixer during calls which in turn causes a feedback loop because the kernel can't set the correct level on the controls.

This bug was revealed when https://github.com/sonyxperiadev/kernel/commit/4f1e50d6a9cf9c1b8c859d449b5031cacfa8404e was applied in c5b5cf05d4b679ba66b87b81088cd6b887d956b7.

The fix is to apply the change that was made in by pulling in https://github.com/sonyxperiadev/kernel/commit/4cf28e9ae6e2e11a044be1bcbcfa1b0d8675fe4d and apply the patch that fixes both changes when https://github.com/sonyxperiadev/kernel/commit/4a698de58dadcca4a62278b3c3d6bcaaeac7476c is applied.

The fix was found by Tan Nayir, the original discussion can found at:
https://lkml.kernel.org/stable/c2163c71-2f71-9011-3966-baeab8e8dc8f@gmail.com/

cc: @sledges @jerpelea

### DIFF
--- a/include/net/bluetooth/hci.h
+++ b/include/net/bluetooth/hci.h
@@ -1642,13 +1642,15 @@ struct hci_cp_ext_adv_set {
 	__u8  max_events;
 } __packed;
 
+#define HCI_MAX_EXT_AD_LENGTH	251
+
 #define HCI_OP_LE_SET_EXT_ADV_DATA		0x2037
 struct hci_cp_le_set_ext_adv_data {
 	__u8  handle;
 	__u8  operation;
 	__u8  frag_pref;
 	__u8  length;
-	__u8  data[HCI_MAX_AD_LENGTH];
+	__u8  data[];
 } __packed;
 
 #define HCI_OP_LE_SET_EXT_SCAN_RSP_DATA		0x2038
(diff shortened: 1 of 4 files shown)


Thaodan:
Also added which we would use in https://github.com/mer-hybris/android_kernel_sony_msm/pull/122

sledges:
Confirming the first two commits fixed the echo squeak on Lena.

The 3rd ([Bluetooth commit](https://github.com/mer-hybris/android_kernel_sony_msm/pull/122), at least on my device) fixed the confused reporting of Bluetooth state by the status icon on Sailfish OS.

PR #2543, opened 2022-12-23
[9.12.r1] platform: msm: Enable GSI for IPA3

voidanix:


Thaodan:
You mean it will be removed if you generate defconfigs?

voidanix:
If you have both `CONFIG_GSI=y` and `CONFIG_IPA3=y` manually inserted in a
defconfig that will be thrown at `update_defconfig.sh`, once that script is
run the generated defconfigs will not feature `CONFIG_GSI=y`.

So yes, if that is what you meant.

MarijnS95:
https://unix.stackexchange.com/questions/117521/what-is-the-difference-between-select-vs-depends-in-the-linux-kernel-kconfig

This PR doesn't make ~a lot of~ sense. @thaodan is right, `defconfig` != `.config`.

PR #2544, opened 2023-01-09
input: ts: sec_ts: Convert all diagnostic data to input_dbg

Thaodan:
Convert all diagnostic data to input_dbg and pass debug_flag over. Before touchscreen presses were always written to kernel log.
Some data is left, I'm not exactly sure how much of it should be debug.


mlehtima:
Tested that these changes silence reporting touch coordinates to dmesg on Xperia 10 III.

jerpelea:
we should avoid spamming the kernel log with information that is not needed during normal operations

PR #2547, opened 2023-01-18
 [9.12.r1] input: sec_ts: Don't use ts->debug_flag to early

Thaodan:
ts->debug_flag isn't set during sec_ts_parse_dt() it's to early for that. Default to use input_dbg() with dev_dbg which disables the message unless the kernel is compiled with deb_dbg() enabled.

- Fixes: 56245b8160ab67299e5654ed459d711fcb766486
- cc: @jerpelea @voidanix


voidanix:
Can confirm it builds now, thanks!

MarijnS95:
too*

PR #2549, opened 2023-01-22
[7.1.r1] usb: gadget: ffs: Remove IPC logging support

voidanix:
It could be useful to get into 9.12.r1 as well, in case somebody wants to
continue merging new LTS releases.


jerpelea:
please do

PR #2550, opened 2023-01-23
[7.1.r1] linux: 4.14.274

voidanix:
Currently includes #2549 to both simplify merging and avoid a NULL ptr dereference (that panicked the device when unplugging USBs).

Tested on kumano.


Thaodan:
Thank you for your work, so good!

jerpelea:
Thanks!

PR #2545, opened 2023-01-12
[9.12.r1] input: sec_ts: Convert debug information to input_dbg

Thaodan:
Convert all debug information that falsely used input_dbg. The driver doesn't really differentiate between debug or diagnostic information that is helpful for the user such as firmware version or supported features. In quite many cases the  are mixed in between debug information with messages such as start foo and end foo.

I changed all debug information to input_dbg except those regarding firmware, config versions and supported features.

Messages such as screen of and on should be gone unless ts->debug_flags is true.

PR #2554, opened 2023-02-19
input: sec_ts: Don't dereference `ts` pointer before it has been allocated

MarijnS95:
Cc #2545, @thaodan

Some unreviewed and untested change derferences the `ts` pointer in `_probe` before being allocated and assigned to, causing Seine (and likely also Lena) to fail booting.

Fixes: 27e7bb91f9357 ("input: sec_ts: Convert debug information to input_dbg")
Signed-off-by: Marijn Suijten <marijn.suijten@somainline.org>


Thaodan (approved):
LGTM, thanks for fixing my oversight.

PR #2557, opened 2023-04-04
Revert "rpmsg: glink_native: bring in 5.4 changes to wake the bus up"

faenil:
Fixes: [#795](https://github.com/sonyxperiadev/bug_tracker/issues/795)
cc @konradybcio 

Revert https://github.com/sonyxperiadev/kernel/commit/d2fabd8d03f2679dfca23a7f433df9189951ab27 ("rpmsg: glink_native: bring in 5.4 changes to wake the bus up").

The reverted commit was an amalgamation of the code around the original
devm_request_irq prior to
https://github.com/sonyxperiadev/kernel/commit/b21bd352af382264b3d5b7acb2be24e7794a8f17 ("rpmsg: glink: Use threaded interrupts")
and
https://github.com/sonyxperiadev/kernel/commit/f551bd17e4fe7df365799ebaecf2131876537402 ("rpmsg: glink: Request for threaded irq in start phase").
That logic has since been refactored (by those commits), but it is still there.

The reverted commit adds a second request for the same IRQ already requested
by existing code in `qcom_glink_native_start` but it does so via a different API
and using different flags, leading to the pre-existing IRQ request in
`qcom_glink_native_start` failing to complete due to IRQ flags mismatch (on top
of other side effects caused by the duplication of the code creating the
device, sending the version, etc).

The `qcom_glink_native_start` failure causes `adsprpcd` to never complete its init
procedure, which is retried every 25msecs in an endless loop.
```
    genirq  : Flags mismatch irq 407. 00006001 (glink-native-adsp) vs. 00004081 (glink-native-adsp)
    ...
    3000000.remoteproc-adsp: glink-edge: failed to request IRQ with -16
    remoteproc1: failed to probe subdevices for 3000000.remoteproc-adsp: -16
    ...
    adsprpcd: vendor/qcom/proprietary/adsprpc/src/fastrpc_apps_user.c:3426: fastrpc_apps_user_init done
    adsprpcd: vendor/qcom/proprietary/adsprpc/src/adsprpcd.c:41:adsp_default_listener_start called
    adsprpcd: vendor/qcom/proprietary/adsprpc/src/adsp_default_listener.c:67:adsp_default_listener_start started
    adsprpcd: vendor/qcom/proprietary/adsprpc/src/fastrpc_apps_user.c:1437: remote_handle_open: Successfully opened handle 0xffffffff for '":;./\attachguestos on domain 0
    adsprpcd: vendor/qcom/proprietary/adsprpc/src/fastrpc_apps_user.c:3167: Error 0x200: apps_dev_init failed for domain 0, errno No such device, ioErr -1
    ...
    adsprpcd: vendor/qcom/proprietary/adsprpc/src/adsprpcd.c:53:adsp daemon will restart after 25ms...
```
As there is no clarity around whether this change was needed in the first
place, revert it in order to restore the previous behaviour, thus fixing the
flags mismatch and `adsprpcd` init failure loop.

Signed-off-by: Andrea Bernabei <faenil@bernabei.me>

### DIFF
--- a/drivers/rpmsg/qcom_glink_native.c
+++ b/drivers/rpmsg/qcom_glink_native.c
@@ -2158,9 +2158,6 @@ struct qcom_glink *qcom_glink_native_probe(struct device *dev,
 					   bool intentless)
 {
 	struct qcom_glink *glink;
-	u32 *arr;
-	int size;
-	int irq;
 	int ret;
 
 	glink = devm_kzalloc(dev, sizeof(*glink), GFP_KERNEL);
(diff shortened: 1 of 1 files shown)


MarijnS95 (approved):
Make that a `Fixes: #795` to auto-close the report.

The important context here is that the port is an amalgamation of old 5.4 code. All of it has been slightly refactored or moved but is all still there in 5.10 (the second IRQ...) yet the porter didn't notice/blamed this and assumed it was missing. Perhaps a good idea to link the most relevant commits that I shared in the issue.

We'll need to see how this affects 6375, for which it was originally ported...

faenil:
updated, let me know if that reads ok to you.
I'd like to update the commit description, too, once the wording is agreed upon

>We'll need to see how this affects 6375, for which it was originally ported...

indeed... (I unfortunately don't have one to verify myself)

MarijnS95:
Don't want to nit this too much, but I'd replace "changed" with a more specific "moved or refactored".

Then for the second paragraph, rather than just stating that it introduces a second IRQ, I'd say that it's _**requesting** the same IRQ_ but via a different API with slightly different arguments, causing a collision on the flags (and any other weird behaviour we might now have, registering the early notifier twice, sending the version twice, creating the chardev twice).

Complete that with explaining that it - likely? - fails the driver probe entirely, or does it only affect an ioctl sent later by userspace?

The logs you shared - pruned down - go a long way to demonstrate the failure and have a right to be part of the commit message. Don't forget to indent them with four spaces.

For the last paragraph... We just explained in detail why it's plain wrong, calling it "no evidence of serving its purpose" is already giving it too much credit (I'm joking 😁).

(Don't copy that literally, I'm just rambling suggestions on a mobile keyboard.)

faenil:
all valid points. Updated :+1: 

>fails the driver probe entirely, or does it only affect an ioctl sent later by userspace?

probe does not fail, as far as I can tell. The IRQ request in the probe succeeds, it's the pre-existing one in `qcom_glink_native_start` that fails :+1:

faenil:
@jerpelea please wait before merging as there is something else I want to verify (but it's too late now :cry: ) before updating the commit message. Thanks

faenil:
ok all good on my side

MarijnS95:
> probe does not fail, as far as I can tell. The IRQ request in the probe succeeds, it's the pre-existing one in `qcom_glink_native_start` that fails

Ow yeah that makes this a bit confusing to read as the pre existing request is actually called later (hence failing) than the one ported here into probe (which is the first to be called)...

Good to see that you found how to reference commits in text with `<hash> ("<title>")` but feel free to just flow them in-line (the `("` and `")` already delimit the line-broken title).

---

(But that's all nits for the next time, glad to see this merged!)

faenil:
>Ow yeah that makes this a bit confusing to read

yeah, it's unfortunate that `_start` is described as _probe_ in the log messages

> feel free to just flow them in-line

ok ;) I found it tidier to have them on new lines in this case but will keep that in mind :+1:

MarijnS95:
> > Ow yeah that makes this a bit confusing to read
> 
> yeah, it's unfortunate that `_start` is described as _probe_ in the log messages

That isn't too surprising. In order for the `remoteproc` to _probe_, it _probes and starts_ rpmsg glink first. 


Here I'm referring to the explanation in the commit message ;)
> > feel free to just flow them in-line
> 
> ok ;) I found it tidier to have them on new lines in this case but will keep that in mind

In that case add two empty lines around it, and indent with two spaces to truly make it stand out and get formatted as a codeblock.

PR #2566, opened 2023-08-18
[9.16.r1] Compilation fixes for Сlang 14.0.6 (clang-r450784d)

bartcubbins:
A set of changes to successfully compile the kernel without warnings using Сlang 14.0.6 (clang-r450784d).
This is one of the initial patches of my 9.16 journey.

### DIFF
--- a/arch/arm64/boot/dts/somc/Makefile
+++ b/arch/arm64/boot/dts/somc/Makefile
@@ -1,30 +1,26 @@
 ifeq ($(CONFIG_BUILD_ARM64_DT_OVERLAY),y)
 
-ifeq ($(CONFIG_MACH_SONY_PDX214),y)
-dtbo-$(CONFIG_ARCH_SONY_SAGAMI) += lahaina-sagami-pdx214_generic-overlay.dtbo
-endif
-ifeq ($(CONFIG_MACH_SONY_PDX215),y)
-dtbo-$(CONFIG_ARCH_SONY_SAGAMI) += lahaina-sagami-pdx215_generic-overlay.dtbo
-endif
-ifeq ($(CONFIG_MACH_SONY_PDX225),y)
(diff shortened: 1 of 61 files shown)


arch/arm64/boot/dts/somc/Makefile
```diff
@@ -1,30 +1,26 @@
...
-endif
-ifeq ($(CONFIG_MACH_SONY_PDX215),y)
-dtbo-$(CONFIG_ARCH_SONY_SAGAMI) += lahaina-sagami-pdx215_generic-overlay.dtbo
-endif
-ifeq ($(CONFIG_MACH_SONY_PDX225),y)
-dtbo-$(CONFIG_MACH_SONY_PDX225) += blair-murray-pdx225_sm5038_generic-overlay.dtbo
-endif
+# SoMC Sagami
+dtbo-$(CONFIG_MACH_SONY_PDX214) += lahaina-sagami-pdx214_generic-overlay.dtbo
+dtbo-$(CONFIG_MACH_SONY_PDX215) += lahaina-sagami-pdx215_generic-overlay.dtbo
```
MarijnS95, line 5: Can we please just get rid of the MACH altogether and build a unified kernel and separate dtb(o) files per platform? Look what I did for nagara on 5.10, it should be easy to cherry-pick.

Otherwise I can contribute that myself.

bartcubbins: Sure, I will unify it when I have a chance to test it directly on devices

jerpelea: should work on all 5.4+ devices

MarijnS95: @bartcubbins I can port my own patches and test them directly on these devices, if that is useful to you. But after the devices are at least booting to some extent and development is settling here, let me know!

bartcubbins: @MarijnS95 it would be useful while my devices are on their way to me. I'll let you know when I get things in order!

PR #2574, opened 2023-09-07
[9.16.r1] Audio support for SoMC Zambezi

bartcubbins:
Adds audio support and is part of this patch https://github.com/sonyxperiadev/kernel-techpack-audio/pull/57

The patch set has been successfully tested on Murray PDX225 and Zambezi PDX235 devices.


jerpelea:
nice upgrade

bartcubbins:
yes, Awinic got their driver in shape a bit, which allowed us to avoid the terrible binding to the i2c bus name

PR #2586, opened 2023-11-08
[7.1.r1] arm64: dts: qcom: sm8150: Correct addresses from 55.2.A.xxx

voidanix:
**Actually** fixes https://github.com/sonyxperiadev/bug_tracker/issues/694

### DIFF
--- a/arch/arm64/boot/dts/qcom/sm8150.dtsi
+++ b/arch/arm64/boot/dts/qcom/sm8150.dtsi
@@ -648,49 +648,49 @@
 		pil_modem_mem: modem_region {
 			compatible = "removed-dma-pool";
 			no-map;
-			reg = <0x0 0x8dc00000 0x0 0x9600000>;
+			reg = <0x0 0x8dc00000 0x0 0x9700000>;
 		};
 
 		pil_video_mem: pil_video_region {
 			compatible = "removed-dma-pool";
 			no-map;
-			reg = <0x0 0x97200000 0x0 0x500000>;
+			reg = <0x0 0x97300000 0x0 0x500000>;
 		};
 
 		pil_slpi_mem: pil_slpi_region {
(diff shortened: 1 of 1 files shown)


voidanix:
In case you are wondering, running a stock-based kernel with these updated addresses works just fine with both the A10 and A11 modem firmware, so I expect it to "just werk" here as well.

jerpelea (approved):

jerpelea:
not the best explanation but a valid patch

MarijnS95:
The modem subsystem probably needs a bigger memory region (presumably for firmware, PIL = Peripheral Image Loader) but nothing in this patch explains it.

Besides, I'm certain this is _not_ about the addresses anyway. You're increasing the size of the modem PIL memory area, and for that the rest simply has to move out of the way.

jerpelea:
@MarijnS95 valid explanation  ;)

MarijnS95:
If only it would have made it into the commit message :smirk::wink::smirk::wink:. Who knows, _maaaaaybe_ future contributions.

voidanix:
@MarijnS95 yeah, my bad.

The qcom commits regarding this just go "we got new memory mappings lol" without explaining anything: I just went ahead and explained the consequences of not having the DTS updated.

jerpelea:
@voidanix you will do it in the future ;)

Thaodan:
> The modem subsystem probably needs a bigger memory region (presumably for firmware, PIL = Peripheral Image Loader) but nothing in this patch explains it.
> 
> Besides, I'm certain this is _not_ about the addresses anyway. You're increasing the size of the modem PIL memory area, and for that the rest simply has to move out of the way.

I just catched up with this, yeah very late.

Thanks for explaining the change, it does not make sense outright because the context is missing.
Issues like these also very easily explain bugs that happened when people flashed the "wrong" stock Android.

PR #2591, opened 2023-11-22
[7.1.r1] arm64: dts: SoMC: Akatsuki: Set SDE to max performance mode

voidanix:
Turns out that akatsuki's clock calculation shenanigans do not only apply to 4.19.

See https://github.com/sailfishos-sony-tama/main/issues/150 and https://github.com/sailfishos-sony-tama/main/issues/203 for context (and epic artifacts).

### DIFF
--- a/arch/arm64/boot/dts/qcom/dsi-panel-somc-akatsuki.dtsi
+++ b/arch/arm64/boot/dts/qcom/dsi-panel-somc-akatsuki.dtsi
@@ -20,6 +20,11 @@
 #include "dsi-panel-somc-atmel-lgd-1440p2880-cmd.dtsi"
 #include "dsi-panel-somc-akatsuki-id5_pcc.dtsi"
 
+&mdss_mdp {
+	/* Set SDE MAX performance */
+	qcom,sde-perf-default-mode = <3>;
+};
+
 &mdss_dsi0 {
 	/delete-property/ qcom,null-insertion-enabled;
 };


MarijnS95:
The funny thing is that I see a similar glitch - on non-Tama platforms - when changing brightness (the extra command is interrupting or otherwise affecting [available bandwidth for] the datastream).

voidanix:
Out of curiosity, do you also have some white pixelated/noisy, full-brightness screen flashes when pressing the power button? That seems to affect several devices too, but should be unrelated to this.

bartcubbins (approved):

bartcubbins:
> Out of curiosity, do you also have some white pixelated/noisy, full-brightness screen flashes when pressing the power button? That seems to affect several devices too, but should be unrelated to this.

this was fixed at some stage in the 9.12 kernel tag

voidanix:
> this was fixed at some stage in the 9.12 kernel tag

It still persists on kumano (both 4.14 and 4.19), no idea about akatsuki on 4.19 because it gets stuck on the bootloader logo (both with precompiled and built kernel w/ A12 and A13 clang).

bartcubbins:
> It still persists on kumano

AOSP?


> because it gets stuck on the bootloader logo

it's a blobs issue

MarijnS95:
> Out of curiosity, do you also have some white pixelated/noisy, full-brightness screen flashes when pressing the power button? That seems to affect several devices too, but should be unrelated to this.

Not on mainline that I recall, but the panel off/on sequence is a bit off with "desktop-y" window/session managers.

EDIT: I can retry with the `ACTIVE` DRM connector/CRTC property, we've used that to test our shutdown paths but I don't remember what corruption appeared; the DSI power sequence is still being amended upstream due to not perfectly fitting with the `drm_bridge` design.

voidanix:
> AOSP?

Yeah AOSP11 and AOSP12, the issue gets down to SFOS too.

> it's a blobs issue

I see... Hope there will be a fix for this eventually.

PR #2601, opened 2024-01-31
[9.16.r1] SoMC Murray/Zambezi camera flash support + sm5038 cleanup

bartcubbins:
1 Add flash nodes for cameras, define gpios-shared for sharing flash gpio among multiple camera nodes. Gpio sharing
among nodes allows each camera to have its own unique flash node.
2. Enable PM6125 PWM and add a driver node that will control its state.
3. Cleanup function prototypes for SM5038 drivers. Because some prototypes do not correspond to the definition.


jerpelea:
@sraase

PR #2607, opened 2024-03-17
[7.1.r1] Tama LMh fix

voidanix:
Also resolve some section mismatches that spits warnings.

### DIFF
--- a/arch/arm64/boot/dts/qcom/sdm845.dtsi
+++ b/arch/arm64/boot/dts/qcom/sdm845.dtsi
@@ -4129,6 +4129,8 @@
 		reg = <0x17d78800 0x1000>,
 			<0x17d43000 0x1000>;
 		#thermal-sensor-cells = <0>;
+		qcom,plat-mitigation-disable;
+		qcom,legacy-lmh-enable;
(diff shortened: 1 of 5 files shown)


MarijnS95:
Addresses don't seem to be adjusted at all, just the node name? The reg addresses are only being lowercased.

voidanix:
> Addresses don't seem to be adjusted at all, just the node name?

Yes, I just wanted a rename in 0b4daf62226abfc0fd5794ce0859f88f101ee604. Totally butchered the terminology to fit into 50 chars. :|

MarijnS95:
Just be careful that trimming the title _does not_ change its meaning, otherwise a self-imposed char limit serves no purpose at all.

ShujathMohd:
I am no expert here, just tested the patch and here are the results 
Antutu results without this patch

without

With the patch 
with

ShujathMohd:
@bartcubbins. can you please review

bartcubbins:
This change corresponds to SDM845 and restores the LMH DCVS behavior in accordance with what is in the original 4.9 kernel

PR #2608, opened 2024-03-27
dts: somc: murray/zambezi: Fix maximum backlight brightness values

mlehtima:
Fix maximum backlight brightness values for murray and zambezi. Only 11 bits are used for brightness value.

Testing on devices showed that values up to 2047 gave increasing brightness and 2048 made display turn black and higher values would start increasing the brightness as it would start from 0.


bartcubbins (approved):
LGTM

MarijnS95 (approved):

jerpelea (approved):

PR #2615, opened 2024-12-12
[1.0.r1] Initial support for Murray platform

bartcubbins:
This is the bare minimum initial support for the SoMC Murray platform and PDX225 (Xperia 10 IV) device.
The provided set of changes allows to boot into recovery or even into AOSP, but there is a bug
with the display, in which it function until the first reconfiguration. The bug will be isolated
and fixed in the next patches, along with the touchscreen driver and others.


bartcubbins:
Zambezi support, userspace and techpack changes will be PRed soon

jerpelea (approved):

PR #2624, opened 2025-01-27
[9.16.r1] dts: somc: Sagami: Disable hv_haptics driver

bartcubbins:
Sagami platform devices use CS40l2A haptic. Probing of
multiple haptics drivers may result in the first of them
and not necessarily the correct one will be registered
as a led vibrator class device.

### DIFF
--- a/arch/arm64/boot/dts/somc/lahaina-sagami-common.dtsi
+++ b/arch/arm64/boot/dts/somc/lahaina-sagami-common.dtsi
@@ -3722,6 +3722,10 @@
 
 };
 
+&pm8350b_haptics {
+	status = "disabled";
+};
+
 &pm8350c_rgb {
 	red {
 		somc,color_variation_max_num = <4>;


bartcubbins:
@jerpelea

MarijnS95 (commented):
(Normally this should have a default `status="disabled"` in the SoC / PMIC DTSI... but we're downstream anyway)

bartcubbins:
https://github.com/sonyxperiadev/kernel/blob/aosp/LA.UM.9.16.r1/arch/arm64/boot/dts/qcom/lahaina-mtp.dtsi#L107

bartcubbins:
https://github.com/sonyxperiadev/kernel/blob/aosp/LA.UM.9.16.r1/arch/arm64/boot/dts/qcom/pm8350b.dtsi#L67

jerpelea (approved):

PR #2628, opened 2025-04-04
[1.0.r1] SoMC Murray touchscreen support + some other fixes

bartcubbins:
- Add touchscreen driver and adapt it for k5.10.
- SoC Blair Alpha PLL require the FSM legacy mode bit to be set for the PLL to be operational.
- Use MDT loader to boot IPA firmware.
- Check DMA coherency support before allocating memory for GPI channels (Probably the most **important** change, as SoC Blair does not support I/O coherency).

### DIFF
--- a/arch/arm64/boot/dts/qcom/blair.dtsi
+++ b/arch/arm64/boot/dts/qcom/blair.dtsi
@@ -3236,14 +3236,6 @@
 		qcom,ipa-napi-enable;
 	};
 
-	qcom,ipa_fws {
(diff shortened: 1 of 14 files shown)


bartcubbins:
The camera changes were not pushed because the camera framework leads to a kernel freeze, which then triggers a watchdog bark (likely due to a DMA issue). In the next stage, Zambezi support and audio support will be added, and only then will we return to the camera framework.

jerpelea (approved):

PR #2629, opened 2025-04-09
[9.16.r1] power: supply: sm5038: Hardcode USB PD current to 3A

luk1337:
This restores original SM5038 behavior where max charging current is
determined by the cable type, in this case MICRO_CURR_3P0A is used for
POWER_SUPPLY_TYPE_USB_PD. Prior to this change, max charging current
was intended to be set by proprietary SOMC charge_service daemon.


jerpelea (approved):

bartcubbins (approved):

jerpelea (approved):

luk1337:
BTW it'd make sense to cherry-pick this to other branches that support murray ^^

PR #2631, opened 2025-04-16
[1.0.r1] SoMC Murray audio support + some other fixes

bartcubbins:
This PR provides audio support for SoC Blair and Murray platform.
Unfortunately, due to incompatibility with the ADSP firmware supplied with our devices,
we are forced to switch back to the legacy Elite audio framework.

The change includes the full set of modifications necessary for the proper functioning of
the audio subsystem. It also contains fixes for DSP/modem firmware initialization, restart
reasons, Wi-Fi fix, as well as a snapshot of the fingerprint driver.

The remaining part of the patch will be submitted to the techpack audio repository.

### DIFF
--- a/arch/arm64/boot/dts/qcom/blair-murray-common.dtsi
+++ b/arch/arm64/boot/dts/qcom/blair-murray-common.dtsi
@@ -178,6 +178,8 @@
 };
 
 &tlmm {
+	qcom,gpios-reserved = <13 14 15 16>;
+
 	pmx_sde_en: pmx_sde_en {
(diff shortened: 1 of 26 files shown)


jerpelea (approved):

MarijnS95:
> as well as a snapshot of the fingerprint driver.

Are you going to rewrite our HAL to use the downstream Kernel API?

bartcubbins:
this is the same driver that we have on the 5.4 branch, just all the changes are squashed into one commit. I didn't plan to ruin your HAL work <3

